PR #1, opened 2017-03-11
Make 'Stream' classes implement Serializable

mauriciocolli:
We have to implement `Serializable` in order to be able to pass Objects (lists, streams) between activitys, services through Intents (putExtra) .
I've been implementing the quality selector, so it's necessary.

PS: I was thinking in implement the `Parcelable` (faster) , but that would break the command-line, as this class only exist in Android

### DIFF
--- a/stream_info/AudioStream.java
+++ b/stream_info/AudioStream.java
@@ -1,5 +1,7 @@
 package org.schabi.newpipe.extractor.stream_info;
 
+import java.io.Serializable;
+
 /**
  * Created by Christian Schabesberger on 04.03.16.
  *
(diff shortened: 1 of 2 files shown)


theScrabi:
Right breaking the java support is a no go.

mauriciocolli:
Yes, but will you merge this?

 I guess you didn't get what I said, I **_would_** implement the `Parcelable`, but as this breaks the Java version, I stick with the `Serializable`.

mauriciocolli:
Will you merge this?

theScrabi:
Yes

PR #2, opened 2017-03-12
ADD basic playlist support + youtube playlist support

BlenderViking:
Hi,

I begin the refactor of my previous pull request : https://github.com/TeamNewPipe/NewPipe/pull/377

To avoid the same problem as previously encountered (to big code base change), I begin to submit multiple small pull request.

To avoid the same problem as previously encountered, I will send the code by multiple small pull request.

So here the code for import an external playlist in the core extractor app.

I made a unit test but I didn't found how to commit it in the core part.
So i made a gist for store it : https://gist.github.com/BlenderViking/c73c6776580ed661f78f6b0739a66e73

### DIFF
--- a/StreamingService.java
+++ b/StreamingService.java
@@ -52,7 +52,10 @@ public abstract class StreamingService {
     public abstract SearchEngine getSearchEngineInstance();
     public abstract UrlIdHandler getStreamUrlIdHandlerInstance();
     public abstract UrlIdHandler getChannelUrlIdHandlerInstance();
+	public abstract UrlIdHandler getPlayListUrlIdHandlerInstance();
     public abstract ChannelExtractor getChannelExtractorInstance(String url, int page)
+            throws ExtractionException, IOException;
+	public abstract PlayListExtractor getPlayListExtractorInstance(String url, int page)
             throws ExtractionException, IOException;
     public abstract SuggestionExtractor getSuggestionExtractorInstance();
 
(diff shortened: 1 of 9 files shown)


theScrabi:
The unit tests are still part of the NewPipe app itself.

Nice work. :)

StreamingService.java
```diff
@@ -52,8 +52,11 @@ public abstract StreamExtractor getExtractorInstance(String url)
     public abstract SearchEngine getSearchEngineInstance();
     public abstract UrlIdHandler getStreamUrlIdHandlerInstance();
     public abstract UrlIdHandler getChannelUrlIdHandlerInstance();
+	public abstract UrlIdHandler getPlayListUrlIdHandlerInstance();
```
theScrabi: Please fix the spacing.

BlenderViking:
Ok, It's fixed.
I just a question, how to correctly map the git submodules on Android Studio in the main project ?

theScrabi:
Ow good question, I did it using the command line.

theScrabi:
I've [updated](https://github.com/TeamNewPipe/NewPipe/commit/8735cf931adcd2a3c711a7ccbcb386e1696bd5aa) the extractor repository.

BlenderViking:
I found, after clone the repos, we just need to tape in command line :
The init the submodule :
- git submodule init
The upgrade of the submodule :
- git submodule update --recursive --remote
And then in AndroidStudio define in the git config the submodule : 
submodule integration
And we can see the repos of the submodule in Android Studio : 
image

theScrabi:
ah alright

PR #5, opened 2017-04-21
Make some classes Serializable

mauriciocolli:
- Make things serializable for saving in the NewPipe (bundle, intents...) that I'm almost finished
- Fix channel url feed
   - Now just extract the channel `id` and concatenate with a `base url`
- Remove `static` for things that don't need it
    - Before you can't create more instances of the extractors, because of these damns `statics`...

### DIFF
--- a/AbstractStreamInfo.java
+++ b/AbstractStreamInfo.java
@@ -18,8 +18,10 @@ package org.schabi.newpipe.extractor;
  * along with NewPipe.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import java.io.Serializable;
+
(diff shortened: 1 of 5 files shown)


mauriciocolli:
I had to revert some changes because of some other bugs.
For example, the next page of the channel doesn't work, because is  **depending on the previous call** for the `nextPageUrl` to be set (because is static).

We need a refactor to eliminate these statics variables, but for now, just the Serializable is enough.

mauriciocolli:
Hey, any updates?

theScrabi:
No sorry just didn't have time to merge so far.

PR #8, opened 2017-05-27
Extractor is now able to detect if a video is not available to a country

ram-on:
- Users are now informed if a video is not available in their country.
- Minor typo fix


mauriciocolli:
Hey, I think you should squash those commits, there's just too much 'temporary' commits, also some names are repeated, but that should be solved by squashing them.

ram-on:
I've squashed the commits as requested.

PR #11, opened 2017-06-11
Refactor extractor

mauriciocolli:
- Create some abstract classes to minimize repeated code
    - `Extractor` and `Info`
- Channel and Playlist extractors have a `getNextStreams` method now
    - Fix channel's feed URL (closes TeamNewPipe/NewPipe#585)
- Organize some packages and classes
    - Rename package `stream_info` to just `stream`
- Refactor and fix indentation in general

Closes #6

PR #13, opened 2017-07-09
Fix wrong view count

sirekanyan:
This commit fixes view counter on channel page.

The number of views came with non-breaking spaces like `2&nbsp;782&nbsp;442 views` (for 2 782 442 views). Parser fails and I see 2 views instead of 2M views:

screenshot_1499564766

### DIFF
--- a/services/youtube/YoutubeChannelExtractor.java
+++ b/services/youtube/YoutubeChannelExtractor.java
@@ -244,7 +244,8 @@ public class YoutubeChannelExtractor extends ChannelExtractor {
                             return -1;
                         }
 
-                        output = Parser.matchGroup1("([0-9,\\. ]*)", input)
+                        output = Parser.matchGroup1("([0-9,\\. \u00a0]*)", input)
+                                .replace("\u00a0", "")
                                 .replace(" ", "")
                                 .replace(".", "")
                                 .replace(",", "");


mauriciocolli:
That was a oversight in my older pull request (#9), but it's already fixed in my newer (#11).

Also I think it's better to remove any character that it's not numeric (like [this](https://github.com/mauriciocolli/NewPipeExtractor/blob/21e542e7d2fe31dd7d882389567a9848b5428d87/services/youtube/YoutubeChannelExtractor.java#L310)), otherwise we'd to keep putting some edgy cases to handle every single character that could be in between.

sirekanyan:
Sure, `input.replaceAll("\\D+", "")` will just work. I think this PR can be closed.

TheAssassin:
@sirekanyan thanks for your help, it's always appreciated. @mauriciocolli, are you going to implement @sirekanyan's last suggestion? If yes, I'd ask you to close the issue, otherwise @sirekanyan could send another PR.

mauriciocolli:
Yes, it's already included in my last pr (#11), but i can't close it as I don't have permission in this repository.

theScrabi:
@mauriciocolli permission granted.

TheAssassin:
@mauriciocolli yes I actually  asked you, since you took over the issue. I think you have permission to do so in the future now.

PR #14, opened 2017-07-11
Fix dash parser and more refactor

mauriciocolli:
- Refactor some classes
   - Expose more fields for `InfoItem` (like the service id)
- Refactor dash manifest parser
   - Fix `DashMpdParser` creating audio streams with 0 bitrate (closes TeamNewPipe/NewPipe#619)
   - Now it takes the id from the dash, compares it with the Itag list, and if it exists, add the streams to the StreamInfo 
   - Due to this, it extract everything that have in the manifest (video, video only and audio).
- Add some more itags
   - 144p WebM and Mp4 versions and 48kbps m4a audio (see TeamNewPipe/NewPipe#613)


theScrabi:
All right sounds good to me :)

utack:
@mauriciocolli if i were to uncommend the opus lines and build newpipe myself, would it be using opus automatically?

theScrabi:
By the way, i found out we generally had to enhance our dashparser a little to enable livestream support in thr backend. @mauriciocolli could you please take a look at it? :)

utack:
@mauriciocolli could you please tell me where the audio list moved?
Is it still in use?
Thanks

Issue #4, opened 2017-04-18
Other services


wb9688:
Is this code already ready for adding other services? If so, could I start implementing some other services?

theScrabi:
I've never tried it. Would be interesting to see if it works.
For the beginning I recommend to use [np-cli](https://github.com/TeamNewPipe/np-cli). It might be less bitchy about such changes than the Android UI.

wb9688:
OK! Btw can `np-cli` be built from the command line?

theScrabi:
yes/no it should be easy to build it from commandline ... even without a make tool (I think) but I've never tried that :P

wb9688:
OK, why aren't we using eg. Gradle? Btw what should I do with naming? Not
every service calls an user a channel like YouTube does.

theScrabi:
Hm I schould create an empty teamplate service, so It's understandable for other people how services work.

If a service has users, use the channel anyway (but naming is a problem, you are right). Maybe we should rename tha channel class esecially.

However, if your service does not support channel/user return null. The frontend is required to understand what is possible with which service an dwhich is not (however NP Gui can not handle that so far.

wb9688:
There are services that need both a channel ID and a stream ID for a
stream, how should I write the `UrlIdHandler` for them? Currently
`UrlIdHandler` only supports passing one `String`.

wb9688:
OK, I totally forgot about this, since I didn't have much time. I've started working on a SoundCloud service, since I'm familiar with it. I've only done searching and streams (as they're called in NewPipe, in SoundCloud they're called tracks). I've also started working on a TUI, so I could test it. It looks like [this](https://i.imgur.com/vC9idbK.gifv). When I'm finished, I'll make a PR.

TheAssassin:
*Offtopic:* I *love* that TUI you built. Old-ish user interface for modern internet technology.

wb9688:
I've continued working on my SoundCloud service (and my NewPipe TUI) and I only have to implement playlists (as they're called in NewPipe, or sets as they're called on SoundCloud). [Here](https://i.imgur.com/W0h4uGL.gifv) is an updated gifv of it. Btw do we already have ideas for NewPipe's GUI?

TheAssassin:
[...]

PR #15, opened 2017-07-31
Add SoundCloud service

wb9688:
As per #4, I've created a SoundCloud service. I'm sorry that it's one big commit. You can test it with [my TUI](https://github.com/wb9688/np-cli/tree/tui). I'd like to become a maintainer of this. I also want to have access to the Sentry, so that I can fix eventual bugs when it's released in NewPipe. I'll help you with the NewPipe part of this, when we've decided about the design. I'll also refactor NewPipeExtractor a bit after this is merged.

### DIFF
--- a/services/soundcloud/SoundcloudChannelUrlIdHandler.java
+++ b/services/soundcloud/SoundcloudChannelUrlIdHandler.java
@@ -1,7 +1,5 @@
 package org.schabi.newpipe.extractor.services.soundcloud;
 
-import java.io.IOException;
-
 import org.json.JSONObject;
 import org.jsoup.Jsoup;
 import org.jsoup.nodes.Document;
@@ -10,8 +8,6 @@ import org.schabi.newpipe.extractor.Downloader;
 import org.schabi.newpipe.extractor.NewPipe;
 import org.schabi.newpipe.extractor.UrlIdHandler;
 import org.schabi.newpipe.extractor.exceptions.ParsingException;
-import org.schabi.newpipe.extractor.exceptions.ReCaptchaException;
-import org.schabi.newpipe.extractor.utils.Parser.RegexException;
 
 public class SoundcloudChannelUrlIdHandler implements UrlIdHandler {
 
@@ -22,40 +18,52 @@ public class SoundcloudChannelUrlIdHandler implements UrlIdHandler {
     }
 
     @Override
-    public String getUrl(String channelId) throws RegexException, ReCaptchaException, IOException {
-        Downloader dl = NewPipe.getDownloader();
+    public String getUrl(String channelId) throws ParsingException {
+        try {
+            Downloader dl = NewPipe.getDownloader();
 
-        String response = dl.download("https://api-v2.soundcloud.com/user/" + channelId
-                + "?client_id=" + SoundcloudParsingHelper.clientId());
-        JSONObject responseObject = new JSONObject(response);
+            String response = dl.download("https://api-v2.soundcloud.com/user/" + channelId
+                    + "?client_id=" + SoundcloudParsingHelper.clientId());
+            JSONObject responseObject = new JSONObject(response);
 
-        return responseObject.getString("permalink_url");
+            return responseObject.getString("permalink_url");
+        } catch (Exception e) {
+            throw new ParsingException(e.getMessage(), e);
+        }
     }
 
     @Override
-    public String getId(String siteUrl) throws ReCaptchaException, IOException {
-        Downloader dl = NewPipe.getDownloader();
+    public String getId(String siteUrl) throws ParsingException {
+        try {
+            Downloader dl = NewPipe.getDownloader();
 
-        String response = dl.download(siteUrl);
-        Document doc = Jsoup.parse(response);
+            String response = dl.download(siteUrl);
+            Document doc = Jsoup.parse(response);
 
-        Element androidElement = doc.select("meta[property=al:android:url]").first();
-        String id = androidElement.attr("content").substring(19);
+            Element androidElement = doc.select("meta[property=al:android:url]").first();
+            String id = androidElement.attr("content").substring(19);
 
-        return id;
+            return id;
+        } catch (Exception e) {
+            throw new ParsingException(e.getMessage(), e);
+        }
     }
 
     @Override
-    public String cleanUrl(String siteUrl) throws ParsingException, ReCaptchaException, IOException {
-        Downloader dl = NewPipe.getDownloader();
+    public String cleanUrl(String siteUrl) throws ParsingException {
+        try {
+            Downloader dl = NewPipe.getDownloader();
 
-        String response = dl.download(siteUrl);
-        Document doc = Jsoup.parse(response);
+            String response = dl.download(siteUrl);
+            Document doc = Jsoup.parse(response);
 
-        Element ogElement = doc.select("meta[property=og:url]").first();
-        String url = ogElement.attr("content");
+            Element ogElement = doc.select("meta[property=og:url]").first();
+            String url = ogElement.attr("content");
 
-        return url;
+            return url;
+        } catch (Exception e) {
+            throw new ParsingException(e.getMessage(), e);
+        }
     }
 
     @Override
--- a/services/soundcloud/SoundcloudChannelUrlIdHandler.java
+++ b/services/soundcloud/SoundcloudChannelUrlIdHandler.java
@@ -8,6 +8,7 @@ import org.schabi.newpipe.extractor.Downloader;
 import org.schabi.newpipe.extractor.NewPipe;
 import org.schabi.newpipe.extractor.UrlIdHandler;
 import org.schabi.newpipe.extractor.exceptions.ParsingException;
+import org.schabi.newpipe.extractor.utils.Parser;
 
 public class SoundcloudChannelUrlIdHandler implements UrlIdHandler {
 
@@ -68,14 +69,8 @@ public class SoundcloudChannelUrlIdHandler implements UrlIdHandler {
 
     @Override
     public boolean acceptUrl(String channelUrl) {
-        return channelUrl.startsWith("https://soundcloud.com/")
-                && (channelUrl.split("/").length == 4
-                || channelUrl.endsWith("/tracks")
-                || channelUrl.endsWith("/albums")
-                || channelUrl.endsWith("/sets")
-                || channelUrl.endsWith("/reposts")
-                || channelUrl.endsWith("/followers")
-                || channelUrl.endsWith("/following"));
+        String regex = "^https?://(www\\.)?soundcloud.com/[0-9a-z_-]+(/((tracks|albums|sets|reposts|followers|following)/?)?)?([#?].*)?$";
+        return Parser.isMatch(regex, channelUrl);
 
     }
 }
--- a/services/soundcloud/SoundcloudChannelUrlIdHandler.java
+++ b/services/soundcloud/SoundcloudChannelUrlIdHandler.java
@@ -70,7 +70,7 @@ public class SoundcloudChannelUrlIdHandler implements UrlIdHandler {
     @Override
     public boolean acceptUrl(String channelUrl) {
         String regex = "^https?://(www\\.)?soundcloud.com/[0-9a-z_-]+(/((tracks|albums|sets|reposts|followers|following)/?)?)?([#?].*)?$";
-        return Parser.isMatch(regex, channelUrl);
+        return Parser.isMatch(regex, channelUrl.toLowerCase());
 
     }
 }
--- a/services/soundcloud/SoundcloudChannelUrlIdHandler.java
+++ b/services/soundcloud/SoundcloudChannelUrlIdHandler.java
@@ -1,0 +1,73 @@
+package org.schabi.newpipe.extractor.services.soundcloud;
+
+import java.io.IOException;
+
+import org.json.JSONObject;
+import org.jsoup.Jsoup;
+import org.jsoup.nodes.Document;
+import org.jsoup.nodes.Element;
+import org.schabi.newpipe.extractor.Downloader;
+import org.schabi.newpipe.extractor.NewPipe;
+import org.schabi.newpipe.extractor.UrlIdHandler;
+import org.schabi.newpipe.extractor.exceptions.ParsingException;
+import org.schabi.newpipe.extractor.exceptions.ReCaptchaException;
+import org.schabi.newpipe.extractor.utils.Parser.RegexException;
+
+public class SoundcloudChannelUrlIdHandler implements UrlIdHandler {
+
+    private static final SoundcloudChannelUrlIdHandler instance = new SoundcloudChannelUrlIdHandler();
+
+    public static SoundcloudChannelUrlIdHandler getInstance() {
+        return instance;
+    }
+
+    @Override
+    public String getUrl(String channelId) throws RegexException, ReCaptchaException, IOException {
+        Downloader dl = NewPipe.getDownloader();
+
+        String response = dl.download("https://api-v2.soundcloud.com/user/" + channelId
+                + "?client_id=" + SoundcloudParsingHelper.clientId());
+        JSONObject responseObject = new JSONObject(response);
+
+        return responseObject.getString("permalink_url");
+    }
+
+    @Override
+    public String getId(String siteUrl) throws ReCaptchaException, IOException {
+        Downloader dl = NewPipe.getDownloader();
+
+        String response = dl.download(siteUrl);
+        Document doc = Jsoup.parse(response);
+
+        Element androidElement = doc.select("meta[property=al:android:url]").first();
+        String id = androidElement.attr("content").substring(19);
+
+        return id;
+    }
+
+    @Override
+    public String cleanUrl(String siteUrl) throws ParsingException, ReCaptchaException, IOException {
+        Downloader dl = NewPipe.getDownloader();
+
+        String response = dl.download(siteUrl);
+        Document doc = Jsoup.parse(response);
+
+        Element ogElement = doc.select("meta[property=og:url]").first();
+        String url = ogElement.attr("content");
+
+        return url;
+    }
+
+    @Override
+    public boolean acceptUrl(String channelUrl) {
+        return channelUrl.startsWith("https://soundcloud.com/")
+                && (channelUrl.split("/").length == 4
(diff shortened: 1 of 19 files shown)


TheAssassin:
I saw your access request. I'll accept it soon.

May I invite you to our IRC chan? There are not very many devs in it, but IRC is a rather convenient way for longer discussions, and it's growing slowly but steadily. You can also contact e.g. me quickly via IRC. #newpipe on Freenode.

I'll leave the analysis of the PR to someone else (CC @theScrabi) who has more knowledge about the technical aspects.

Great to hear you're interested in maintaining your work. Keep up your work, and perhaps you get an invitation to the project soon!

By the way, I'd love to see the Travis builds to become green one day...

mauriciocolli:
Some things I noticed:
- `nextStreamsUrl` is not set and instead a local variable is created
- The collector is reused (I think that the method `getStreamPreviewInfoCollector()` is bad and doesn't make much sense, I'll remove it anyway) so the next streams get mixed with the previous (in the `ChannelExtractor`)
- There's some unused variables
- Exceptions in the `UrlIdHandler`
    - We could let just the `ParsingException` and throw the other ones as the cause (as it's done [here](https://github.com/TeamNewPipe/NewPipeExtractor/blob/master/services/youtube/YoutubeStreamUrlIdHandler.java#L123..L127))
- Extensive use of `SoundcloudParsingHelper.clientId()` and every time, we download the homepage of SoundCloud and one more js, I think this could definitely use variable caching.
    - The home page is relatively lightweight (20-40KB)
    - The js, otherwise, is **very** heavy (almost 1MB) because SoundCloud packs multiple js' in one file, so I think we could improve it by finding another way to get the client id.
    - But caching the variable seems reasonable _for now_

I'm currently refactoring the extractor and trying to fix some existing problems, tomorrow (I think) will be done. 

@wb9688 if you want, I could wait to resolve these issues, merge this pull request and then I modify it with the changes later, or do you prefer to wait and re-implement this PR on top of them? 
The choice it's up to you, just let me know.

@TheAssassin do you know what happened to Sentry? (the https://support.schabi.org/sentry/)
Currently, it's just a 404 page with a expired SSL certificate from `orange-it.de` (its pointing to the same IP as the other schabi sites).

wb9688:
@mauriciocolli: I'll fix them.

- In what file?
- How should I fix that?
- Yeah, I'll fix that.
- OK, I'll fix that.
- I think the clientId is only grabbed once, because it's final.

I'd prefer if you modify that after this is merged.

mauriciocolli:
> In what file? I'll have to add this. before it, right?

nextStreamsUrl is a variable from the super class `ListExtractor`, I'll fix it when I do the refactor.

> How should I fix that?

As I said I'll remove it, but the fix is simple, just create a new object of that type.

> I think the clientId is only grabbed once, because it's final.

The modifier `final`, when used in a method, means that they can't be override in some other class (subclasses) and the method already is `static` so it can't be override anyway (since it's static is not a "instance level" method).

Just use a field to hold the clientId, if it's null or empty, fetch the pages, otherwise just return it (the variable will live as long as the app process isn't killed).

And I think the `UrlIdHandlers#acceptUrl` implementations are too much strict, for example, the playlist url have to:
- `startsWith` "https://soundcloud.com/", meaning that only https urls will be accepted (what if the url have a `www.` in front?)
- `endsWith` "/tracks" or "/following" or …, what if the url have some parameter in the end?
The job for cleaning the url is of the method (conveniently) `cleanUrl`

Take a look at the current implementations of Youtube (although some of them have some problems like don't accept urls that have some uppercase in the url (the YoutubeStreamUrlIdHandler fix this using to `String#toLowerCase`)).

---

Regarding Sentry, was it reset because of that? I can't log in and it doesn't have an option to register again.

wb9688:
@mauriciocolli: Well, I'm a Java n00b. So, for `acceptUrl` do you have a better suggestion? RegExp?

mauriciocolli:
Don't worry, you're doing a great job.

> Take a look at the current implementations of Youtube...

You can use `String#contains()` in a `url.toLowerCase()`:
- First you check if the url have `soundcloud` in it (or some other abbreviation/short link that might exist)
- Then if it have some of the allowed paths (/sets, /tracks, /following)

In the case of channels/users, the url can be just the name (e.g. https://soundcloud.com/trapcity), just check the first case and if it have more words after that.

wb9688:
I think a RegExp would be easier. Also how will that be done for Android's
intents?

mauriciocolli:
Yes, you can do whatever you feel it's good for testing.

> Also how will that be done for Android's intents?

`boolean acceptUrl` is just a way to see if the given url is valid for extraction for that given extractor/service, for example the `YoutubeChannelUrlIdHandler`:
```java
    @Override
    public boolean acceptUrl(String videoUrl) {
        return (videoUrl.contains("youtube") ||
                videoUrl.contains("youtu.be")) &&
                (videoUrl.contains("/user/") ||
                        videoUrl.contains("/channel/"));
    }
```
It checks if the url is from Youtube and then check if the url contains `/user` or `/channel` (because it knows how to parse those types of url):

|Url|Accept|
|---|:---:|
| https://www.youtube.com/user/CaptainDisillusion | true|
| https://www.youtube.com/watch?v=q6EoRBvdVPQ | false|
| https://soundcloud.com/trapcity/armnhmr-ft-nkolo-oceans | false|
| https://www.youtube.com/channel/UC65afEgL62PGFWXY7n6CUbA  | true|
| https://www.youtube.com/playlist?list=PL34EA971FCA058B20  | false|

You don't need to worry about android, but just to know, you can see how acceptUrl is used in [`StreamingService`](https://github.com/TeamNewPipe/NewPipeExtractor/blob/master/StreamingService.java#L53..L61).

TheAssassin:
@mauriciocolli on IRC I already announced that our Sentry instance crashed during an update, and due to a bug in the PostgreSQL container, the data was not stored in the volume and as the container got removed was lost forever. I changed my backup system to run `pg_dumpall` in containers and made absolutely sure that the files are now saved on the disk and definitely stored in the volume.

You will have to re-register there, I am at the moment importing all the e-mail contents again. What really got lost is stuff like ignored issues. But the issue contents are there. Unfortunately, Sentry won't accept issues >= 30 days old, but I guess those issues won't have to be fixed any more anyway, and if they should, they'd have re-occurred in the meantime.

mauriciocolli:
Oh that's unfortunate, but looking from the brightside, the old issues was stacking up in there, and little bit of noise too (from other copies of the app).

@TheAssassin I think registration is disabled, there's no option in there to do it.

wb9688:
@mauriciocolli: So, I've made the following regexes:

Channel: `^https?://(www\\.)?soundcloud.com/[0-9a-z_-]+(/((tracks|albums|sets|reposts|followers|following)/?)?)?$`
Playlist: `^https?://(www\\.)?soundcloud.com/[0-9a-z_-]+/sets/[0-9a-z_-]+/?$`
Stream: `^https?://(www\\.)?soundcloud.com/[0-9a-z_-]+/(?!(tracks|albums|sets|reposts|followers|following)/?$)[0-9a-z_-]+$`

How could I use them?

Btw I couldn't find a way to register on Sentry either.

theScrabi:
@wb9688 @mauriciocolli I'm moderator on sentry again, so I can accept your registration requests now.

wb9688:
@theScrabi: Well, how can I even register now? Btw can you maybe answer my
question above?

TheAssassin:
As I told @theScrabi (probably missed it), I will post a follow-up in a separate issue once the system is back up.

theScrabi:
@wb9688 about the regex thing:
You have three methods:
* [getStreamUrlIdHandlerInstance()](https://github.com/TeamNewPipe/NewPipeExtractor/blob/master/StreamingService.java#L31)
* [getChannelUrlIdHandlerInstance()](https://github.com/TeamNewPipe/NewPipeExtractor/blob/master/StreamingService.java#L32)
* [getPlaylistUrlIdHandlerInstance()](https://github.com/TeamNewPipe/NewPipeExtractor/blob/master/StreamingService.java#L33)
Each of these return an [UrlIdHandler](https://github.com/TeamNewPipe/NewPipeExtractor/blob/master/UrlIdHandler.java) (Example [here](https://github.com/TeamNewPipe/NewPipeExtractor/blob/master/services/youtube/YoutubeService.java)). These UrlIdHandler's have multiple methodes for:
* Getting the Id of a video out if its Url and vice versa
* Cleaning an Url (So the Url will only contain the information necessary for a request from NewPipe)
* Accepting the Url

###  Url -> Id / Id -> Url
In Youtube the Url of a Video could be:
Url: https://www.youtube.com/watch?v=UTJdmjoOIwE
The Video Id will than be:
Id: [UTJdmjoOIwE](https://www.youtube.com/watch?v=UTJdmjoOIwE)
because it's the minimum required to Identify a Video.
Same for channel/users:
Url: https://www.youtube.com/user/DirtyWhitePaint
Id: [user/DirtyWhitePaint](https://www.youtube.com/user/DirtyWhitePaint) (There is a difference betwean channels and users in Youtube so the user/*** channel/*** part has to remain in the Id)

### Cleaning an Url
Dirty Url: https://www.youtube.com/watch?v=UTJdmjoOIwE&p=kasjfk&c=a
Clean Url: https://www.youtube.com/watch?v=UTJdmjoOIwE
The best you can do here is to just call `getUrl(getId(dirty_url))` this should clean up your Url.

### Accept an Url
[`acceptUrl(String videoUrl)`](https://github.com/TeamNewPipe/NewPipeExtractor/blob/master/UrlIdHandler.java#L36) will return _true_ if the Url can be handled by the current UrlIdHandler, and _false_ if it can't. This will be used in the [RouterActivity](https://github.com/TeamNewPipe/NewPipe/blob/master/app/src/main/java/org/schabi/newpipe/RouterActivity.java) and in [StreamingService](https://github.com/TeamNewPipe/NewPipeExtractor/blob/master/StreamingService.java#L48) to determine which service, and which part of the service should be used to process the Url.

wb9688:
@theScrabi: Well, I know what the UrlIdHandler does, but I wanted to know
how I could use a regex for accepting the URL.

theScrabi:
Ow you mean when NewPipe gets called by another Service?
You have to add an [`<intent-filter>`](https://github.com/TeamNewPipe/NewPipe/blob/master/app/src/main/AndroidManifest.xml#L103) to the RouterActivity in the [AndroidManifest.xml](https://github.com/TeamNewPipe/NewPipe/blob/master/app/src/main/AndroidManifest.xml).

wb9688:
@theScrabi: No, how I can check if a string matches a regex, so I could
implement that acceptUrl function. Like I've said, I'm a Java n00b.

mauriciocolli:
You can use like this:
```java

try {
    //This method throws Parser.RegexException in case of no-match
    Parser.matchGroup(pattern, test, 0); // Or simply Parser.matchGroup1()
    
    // None exception was thrown, return true to signal a match
    return true;
} catch (Parser.RegexException e) {
    // If the matchGroup throws a exception, is because it didn't match
    return false;
}

```
In fact, I think it's useful to create a method in the `Parser`, something like `isMatch(pattern, input)` that does exactly this, returning a `boolean`, what do you think? (If you agree you can create it)

Keep in mind that some of these regex expressions are not working properly:
- This pass > `https://soundcloud.com/trapcity/armnhmr-ft-nkolo-oceans`
- But this don't  > `https://soundcloud.com/trapcity/armnhmr-ft-nkolo-oceans/`
   ###### Note the /

You can test them here in [regexr.com/3gfrd](http://regexr.com/3gfrd) (this failing test is included).

And you'll have to "sanitize" the input before it's tested with these regex (something like `String.trim()` (remove spaces in the start and end of string) and test a lowercase url (then you don't have to worry about the `SOUndCloUD.com` types of url)).

wb9688:
@mauriciocolli: OK, I'll create that function. Is there a reason that
function is called matchGroup1 instead of just matchGroup? Iirc you could
have functions with the same name but a different amount of parameters in
Java. Btw I've accidentally my old regexes that indeed aren't correct.
Since I don't have time for that today, I think I'll do that tomorrow
morning (my time, Europe/Amsterdam, which is UTC+2 atm).

mauriciocolli:
It wasn't me that created that class, but I think they did that because they wanted to be very explicit about it (that they are capturing the group 1, as this group it's frequently used) and indeed, the method could be overloaded.

theScrabi:
Yea I've created the parser class because I hated to write to much code for only one regex xD

services/soundcloud/SoundcloudChannelUrlIdHandler.java
```diff
@@ -68,14 +69,8 @@ public String cleanUrl(String siteUrl) throws ParsingException {
...
     public boolean acceptUrl(String channelUrl) {
-        return channelUrl.startsWith("https://soundcloud.com/")
-                && (channelUrl.split("/").length == 4
-                || channelUrl.endsWith("/tracks")
-                || channelUrl.endsWith("/albums")
-                || channelUrl.endsWith("/sets")
-                || channelUrl.endsWith("/reposts")
-                || channelUrl.endsWith("/followers")
-                || channelUrl.endsWith("/following"));
+        String regex = "^https?://(www\\.)?soundcloud.com/[0-9a-z_-]+(/((tracks|albums|sets|reposts|followers|following)/?)?)?([#?].*)?$";
```
TheAssassin: There is no need to accept non-HTTPS URLs, is there? I strongly doubt Soundcloud themselves redirect to non-HTTPS any more.

services/soundcloud/SoundcloudPlaylistUrlIdHandler.java
```diff
@@ -68,8 +69,7 @@ public String cleanUrl(String complexUrl) throws ParsingException {
 
     @Override
     public boolean acceptUrl(String videoUrl) {
-        return videoUrl.startsWith("https://soundcloud.com/")
-                && !videoUrl.split("/")[4].equals("sets")
-                && !videoUrl.split("/")[5].equals("");
+        String regex = "^https?://(www\\.)?soundcloud.com/[0-9a-z_-]+/sets/[0-9a-z_-]+/?([#?].*)?$";
```
TheAssassin: See my other comment about non-HTTPS URLs.

services/soundcloud/SoundcloudStreamUrlIdHandler.java
```diff
@@ -70,14 +71,7 @@ public String cleanUrl(String complexUrl) throws ParsingException {
...
-        return videoUrl.startsWith("https://soundcloud.com/")
-                && !videoUrl.split("/")[4].equals("tracks")
-                && !videoUrl.split("/")[4].equals("albums")
-                && !videoUrl.split("/")[4].equals("sets")
-                && !videoUrl.split("/")[4].equals("reposts")
-                && !videoUrl.split("/")[4].equals("followers")
-                && !videoUrl.split("/")[4].equals("following")
-                && (videoUrl.split("/").length == 5
-                || videoUrl.split("/").length == 6);
+        String regex = "^https?://(www\\.)?soundcloud.com/[0-9a-z_-]+/(?!(tracks|albums|sets|reposts|followers|following)/?$)[0-9a-z_-]+/?([#?].*)?$";
```
TheAssassin: See my other comment about non-HTTPS URLs.

wb9688:
@TheAssassin: There's a ? after the s, which means there's either one or
zero s at that place.

TheAssassin:
I know, and I fully understand your regular expression. The question is, why even bother about accepting  and processing non-HTTPS URLs? That's just a no-go for any kind of application nowadays, that relies on a professionally maintained web service.

wb9688:
@TheAssassin: Oh, I misunderstood you. Well, @mauriciocolli said:
>meaning that only https urls will be accepted (what if the url have a
`www.` in front?)
So I thought it should also accept http and www, even though SoundCloud
redirects to https and non-www.

TheAssassin:
@mauriciocolli I don't think that non-HTTPS should be accepted by this method at all. I'd rather rewrite the URL to HTTPS internally before handing it to this method.

mauriciocolli:
The `acceptUrl` method, as I told previously, is just a way to tell the caller "Hey, I can handle this url", this sort of cleanup should be in the `cleanUrl` (see how this is handled in the [`YoutubeChannelUrlIdHandler`](https://github.com/TeamNewPipe/NewPipeExtractor/blob/master/services/youtube/YoutubeChannelUrlIdHandler.java#L36..L49), it'll always return a https url).

Of course, nobody uses/have http urls anymore, just that one user that will be wondering why the link doesn't open.

But just for a matter of information, the current `Downloader` implementation of Newpipe doesn't even download http urls, as we are casting every `url.openConnection()` to HttpsUrlConnection and that can raise an exception (that can even be considered a bug (security feature?) as some service may need it for getting something, but that need to be discussed).

services/soundcloud/SoundcloudParsingHelper.java
```diff
@@ -0,0 +1,80 @@
...
+            }
+            return timeAgoString.toString();
+        } catch (ParseException e) {
+            throw new ParsingException(e.getMessage(), e);
+        }
+    }
+
+    public static String toDateString(String time) throws ParsingException {
+        try {
+            SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-ddTHH:mm:ssZ");
```
mauriciocolli: You should escape the char T with `'` (although Z mean other thing, let's just escape as in the emulator Z it doesn't work, see more on [stackoverflow](https://stackoverflow.com/a/2416814/7762724)):

```java
SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'");
```

You should fix in `toTimeAgoString` too.

services/soundcloud/SoundcloudStreamExtractor.java
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return track.getString("description");
+    }
+
+    @Override
+    public String getUploader() {
+        return track.getString("description");
```
mauriciocolli: I think this were supposed to be something like this?
```java
return track.getJSONObject("user").getString("full_name");
```

mauriciocolli:
Ok, I think it's ready to merge, further bugs will be caught with the unit tests after we address the issue #16.

wb9688:
@mauriciocolli: Yeah, I can do that, if you want.

mauriciocolli:
Yes, you already told me in the #16.

PR #18, opened 2017-08-06
Cache clientId

wb9688:
@mauriciocolli: I think this would cache the clientId.

Btw what's the next thing I could work on? Tests for the SoundCloud service?

And why does Travis fail?

### DIFF
--- a/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
+++ b/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
@@ -22,16 +22,20 @@ public class SoundcloudParsingHelper {
     private SoundcloudParsingHelper() {
     }
 
-    public static final String clientId() throws ReCaptchaException, IOException, RegexException {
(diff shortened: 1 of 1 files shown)


mauriciocolli:
I already fix this problem. Btw, your solution would give it a `NullPointerException`, as you are trying to call `String.isEmpty` in a null value.

Just wait till I merge my changes (very shortly) and then you can create your tests. I had a delay because I was refactoring the Soundcloud* services to fit the new changes.

And those tests fail randomly, as said previously, is an area that need improvements.

wb9688:
@mauriciocolli: Oh, OK, I didn't know you were refactoring it locally. Have
you also renamed viewCount to streamCount (or something like that) in the
ChannelInfoItemExtractor (iirc)?

I'll write my tests after you've merged it. What would you like me to do
after I've written those tests?

I also didn't know that `isEmpty` doesn't test for `null`, thanks for
learning me that.

mauriciocolli:
@wb9688 LOL, I didn't noticed that nonsense, thanks for letting me know, I'll use the same as we are using in the PlaylistInfoItem (`streams_count`).

PR #19, opened 2017-08-06
Refactor and improvements

mauriciocolli:
- Change ServiceList into an enum
- Rename and create new overloads for methods of `Info` classes
   - This creates a nicer api, for example, to get a `StreamInfo`:
   ```java
   StreamInfo.getInfo(Youtube, "https://www.youtube.com/watch?v=BAcaa98qQtM"); 
   // Or simply pass the url, it resolves to the right service
   StreamInfo.getInfo("https://soundcloud.com/shupemoosic/pegboard-nerds-try-this"); 
   ```
- Improve SoundCloud service
    - Reduce the number and size of some downloads (e.g. getting the url id now downloads a lighter page (3KB vs 20-30KB before) and remove the need of knowing the client_id to get it)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,5 +10,8 @@ dependencies {
     implementation 'org.mozilla:rhino:1.7.7.1'
 
     testImplementation 'junit:junit:4.12'
+
+    sourceCompatibility = 1.7
+    targetCompatibility = 1.7
 }
 
--- a/src/main/java/org/schabi/newpipe/extractor/Extractor.java
+++ b/src/main/java/org/schabi/newpipe/extractor/Extractor.java
@@ -1,35 +1,73 @@
 package org.schabi.newpipe.extractor;
 
-import org.schabi.newpipe.extractor.stream.StreamInfoItemCollector;
+import org.schabi.newpipe.extractor.exceptions.ExtractionException;
+import org.schabi.newpipe.extractor.exceptions.ParsingException;
 
-import java.io.Serializable;
+import java.io.IOException;
 
-public abstract class Extractor implements Serializable {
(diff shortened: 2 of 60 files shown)


wb9688:
@mauriciocolli: Why do you use their embed player and their normal API?
AFAIK the whole point of NewPipe is to be a YouTube player that parses
YouTube's website without using their API.

theScrabi:
Well if we have disadvantages through the soundCloud api someone would have to write a parser for it's website. But In general what newpipe does always depends of the implementation of the service.

mauriciocolli:
I used what is easier to work with, the api-v2 sometimes have weirder results than the other one, but I don't see why not as they are using it in their own website.

And if that's the reason, why don't we just parse the html that they send in the first place? It comes with the same result of calling the api-v2 (embed in a script tag).

Regarding the embed player, it's because  is much lighter and you don't need to get a clientId to resolve urls (as we don't have a good way to get the clientId without downloading a big file), and for comparison, we use the embed player of YouTube sometimes to parse too.

theScrabi:
> we use the embed player of YouTube sometimes to parse too

That's right.

But just to know, what is the embedded player of soundcloud doing within the soundcloud service?

mauriciocolli:
It's used for getting the id and the url from the id, I think it's hard to get it by hand (unlike Youtube, that have a more predictable url) so we were just downloading the page and getting the url/id from it.

The reason because I used embed player for that I already said previously:

> ... it's because is much lighter and you don't need to get a clientId to resolve urls (as we don't have a good way to get the clientId without downloading a big file)...

theScrabi:
All right I see.

PR #21, opened 2017-08-07
Added description parsing for channel info.

karyogamy:
To be used for displaying description on NewPipe subscription tab channel items.

### DIFF
--- a/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeUserExtractor.java
+++ b/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeUserExtractor.java
@@ -120,6 +120,15 @@ public class YoutubeUserExtractor extends UserExtractor {
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeUserExtractor.java
```diff
@@ -120,6 +120,15 @@ public long getSubscriberCount() throws ParsingException {
     }
 
     @Override
+    public String getDescription() throws ParsingException {
+        try {
+            return doc.select("meta[name=\"description\"]").first().attr("content");
+        } catch(Exception e) {
+            throw new ParsingException("Could not get channel name", e);
```
mauriciocolli: Wouldn't this be "..user's description"?

mauriciocolli:
I will improve the error handling later (and I'll modify this), so just ignore my comment.

PR #22, opened 2017-08-08
Add tests for SoundCloud

wb9688:
- I forgot to test if it correctly throws an exception for SoundCloud Go tracks
- The tests fail, because of some actual bugs in NewPipeExtractor (@mauriciocolli: Could you please fix that?)


mauriciocolli:
- Use `JSONObject#optString` instead of `JSONObject#getString` to some optional values
- Check if url match the pattern before doing anything (for `getId` and `cleanUrl` in the url handlers)
    - Because, unlike Youtube that have a very  easy pattern to match (and it uses to get id from url and clean it), SoundCloud needs to download some pages to know it, that's why we need to check it first.
- Ignore the suggestion tests
    - They were failing because the suggestion was null (the `setSuggestion` was never called)
    - We could initialize it to a empty string, but I think it's better just to ignore that test for now, at least until we improve that part of the extractor.

wb9688:
@mauriciocolli: Nice! What could I work on now?

PR #25, opened 2017-08-16
Playlist info extraction for search engine

karyogamy:
Allows search engine to now extract playlist item for both youtube and soundcloud.


theScrabi:
That's a fine work, although we now need someone to implement this on the client side.

PR #27, opened 2017-08-20
add Kiosk support

theScrabi:
### Add "Kiosk" function

A kiosk can be described as a page that provides content that you don't have to search for, like "top 50" or "trending".

This Pull Request:
* Supports Multiple Kiosks for one Service
* Add "trending" for Youtube
* Add "Top 50" and "New & Hot" for Soundcloud

The Idea is that with this we can finally get rid of the blank startpage of NewPipe.

### DIFF
--- a/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudStreamInfoItemExtractor.java
+++ b/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudStreamInfoItemExtractor.java
@@ -14,18 +14,18 @@ public class SoundcloudStreamInfoItemExtractor implements StreamInfoItemExtracto
     }
 
     @Override
-    public String getWebPageUrl() {
+    public String getUrl() {
         return searchResult.getString("permalink_url");
     }
 
     @Override
-    public String getTitle() {
+    public String getName() {
         return searchResult.getString("title");
     }
 
     @Override
-    public int getDuration() {
-        return searchResult.getInt("duration") / 1000;
+    public long getDuration() {
+        return searchResult.getLong("duration") / 1000L;
     }
 
     @Override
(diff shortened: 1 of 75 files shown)


mauriciocolli:
I had a quick look at it and there's some points to consider:
- Hardcoded type will make it hard to translate it
- Isn't in sync with the latest codebase (using old Json dependency, for example)

theScrabi:
We should fix this asap.

theScrabi:
@mauriciocolli @wb9688 I resolved the collision.

theScrabi:
Are we ready to merge?

wb9688:
@theScrabi: What about:

>Hardcoded type will make it hard to translate it

mauriciocolli:
There are some work to do here, but ok...

**PS:** I think you should have squashed these commits, they messed with the commit history with some useless commits. (instead of clicking `Merge pull request` use the `Squash and merge` option)

src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudStreamInfoItemExtractor.java
```diff
@@ -30,12 +30,13 @@ public long getDuration() {
 
     @Override
     public String getUploaderName() {
-        return searchResult.getObject("user").getString("username");
+        //return searchResult.getObject("user").getString("username");
+        return searchResult.getObject("track").getObject("user").getString("username");
```
mauriciocolli: This change didn't considered that both charts and search (and others) use it, therefore, breaking some things.

mauriciocolli: Fixed in e9156ee499bec59a67f65ca6b78462cf40d026e9.

theScrabi:
The type should be there to identify the different kiosks (like an Id) it is not there to be translated. We should do the translation on the client side, because otherwise wed have to find a way to make the extractor translatable.

src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudStreamInfoItemExtractor.java
theScrabi: thanks :)

PR #30, opened 2017-09-08
Fix Playlist id extraction when length is not 34

karyogamy:
[Example](https://www.youtube.com/playlist?list=PL3418E30855DF35AB)


mauriciocolli:
I think it would be better to put a minimum characters in there, like `{10,}` instead of *, to ignore and fail straight away for playlists like [Watch Later](https://www.youtube.com/playlist?list=WL) (`playlist?list=WL`).

Issue #33, opened 2017-09-21
Rework README.md


theScrabi:
Now since we offer NewPIpeExtractor as a maven repo we might consider to rework the README.md. Also we might add a guide for how to develop the extractor/own services.

mauriciocolli:
Is "maven repo" correct/precise enough? 

We are currently using [jitpack.io](https://jitpack.io/) and people may confound this with the maven repository itself, and because of that, not be able to compile/fetch the extractor.

TheAssassin:
More off-topic: Can someone update the F-Droid description? It is not really appropriate any more, and for PR reasons, we should provide a good text there.

theScrabi:
@mauriciocolli you are right. We should fix that.
@TheAssasine, I can update the description if we decide on an appropriate text.

PR #34, opened 2017-09-22
[WiP] Improve README

wb9688:
I've started working on improving the README as per #33, but as you might see it isn't done yet, so please don't merge it yet.

What things will I need to add?

Now I think about it again, wouldn't it be better to use Javadoc for the documentation?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # NewPipe Extractor [![Build Status](https://travis-ci.org/TeamNewPipe/NewPipeExtractor.svg?branch=master)](https://travis-ci.org/TeamNewPipe/NewPipeExtractor)
 
-NewPipe Extractor is a library for extracting things from streaming sites.
+NewPipe Extractor is a library for extracting things from streaming sites. It is a core component of [NewPipe](https://github.com/TeamNewPipe/NewPipe), but could be used independently.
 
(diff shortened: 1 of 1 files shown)


TheAssassin:
You removed all information about what the extractor is/does (https://github.com/TeamNewPipe/NewPipeExtractor/pull/34/files#diff-04c6e90faac2675aa89e2176d2eec7d8L5). People looking at the repository probably have no clue about what the extractor does. We should add some information about that again, maybe a bit more elaborate.

theScrabi:
I agree with @TheAssasine we should leave a connection to newpipe frontend.

wb9688:
I forgot that, that's why I asked if I forgot something. Would a sentence
like "It is a core component of NewPipe, but could be used independently."
OK? Btw what do you think about the following?

>wouldn't it be better to use Javadoc for the documentation?

theScrabi:
Javadoc might not be such a idea.

Also its no a core component of NewPipe its THE core component of NewPipe :)

wb9688:
@theScrabi: Why would Javadoc not be such a great idea?

Won't PipeCast also be a core component?

theScrabi:
*such a bad idea

Sorry just woke up.

PR #38, opened 2017-10-21
Makes ChannelInfo.getInfo(String url) able to use channel RSS url

guicale:
This pull request makes ChannelInfo.getInfo(String url) able to use the channel's RSS url (for subscription importing and exporting in NewPipe).


src/main/java/org/schabi/newpipe/extractor/StreamingService.java
```diff
@@ -59,6 +59,11 @@ public PlaylistExtractor getPlaylistExtractor(String url) throws IOException, Ex
         return getPlaylistExtractor(url, null);
     }
 
+    public abstract boolean isFeedUrl(String url);
```
coffeemakr: I don't like these two functions exposed.

PR #40, opened 2017-11-08
Add support for getting uploader URL

wb9688:
As per TeamNewPipe/NewPipe#814, but I haven't tested it though.


coffeemakr:
I think you will have to add the field also to the `StreamInfoItem` and  set the field in the `StreamInfoItemCollector` like this: acf05b748cc367eec9e58dd6918fa38795b60b18

wb9688:
@coffeemakr: I didn't think about that, because I had to do this in 15 minutes (including booting my laptop) at the time. I've added your commit to this PR.

wb9688:
Could this be merged now?

theScrabi:
Merged tho would you mind adding some unit tests :)

PR #42, opened 2017-11-22
Added YouTube subtitles support

tonakriz:
### Description
As I said in https://github.com/TeamNewPipe/NewPipe/issues/55, Ive implemented method which you can get subtitles with.  The method from https://github.com/rg3/youtube-dl/ does not work, so I had to reverse engineer how does YT player get its subtitles, which thanks god was not some 'compiled JavaScript magic'.
WebVTT (VTT) format is used.

### How to use
    Just use method called `getSubtitles()`.

### Data Structure
Data are returned as array of objects `Subtitles` with following structure (example):

    result: [
        {
            String languageName = "English (Auto-genreated)";
            String languageCode = "en";
            String URL = "https://youtube.com/api/timedtext?...";
            boolean autoGenerated = true;
        }, {
            String languageName = "Spanish something something";
            String languageCode = "es-419";
            String URL = "https://youtube.com/api/timedtext?...";
            boolean autoGenerated = false;
        }
    ]


wb9688:
Transcript isn't supported by ExoPlayer, look [here](https://google.github.io/ExoPlayer/supported-formats.html) for what formats ExoPlayer supports. Could you please switch to a subtitle format ExoPlayer supports? Tomorrow I'll take a closer look at your code.

KOLANICH:
I wanna mention that google2srt still works.

wb9688:
It looks OK, but you should create a subtitle class. That class should contain the URL (for WebVTT), language code (only `en` or `nl`, never weird things like `a.en`) and if it's auto-generated. The `downloadSubtitles` function should be removed, since we'll pass that URL straight to ExoPlayer.

tonakriz:
@KOLANICH How is google2srt related to this?

tonakriz:
Ok, it should be done now.

src/main/java/org/schabi/newpipe/extractor/Subtitles.java
```diff
@@ -0,0 +1,31 @@
+package org.schabi.newpipe.extractor;
+
+public class Subtitles {
```
theScrabi: Please add a datatype field to this class.
So we can distinguish between different formats.

wb9688: Yeah, and IMHO we should only have the language code and then get the name of the language in NewPipe (like we also do with the YouTube content language).

src/main/java/org/schabi/newpipe/extractor/stream/StreamExtractor.java
```diff
@@ -109,6 +112,7 @@ protected long getTimestampSeconds(String regexPattern) throws ParsingException
     public abstract List<AudioStream> getAudioStreams() throws IOException, ExtractionException;
     public abstract List<VideoStream> getVideoStreams() throws IOException, ExtractionException;
     public abstract List<VideoStream> getVideoOnlyStreams() throws IOException, ExtractionException;
+    public abstract Subtitles[] getSubtitles() throws IOException, ExtractionException, JsonParserException;
```
wb9688: Could you please use a list instead of an array like we're already doing everywhere?

tonakriz:
Sure, no problem.

src/main/java/org/schabi/newpipe/extractor/stream/StreamExtractor.java
tonakriz: Sure, no problem.

tonakriz:
Should be OK this time

KOLANICH:
@tonacriz: it gets subtitles from youtube in its format and converts them to .srt.

theScrabi:
Please don't throw JsonParsingException. Catch them and throw a ExtractionException.

src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudStreamExtractor.java
```diff
@@ -150,6 +147,16 @@ public String getDashMpdUrl() {
         return null;
     }
 
+    @Override
+    public List<Subtitles> getSubtitlesDefault() throws IOException, ExtractionException, JsonParserException {
+        return null;
+    }
+
+    @Override
+    public List<Subtitles> getSubtitles(SubtitlesFormat format) throws IOException, ExtractionException, JsonParserException {
```
theScrabi: Please don't throw a JsonParserException. Please only throw IOException or ExtractionException.
Catch the JasonParserException and throw a ExtractionException embedding the JsonParserException.

tonakriz: Should be OK now.

theScrabi:
Very nice. Thanks :)

PR #43, opened 2017-11-22
Possible removation of duplicated code

tonakriz:
This code was used both in YouTube and SoundCloud module, so I put it together.
I dont know how much you like having "clean" abstract classes, so I made own pull request for this.


theScrabi:
Ah yes, nice work :)

PR #46, opened 2017-11-26
Fix uploader url for related streams

coffeemakr:
Override method to get the uploader url so no exception is thrown. The uploader url is not available in related streams in the legacy view.


theScrabi:
All right works :)

PR #48, opened 2017-11-29
Don't call fetch page in constructor and fix tests

### LINKED ISSUE #45: prevent Extractor form fetching
I noticed we still fetch data in the constructor of the extractors. We use the [fetchPage()](https://github.com/TeamNewPipe/NewPipeExtractor/blob/master/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeStreamExtractor.java#L89) function for fetching, but I think it would be better if this function would be called manually and not through the extractor.

coffeemakr:
* Remove fetchPage() from `YoutubeStreamExtractor` and `SoundcloudStreamExtractor` constructor.
 * Extract the page fetching logic into `onFetchPage` to provide the downloader instance as an argument and ignore duplicated calls to `fetchPage`.
 * Call fetchPage in the StreamInfo.getInfo convenience function (it can not be called otherwise when using these methods)

Fixes #45


coffeemakr:
Also fixed two tests with the commits 4b092828dbaa04aeed531198d872038e420b02ff and 89f8678912d519d480b4a57c8e65a52323645dde:
  * The first was my fault (removed negation of `isEmpty()` without changing `assertTrue` to `assertFalse`)
  * The second (age restriction) test failed  because it was looking at the Adeles stream from a previous test. I'm puzzeled how this wasn't an issue in NewPipe.

PR #52, opened 2017-12-07
Fast tests

Jawnnypoo:
Heya,

Looks like tests were using `@Before` instead of `@BeforeClass` to initialize static things.  In reality, `@Before` runs before each test, and `@BeforeClass` runs before all tests. 

This reduced tests from taking 2 min 9 seconds on my machine to 15.49 seconds

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ build/
 .idea
 local.properties
 out/
+*.iml
 
 # Ignore Gradle GUI config
 gradle-app.setting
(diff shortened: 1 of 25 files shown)


coffeemakr:
I thought about this too. The only drawback is that the tested methods inside the class may have side effects and change the outcome of another test. But I still think it reasonable to only load the page once.

theScrabi:
I agree. I think this is a good idea :)

theScrabi:
@Jawnnypoo  Nice thanks :)

PR #53, opened 2017-12-07
Fix channel info item collector

coffeemakr:
Fix for TeamNewPipe/NewPipe#901 and TeamNewPipe/NewPipe#905

* Fix switched url and name arguments
 * Add tests


coffeemakr:
@theScrabi Ping for fast integration. Also sorry for introducing this bug.

theScrabi:
Oh, I released v0.11.1 with that big yesterday. I should have tested more :/

PR #54, opened 2017-12-18
Age-gated Video Info Retrieval Fix

karyogamy:
This fix is a reprint of the method used in [youtube-dl](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/youtube.py#L1519). Should also fix gated videos that weren't accessible before the "Content Not Available" errors.

_Shameless plug_: I've also made an attempt to [integrate youtube-dl with Newpipe](https://github.com/karyogamy/NewPipe/tree/youtubedl_integration). Though it kind of worked, it didn't go too well. If anybody is interested, I can create a PR (not to be merged) and give some details.

### DIFF
--- a/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeStreamExtractor.java
+++ b/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeStreamExtractor.java
@@ -543,8 +543,6 @@ public class YoutubeStreamExtractor extends StreamExtractor {
     private static final String HTTPS = "https:";
     private static final String CONTENT = "content";
     private static final String DECRYPTION_FUNC_NAME = "decrypt";
-    private static final String GET_VIDEO_INFO_URL = "https://www.youtube.com/get_video_info?video_id=" + "%s" +
-            "&el=info&ps=default&eurl=&gl=US&hl=en";
 
     private volatile String decryptionCode = "";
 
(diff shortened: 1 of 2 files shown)


theScrabi:
Nice. Actually embedeting yt-dl into an android app was my initial try before i wrote NewPipe :D. However I could not do it. Also I got the feeling yt-dl is horrible spagety code (at least the extractors), and since yt-dl only covers the extraction of stream urls it only hits a part of what we need.

Anyway. How did you get it managed to run python within an app?

karyogamy:
To get python embedded on Android, I had to use [Crystax NDK](https://www.crystax.net/). Most of the processes are described in [PyBridge](https://github.com/joaoventura/pybridge), which I copied over as a template. It goes from Android code -> compiled C (.so) -> Python (stdlib.zip, includes youtube-dl). 

The caveat for using NDK is that we need to have C code compiled for different ABI versions across phones. This means that we need to either include binaries (.so files) in our build or have a custom build process on F-Droid (to include the NDK), which I am [not sure is possible](https://f-droid.org/en/docs/Reproducible_Builds/). Also, this will bloat the app for at least 4-5MB. Though, a good thing about this is that we can use [youtube-dl's postprocessor](https://github.com/rg3/youtube-dl/tree/master/youtube_dl/postprocessor) and [Crystax's ffmpeg](https://github.com/crystax/android-vendor-ffmpeg) to mix audio and video since it is just another lib that can be built and imported as a .so file.

The share button is hotwired to accept all websites, you can take a look to see the extent of what youtube-dl is currently capable of fetching.

>  horrible spagety code

I wouldn't recommend this approach either, this was just an exploration to see what is possible and there are some nasty hacks we need to use to get it working (e.g. ssl certificate on Android is different from what openssl uses). There are successes, such as [arkTube](https://forum.xda-developers.com/general/paid-software/app-arktube-ultimate-youtube-downloader-t3321818), [YTD ](http://dentex.github.io/apps/youtubedownloader/) and [avidd](https://forum.xda-developers.com/android/apps-games/app-youtube-dl-material-gui-android-t3498492), but I don't think we have enough people to overextend.

theScrabi:
yep thank you :)

PR #56, opened 2017-12-29
Adds getInfoFromFeed to ChannelInfo and a few other changes to be able to get channel info from feed URL

guicale:
After my previous pull request I decided to redo my changes from scratch following the recommendations @coffeemakr gave me. I tested my changes with a modfied version of np-cli and everything seem to work fine. If there's anything I should change, let me know and I'll do it.


mauriciocolli:
It seems that you just copied and rename the files, which is not what a feed extractor is (it doesn't just handle the _url_, it extracts base on it, _i.e._ actually fetching it).

I'll let this PR open for awhile in case you decide to do the proper implementation.

guicale:
I'm sorry, I don't really understand what the problem is. The ChannelExtractor and FeedExtractor implementations are similar because the feed don't contain all the information about the channel, so I think it would always be necessary to extract the info from the actual channel page. All my implementation of FeedExtractor does is extract the channel URL from the feed and use that to fetch the page info.

theScrabi:
@mauriciocolli is right, FeedExtractor should not work as a replacement for the ChannelExtractor. FeedExtractor should only return feed information, but not data for which the ChannelExtractor is responsible.

Sich to the Idea of Object orientation. The name of the class FeedExtractor implies that it's extracting data from a feed file.
Now what data is a feed providing?
What kind of data was importand for us?
Now only return this data :)

guicale:
Ok, that makes sense. I'll redo it and get back to you. Sorry for all the trouble.

guicale:
Would it be OK to open another pull request on another branch after all the mess I've made here?

theScrabi:
Thats not trouble. Thats programming ;)

guicale:
I've been thinking and I don't think a FeedExtractor and FeedInfo would be that useful. Originally, I started working on this to add the ability to add the ability to import subscriptions to NewPipe, but I think it's a little overblown to do all of this just for that. Do you think there would be more uses for the ability to handle feeds?

wojcik-online:
@guicale It would be super useful to handle the feeds for the "Whats's New" page because the feeds are smaller and would be downloaded and parsed faster than the YouTube website; see #61.
But not all other services will support the feed vs. website paradigm. For example I've been thinking about implementing a podcast RSS service and there aren't these two possibilities. Even in the case of SoundCloud: What is the advantage of using the feed vs. the API?
IMHO, the `FeedExtractor` should be an internal thing of the YouTube service.

theScrabi:
This is indeed not that easy to answer. But isn't it possible to bring up feed like behaviour even tho no feed is available?

wojcik-online:
What I meant to say:
It's a quirk of YouTube that there is this superior way of getting the channel info via `/videos.xml`, but in some circumstances it doesn't work. So both ways should be supported.
Perhaps the `YoutubeService` should try to get the channel info from the feed first and only after that has failed, fall back to parsing the website?

Other streaming services might have similar quirks, but most of them probably don't and there is a clear best way of getting the info.

guicale:
As @wojcik-online says, feeds aren't a standard feature of media services. For example, SoundCloud has a feed for each user, but not all tracks are available in the feed (this is controlled by the user). This means that a SoundCloud feed would not be reliable to get a user's tracks. However, there are websites that, similarly to YouTube, provide a feed that contains all of a user's latest videos, so it's not a feature unique to YouTube. Should feed handling be implemented individually by the services or should it be included in the base as are channels and playlists and then services just extend FeedExtractor?

theScrabi:
We could give the ability to create and return a feed extractor for trending, channel, playlist

theScrabi:
Then we could put merhodes like "getChannelFeed()" inside the Service class.

theScrabi:
I thought about this, actually feed is a standardised format, therefore the one who implements a service should not take care about actual feed extraction.
Therefore the NewPipeExtractor itself should add support for feed extraction it self, either through an external librarie, or through our own implementation (not recomended since more code to maintain). The extractors should than only return an address to the feed, just like its already been done [here](
https://github.com/teamnewpipe/newpipeextractor/blob/master/src/main/java/org/schabi/newpipe/extractor/channel/ChannelExtractor.java#L55).

PR #57, opened 2017-12-30
Fix channel extractor error

karyogamy:
getInfo is using a new extractor instead of using the loaded extractor, which later causes NPE.


theScrabi:
thx

PR #59, opened 2018-01-04
Add more tests and fix 2 bugs

coffeemakr:
Bugs fixed:
 * Uploader URL was relative for playlist items.
 * Soundcloud extractor returned http:// link for avatar and uploader URLs

Tests improved:
 * Add `assertIsSecureUrl` method
 * Add common base classes for search engine tests so the result testing can be improved at a single location.

### DIFF
--- a/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeStreamInfoItemExtractor.java
+++ b/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeStreamInfoItemExtractor.java
@@ -94,23 +94,13 @@ public class YoutubeStreamInfoItemExtractor implements StreamInfoItemExtractor {
         }
     }
 
-    private static String fixRelativeUrls(String url) {
-        if(url == null || url.isEmpty()) return url;
-        if(url.startsWith("/")) {
-            return "https://youtube.com" + url;
-        } else {
-            return url;
-        }
-    }
-
     @Override
     public String getUploaderUrl() throws ParsingException {
         try {
             try {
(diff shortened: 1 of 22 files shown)


src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeStreamInfoItemExtractor.java
```diff
@@ -94,13 +94,23 @@ public String getUploaderName() throws ParsingException {
...
+
     @Override
     public String getUploaderUrl() throws ParsingException {
         try {
             try {
-                return item.select("div[class=\"yt-lockup-byline\"]").first()
+                String url = item.select("div[class=\"yt-lockup-byline\"]").first()
                         .select("a").first()
                         .attr("href");
+                return fixRelativeUrls(url);
```
mauriciocolli: Instead of checking manually, just use `abs:<attr>` like this: [YoutubeStreamInfoItemExtractor.java#L55](https://github.com/TeamNewPipe/NewPipeExtractor/blob/master/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeStreamInfoItemExtractor.java#L55)

In fact, there's some more places where this happens, you can fix them in this pull request if you want.

coffeemakr: Good to know. Fixed

wb9688:
I know this has already been merged, but couldn't those `SoundcloudExtractorHelper` functions be in `SoundcloudParsingHelper`?

coffeemakr:
Yes you're right, the methods would fit there.

PR #61, opened 2018-01-28
Parse the upload dates of StreamInfoItems.

wojcik-online:
The class `StreamInfoItem` gets a new field with the parsed upload date. It is the responsibility of the `StreamingService` to parse the date.

Services using RSS feeds or an API should have no problem with this. For YouTube a relative date in the format '*2 days ago*' must be parsed.
For this a simple approach is used where every time unit (minute / day / week / ...) gets a list of phrases. If the date from YouTube contains a phrase, it is assumed that the corresponding time unit was meant. (e.g '*2 minutes ago*' contains '*min*') The phrases are provided by the front end.

Linked to: Sort the items in the "What's New" feed by the parsed upload date.

### Still to be done (optionally)

 - [ ] **1.** 
   Parse the dates for every `Info` / `InfoItem`. This is not really necessary because the dates would not be used, but it might be useful in the future.
    And for the sake of consistency it could be implemented, now. Only drawback: computational overhead.

 - [ ] **2.**
    Get the exact upload dates from another source. The YouTube Channel RSS feed might be a good idea, but apparently it is not always available. Perhaps the `YoutubeService` could try to use the RSS feed and only upon failure fall back to the HTML page.
   The overhead should be negligible because most channel have a RSS feed. Which URLs should be used in that case?

### DIFF
--- a/src/main/java/org/schabi/newpipe/extractor/Extractor.java
+++ b/src/main/java/org/schabi/newpipe/extractor/Extractor.java
@@ -7,12 +7,12 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.io.IOException;
 
-public abstract class Extractor {
+public abstract class Extractor<S extends StreamingService> {
     /**
      * {@link StreamingService} currently related to this extractor.<br/>
      * Useful for getting other things from a service (like the url handlers for cleaning/accepting/get id from urls).
      */
-    private final StreamingService service;
+    private final S service;
 
     /**
      * Dirty/original url that was passed in the constructor.
@@ -36,7 +36,7 @@ public abstract class Extractor {
     private boolean pageFetched = false;
     private final Downloader downloader;
 
-    public Extractor(StreamingService service, String url) throws ExtractionException {
+    public Extractor(S service, String url) throws ExtractionException {
         if(service == null) throw new NullPointerException("service is null");
         if(url == null) throw new NullPointerException("url is null");
         this.service = service;
@@ -113,7 +113,7 @@ public abstract class Extractor {
     }
 
     @Nonnull
-    public StreamingService getService() {
+    public S getService() {
         return service;
     }
 
--- a/src/main/java/org/schabi/newpipe/extractor/ListExtractor.java
+++ b/src/main/java/org/schabi/newpipe/extractor/ListExtractor.java
@@ -10,13 +10,13 @@ import java.util.List;
 /**
  * Base class to extractors that have a list (e.g. playlists, users).
  */
-public abstract class ListExtractor extends Extractor {
+public abstract class ListExtractor<S extends StreamingService> extends Extractor<S> {
     protected String nextStreamsUrl;
 
     /**
      * Get a new ListExtractor with the given nextStreamsUrl set.
      */
-    public ListExtractor(StreamingService service, String url, String nextStreamsUrl) throws ExtractionException {
+    public ListExtractor(S service, String url, String nextStreamsUrl) throws ExtractionException {
         super(service, url);
         setNextStreamsUrl(nextStreamsUrl);
     }
--- a/src/main/java/org/schabi/newpipe/extractor/channel/ChannelExtractor.java
+++ b/src/main/java/org/schabi/newpipe/extractor/channel/ChannelExtractor.java
@@ -29,9 +29,9 @@ import java.io.IOException;
  * along with NewPipe.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-public abstract class ChannelExtractor extends ListExtractor {
+public abstract class ChannelExtractor<S extends StreamingService> extends ListExtractor<S> {
 
-    public ChannelExtractor(StreamingService service, String url, String nextStreamsUrl) throws IOException, ExtractionException {
+    public ChannelExtractor(S service, String url, String nextStreamsUrl) throws IOException, ExtractionException {
         super(service, url, nextStreamsUrl);
     }
 
--- a/src/main/java/org/schabi/newpipe/extractor/kiosk/KioskExtractor.java
+++ b/src/main/java/org/schabi/newpipe/extractor/kiosk/KioskExtractor.java
@@ -28,11 +28,11 @@ import org.schabi.newpipe.extractor.exceptions.ParsingException;
 import javax.annotation.Nonnull;
 import java.io.IOException;
 
-public abstract class KioskExtractor extends ListExtractor {
+public abstract class KioskExtractor<S extends StreamingService> extends ListExtractor<S> {
     private String contentCountry = null;
     private final String id;
 
-    public KioskExtractor(StreamingService streamingService,
+    public KioskExtractor(S streamingService,
                           String url,
                           String nextStreamsUrl,
                           String kioskId)
--- a/src/main/java/org/schabi/newpipe/extractor/playlist/PlaylistExtractor.java
+++ b/src/main/java/org/schabi/newpipe/extractor/playlist/PlaylistExtractor.java
@@ -9,9 +9,9 @@ import org.schabi.newpipe.extractor.exceptions.ParsingException;
 import javax.annotation.Nonnull;
 import java.io.IOException;
 
-public abstract class PlaylistExtractor extends ListExtractor {
+public abstract class PlaylistExtractor<S extends StreamingService> extends ListExtractor<S> {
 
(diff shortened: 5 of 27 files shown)


mauriciocolli:
> The phrases are provided by the front end.

I don't know about this, this project aims to be standalone, and removing core-parsing breaks it.

IMO, the front-end give an instruction to get something, the back-end handles whatever comes (like, check if it is supported language) and then give back the data (in this case, the `Date` object).

What's your opinion?

Also, I saw that you moved to plurals, but think again, is it worth to let people translate it? We would have to check every single translation as people may confuse it, and I agree with what you said in that issue, this won't change for a while. And initial translation is not a problem, see below.

> Parse the dates for every Info / InfoItem

We can let this one for another time.

> Get the exact upload dates from another source [...] Youtube RSS feed

One disadvantage is that the RSS doesn't seem to respect the `Accept-Language` header, which means that the original titles are returned, and in this case, is a bad thing if the author translate their contents ( as we are going to support subtitles soon).

But I don't know if this downside outweighs the good one, the feed would be a LOT lighter, as the RSS page is much more lightweight, definitely is a thing to consider.

---

There's still something to do: parse relative dates that don't have numbers in it. For example:
 "1 year ago" in `gs`: "Hai un ano"
"1 month ago" in `iw (he)` : "לפני חודש"

##### Lucky for us, the only case that YouTube give us a date not containing a number in some languages, is 1.


My idea is that we initialize NewPipe with a locale (or get the current) and compare if this language is in the list of supported phrases/languages, if not, default it to "en".

And I'm almost finished gathering all the supported languages phrases (80 in the language list), only thing left to do is create the rules and some tweaks (got every case directly from YouTube).

wojcik-online:
> And I'm almost finished gathering all the supported languages phrases

Wow, that sounds like a lot of work. 👍 

> is it worth to let people translate it?

I always assumed that we wouldn't be able to provide all translations, since I probably wouldn't be comfortable with anything other than European languages. But if you can gather all translations from YouTube, then yeah, translating it seems unnecessary.

---

> this project aims to be standalone, and removing core-parsing breaks it.

The way I thought about this:
The language in which to parse the dates is dependent on the language the site was downloaded in and that is dependent on the `Downloader` which is also provided by the front end. It is a simple dependency injection mechanism.

TL;DR I would keep the translated phrases on the front end _or_ put them in a dedicated library to parse the relative dates.

A couple of points on this:
 1. This is not true on the search page, I just realized. Here the language is loaded from `preferences.getString(getString(R.string.content_country_key))`, also provided by the front end. I wonder whether this value should be used in every HTTP request.

  2. In the HTML returned by YouTube, I didn't find any reasonable indication about the language. So the front end will have to provide the language to the parser anyway.

 3. Where would you store the translated phrases if not in the resources on the front end?
I guess gathering the phrases for time units in 80 languages is enough to create a new library that might be useful beyond NewPipe.

 4. I've been experimenting with `PrettyTime` and it works perfectly, although it _only_ supports 33 languages. 
Combining this with the first point, the downloaded language (and the language used to parse the dates) might be different from the language used to display the content. This, of course, breaks down if the author provides translations for the YouTube video names.

---

> There's still something to do

I'll look into it.
Is there a phrase that says "moments ago"? Is the smallest amount "1 second ago"?

mauriciocolli:
Sorry for the huge delay, my time is really short atm, but the only things left to do are some minor things in some very different languages. The ones that use spaces often are quite easy: a split and a check if each word don't repeat is enough to reveal the different words that we can use as keywords.

>Is there a phrase that says "moments ago"? Is the smallest amount "1 second ago"?

The minimum that I could get from YouTube is "1 second ago", but anyway, these are not that frequent as it's only possible to find them through coincidence or some search filters like "Sort by upload date" with "From last hour" (even though the minimum I could get from this is "15 seconds ago", as it seems that YouTube does some checks/operations before "indexing" it).

So the options to where I could put the phrases are:
- Front End (android resource folder)
    - Allows another names for the resources files, which would not bloat the default `string.xml` file.
- Separate gradle module in this repository
- Separate library/project

I'm more inclined to the last two options (what my original goal was) but there's some challenges like where exactly would I put them (classes or resources files) which'd have to still be compatible with pure-java and the front-end (resources loading on android), maybe a separate build type targeting android-library.

Still studying the options, tell me if you have suggestions.

wojcik-online:
It's probably most important to store the translated phrases in a standard way used for i18n.
For example [`ResourceBundle`s](https://docs.oracle.com/javase/8/docs/api/java/util/ResourceBundle.html#getBundle-java.lang.String-java.util.Locale-). They work with pure Java, but are very similar to the android `strings.xml` approach. And they are usually used for i18n (but they kinda suck when [dealing with plurals](https://stackoverflow.com/questions/14326653/java-internationalization-i18n-with-proper-plurals)).

A possible directory structure:
This way the phrases are not buried somewhere deep in the package structure.
```
project
 └─ src/
 |     └─ main/java/org...
 |     └─ test/java/org...
 └─ resources/
       └─ time_units.properties
       └─ time_units_de.properties
       └─ time_units_ ...
```

That's the option I would go for.

mauriciocolli:
Also, there's this to fix:

rawtypec

Because `getStreamExtractor()` is returning a rawtype, we may need to rethink this design. (notice that `getAudioStreams` it's returning a `List `instead of a `List<AudioStream>`, which the type were erased from using the rawtype).

I've already changed the `ServiceList` so we can access the services instances directly, instead of the old enum. I think it's reasonable that the implementation could access its own service from there.

> A possible directory structure:

That's what I had in mind, but I was not certain that this would work in Android. Here some tests/infos:
- Gradle put the files from `resources` to the root of the generated jar
- Importing this to a android project means that those this same files being, now, in the root of the apk:

mauriciocolli:
###### Acidental Ctrl+enter, ignore the previous comment

---

Also, there's this to fix:

rawtypec

Because `getStreamExtractor()` is returning a rawtype, we may need to rethink this design (notice that `getAudioStreams` it's returning a `List `instead of a `List<AudioStream>`, which the type were erased as a result of the rawtype).

I've already [changed the `ServiceList`](https://github.com/TeamNewPipe/NewPipeExtractor/commit/f3865445f5cf6d2ba00b3a91ae96f6c6cfe80c41) so we can access the services instances directly, instead of the old enum. I think it's reasonable that the implementation could access its own service from there.

> A possible directory structure:

That's what I had in mind, but I was not certain that this would work in Android. Here some tests/infos:
- Gradle put the files from `resources` in the root of the generated jar
- The android project will have the same files in the root of the apk:

root

This worked in the device I tested (an Android API 22), but if doesn't work properly in others, we might consider what I suggested before (the android-library build/flavor).

wojcik-online:
> Gradle put the files from resources in the root of the generated jar

The files were just dumped into the root of the apk? That doesn't sound ok. I guess a solution would be to put them in a subdirectory `resources/i18n/time_units.properties` and load them with `getBundle("i18n.time_units")`.

theScrabi:
What is the status of this PR?

wojcik-online:
@theScrabi Sorry for the radio silence.
For the last few weeks I've been using the feed every day and it works very well.
The only thing that is missing is the translation into all languages that @mauriciocolli has been working on. And perhaps some optimization of downloading the channel infos.

Have there been any new relevant developments? If not, I can resolve all conflicts and prepare both PRs to be merged.

theScrabi:
> Sorry for the radio silence.

I wasn't available as well.

mauriciocolli:
I had some time and finally finished it.

Also, I created a new branch (`dev`) with the new Gradle module, so please, change the target of this pull request to there.

And tell me your opinion about the new module vs a separate repository just for that (right now, I'm more inclined to create a new module).

theScrabi:
@mauriciocolli & @wojcik-online there are many conflicts atm. Please fix those.

wojcik-online:
@theScrabi Sure, I'll resolve the conflicts.

@mauriciocolli A module is probably easier to maintain than a separate repository, so it's better. It can be split in the future if there is a need for that.

mauriciocolli:
Finished and already added the time units to the repository: 27f61e049d92c6845a00e82eefa8df1072867d64

Notice that resource bundle don't support UTF-8 so I leave the fix already in place. It also doesn't support arrays in the properties file, fixed by using separators and `\` to "simulate" new lines.

theScrabi:
Sorry to say we need to fix the collisions once again :/

theScrabi:
that-moment-when-you-do-cr-and-you-start-to-think-either-you-are-a-bad-coder-or-the-others-are-just-

theScrabi:
@wojcik-online will you work on this?

PR #63, opened 2018-02-01
Subtitle Extraction Fix

karyogamy:
Took a look into getting subtitles working, found out the current extraction method is only capable of extracting transcripts, which is incompatible with ExoPlayer. Fixed it by reprinting the what is done on youtube-dl:

Available subtitles listing:
`https://video.google.com/timedtext?&type=list&v={video_id}`

Subtitle url:
`https://www.youtube.com/api/timedtext?lang={lang_code}&fmt={vtt, ttml}&name=&v={video_id}`

Also started prototyping it on the client and here's a screenshot:

The only thing left to add is the UI and a track selector for toggle/languages.


karyogamy:
Just realized the current way is correct as well, and is capable of downloading automatic captions, but is missing a few things. Closing this for now, will fix and reopen.

PR #68, opened 2018-02-12
add support for hooktube links

theScrabi:
Add support for hooktube links.


wb9688:
You also need to add it to the Android manifest.

theScrabi:
yea thats on its way

PR #70, opened 2018-02-15
Fix OkHttp

wb9688:

This commit fixes OkHttp support. All unit tests still pass with HttpsUrlConnection. Should I replace the test `Downloader.java` with OkHttp support or should there be 2 test variants (1 with HttpsUrlConnection and 1 with OkHttp)? I'll also create a PR to NewPipe to add OkHttp support and use the new NewPipeExtractor API.

### DIFF
--- a/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeStreamExtractor.java
+++ b/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeStreamExtractor.java
@@ -678,7 +678,7 @@ public class YoutubeStreamExtractor extends StreamExtractor {
 
             String helperPattern = "(var "
                     + helperObjectName.replace("$", "\\$") + "=\\{.+?\\}\\};)";
-            helperObject = Parser.matchGroup1(helperPattern, playerCode);
+            helperObject = Parser.matchGroup1(helperPattern, playerCode.replace("\n", ""));
 
 
             callerFunc = callerFunc.replace("%%", decryptionFuncName);


theScrabi:
If it was easy to do please add tests for both.

wb9688:
>If it was easy to do please add tests for both.

I don't know enough about Gradle for that, maybe @mauriciocolli knows how to do that. If not, should I just use OkHttp for the tests?

theScrabi:
I guess you can introduce it.

mauriciocolli:
I just figured why OkHttp didn't work right way, the cause is this: [Downloader.java#L118..L120](
https://github.com/TeamNewPipe/NewPipeExtractor/blob/master/src/test/java/org/schabi/newpipe/Downloader.java#L118..L120).

You can notice that this is removing all the line separators, instead, it should be just copying the bytes from the InputStream, which is probably what OkHttp is doing.

This _may_ break other things in NewPipe, but it's for the better. And I really like the idea of using OkHttp as the downloader as we'll have some nice features like retry (useful in unstable networks) and better yet, we will know that we are using the same implementation across all devices (which _may_ not be true for the java implementation because of API version, java version, manufacturer, etc.)

theScrabi:
> You can notice that this is removing all the line separators

I see. Well than newpipe needs to be able to work without line separators.

wb9688:
@theScrabi: Uhmm… that's what I fixed in this PR ;)

theScrabi:
> @theScrabi: Uhmm… that's what I fixed in this PR ;)

I see :D.

theScrabi:
Why are the unit tests failing?

wb9688:
@theScrabi: On my PC, they aren't failing. On Travis, this is just one of
these times it's incorrectly failed with the search extraction (it expected
a channel, but it was a video), which can't be a result of this code, since
this doesn't touch that part at all.

PR #73, opened 2018-02-26
HLS Livestream Extraction

karyogamy:
- Added HLS url extraction.
- Changed subtitles format to TTML instead of VTT because of overlapping issue on auto-generated captions.
- Changed dash and hls url to return nonnull empty strings if not available.
- Fixed Youtube stream info extration to correctly report stream type.
- Removed live stream exceptions.

### DIFF
--- a/src/main/java/org/schabi/newpipe/extractor/search/SearchResult.java
+++ b/src/main/java/org/schabi/newpipe/extractor/search/SearchResult.java
@@ -44,23 +44,10 @@ public class SearchResult {
         this.errors = Collections.unmodifiableList(new ArrayList<>(errors));
     }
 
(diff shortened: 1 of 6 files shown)


wb9688:
SoundCloud also has HLS (at least in API v2, I'm not sure what we're using
atm for extracting the streams), but I'm not sure how useful that'd be.

karyogamy:
Since SoundCloud [doesn't support livestreaming](https://soundcloudcommunity.com/podcasting-230062/does-soundcloud-offer-live-stream-7370509), HLS manifests is not really that useful at the moment. However, they do support adaptive track selection (according to the network), so it could be useful in the future. Just to keep a record, do you mind provide an example url or the support docs for getting hls?

wb9688:
In SoundcloudStreamExtractor on line 129, you could see how it gets the MP3
URL, in the same JSON file there's also a HLS URL.

PR #74, opened 2018-02-28
fix NPE when captionsArray is null

forsythe:
I was trying to watch [this video](https://www.youtube.com/watch?v=OyLafqSSJF8), but it kept throwing NPE at this line. It would also throw NPE for other videos in the same [series](https://www.youtube.com/watch?v=4R0KwjwYrkE). Need to check if captionsArray is null before checking its size.

### DIFF
--- a/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeStreamExtractor.java
+++ b/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeStreamExtractor.java
@@ -739,6 +739,9 @@ public class YoutubeStreamExtractor extends StreamExtractor {
         // todo: use this to apply auto translation to different language from a source language
         final JsonArray autoCaptionsArray = renderer.getArray("translationLanguages", new JsonArray());
 
+        // captionsArray object may be null, e.g. https://www.youtube.com/watch?v=OyLafqSSJF8
+        if (captionsArray == null) return Collections.emptyList();
+		
         // This check is necessary since there may be cases where subtitles metadata do not contain caption track info
         // e.g. https://www.youtube.com/watch?v=-Vpwatutnko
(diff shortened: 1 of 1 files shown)


mauriciocolli:
See the [line 738](https://github.com/TeamNewPipe/NewPipeExtractor/pull/74/files#diff-eabbe0ba204823c144fd4551d4ba4f43R738), notice that it's trying to get the array providing a default, that means that if the video doesn't have captions, it will be caught by the check below ([line 745](https://github.com/forsythe/NewPipeExtractor/blob/d008ec926f2fd9c9b0fcdf0b17681fd78d23402a/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeStreamExtractor.java#L747..L748)) because of an empty list.

Also, did you test NewPipe with the latest version of the extractor? Probably you are using it with a previous one, since this changes were made recently.

forsythe:
Yes, I'm testing NewPipe with the latest version of the extractor (i.e. after the refactoring). I noticed the check at line 745, but I am consistently able to generate the NPE at the following line:
`final int captionsSize = captionsArray.size();`

mauriciocolli:
I should be more specific, is it from the latest extractor commit? This behavior doesn't show up in the latest one.

forsythe:
The version after the refactoring is the version currently used by the latest NewPipe on fdroid right?

mauriciocolli:
I don't remember exactly, but probably no.

I'm closing this because there's no way that `captionArray` could be null, given that we are providing a default value (that is used [in case of nonexistent or wrong type](https://github.com/mmastrac/nanojson/blob/ac30675cb8a2669bd67eaae7727ca35fc7c0b790/src/main/java/com/grack/nanojson/JsonObject.java#L69-L76)).

PR #75, opened 2018-03-08
refactor InfoItemParser using pages

theScrabi:
refactor InfoItemExtractor using pages


theScrabi:
@mauriciocolli what is the status on this? Did you fix the bugs?

mauriciocolli:
> Did you fix the bugs?

That I mentioned in that comment, yes.

But I'm still working on it (thinking about some API changes and some other things), so don't merge yet.

theScrabi:
All right :)

karyogamy:
Skimmed through it, so far so good. I'll go a little more in-depth after @mauriciocolli pushes his changes.

I've also looked into comments extraction on Youtube and it seems to be more complicated than all other extractions we are doing right now. Here's a working [Python source](https://github.com/egbertbouman/youtube-comment-downloader/blob/master/downloader.py). 

In short, to extract comments, we need to expand the current `Downloader` to also support `POST` requests. This may impact how list extraction passes the next url info since we now need to take form-data into account. In particular, Youtube comments requires a static `XSRF_TOKEN` as well as a `data-token` for each page, both are part of the form-data, which cannot be passed using just a string url. With this in mind, maybe we should consider passing around a data class rather than a String on `getNextPageUrl()`?

theScrabi:
Well you may know that i am not such a great fan of commend support, although I would not go against it.
However please lets not add commend support in this commit. This is only ment for refactoring current funktionallity

karyogamy:
Yeah, I see :ok_hand:. It's probably a good idea to do that in a separate PR just for `POST` support, and then another PR for comments.

mauriciocolli:
@karyogamy or we could use the GET request that the mobile site does to get the comments, it's very lightweight (20KB to the page + 4KB comment's request) so I think we can work with it.

But I agree with you, we need to support more advanced request and response because eventually, some service will need it exclusively (a response code or message, for example).

karyogamy:
@mauriciocolli Nice find =D I gave it a shot but it seems like we still need to keep the cookies for the mobile link to work. 
Nonetheless, right now we should probably leave comments for later, block new features to the extractor after this PR and focus on getting the extractor to work with the frontend so we can release soon.

PR #87, opened 2018-05-19
Fixed long links in description.

theScrabi:
Fixed longer links not working by changing their href to the link they should redirect to.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
@@ -24,6 +24,7 @@ import org.schabi.newpipe.extractor.utils.Utils;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.io.IOException;
+import java.io.UnsupportedEncodingException;
 import java.util.*;
 
 /*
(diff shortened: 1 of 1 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -151,12 +152,44 @@ public String getThumbnailUrl() throws ParsingException {
...
     }
 
+    private String fixDescriptionLinks(String description) throws ParsingException, UnsupportedEncodingException {
+        boolean everythingIsFine = true;
+        String descriptionE = java.net.URLDecoder.decode(description, "UTF-8");
+        Parser.getLinksFromString(descriptionE);
+        int endlessloop = 0;
+        while(everythingIsFine) {
+            try {
+                String[] FirstCut = description.split("<a href=\"",2);
```
theScrabi: variable names are always small in java

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -151,12 +152,44 @@ public String getThumbnailUrl() throws ParsingException {
...
+                if(FirstCut.length==1) {
+                    everythingIsFine = false;
+                }
+
+                String Beginning = FirstCut[0];
+                String[] SecondCut = FirstCut[1].split("\"",2);
```
theScrabi: what are you going todo if ther is a `\\"` in the url which gets encoded into a `\"`

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -151,12 +152,44 @@ public String getThumbnailUrl() throws ParsingException {
...
+    private String fixDescriptionLinks(String description) throws ParsingException, UnsupportedEncodingException {
+        boolean everythingIsFine = true;
+        String descriptionE = java.net.URLDecoder.decode(description, "UTF-8");
+        Parser.getLinksFromString(descriptionE);
+        int endlessloop = 0;
+        while(everythingIsFine) {
+            try {
+                String[] FirstCut = description.split("<a href=\"",2);
+                if(FirstCut.length==1) {
+                    everythingIsFine = false;
```
theScrabi: what does everythingIsFine mean? when can this case occure.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -151,12 +152,44 @@ public String getThumbnailUrl() throws ParsingException {
...
+                    String LinkToBeFixed = SecondCut[0].split("q=")[1].split("&amp")[0];
+                    String Link = java.net.URLDecoder.decode(LinkToBeFixed, "UTF-8");
+
+                    description = Beginning + "<a  href=\"" + Link + "\"" + End; //I'm inserting a double space between "<a" and "href" here so the next cut doesn't cut here.
+                } else { //Timestamps and other links to youtube videos are processed here
+                    description = Beginning + "<a  href=\"" + SecondCut[0] + "\"" + End;
+                }
+            } catch (ArrayIndexOutOfBoundsException | UnsupportedEncodingException end) {
+                //this means we have run out of Links because there are no more <a href=" to split the text in so the Array has only one Element or that something else went wrong.
+                endlessloop = endlessloop + 1;
```
theScrabi: This is not good java style

theScrabi:
I think since you are parsing an html document the best solution was to use our jsoup parser, and that user java Uri to parse the url. You will save yourself some code, and it is more likely to not brake :)

Somethingweirdhere:
@theScrabi 
Alright, sorry for the code quality, it's the first big project I am working on so I never learned best/good practices before. Also sorry for the mess I made in Git, I am still getting used to it.

Now, about the code itself:

- I have made sure all the variables are small
- \\" shouldn't matter at all. The \ in my code only escapes the " that I am splitting, as far as I know any \ in the URL itself don't matter, after all the https://www doesn't get messed up.
- I have given the variable a more fitting name, as long as it is true the loop continues and looks for the next link to fix.
- Alright, I hope that's better.

I have taken a look at the jsoup parser before but haven't been able to extract the querry parameter we need with it and sometimes it ended up in endless loops. I found this approach easier, but I can try using jsoup again if it's cleaner.

PR #94, opened 2018-07-01
[WIP] deep refactoring search

theScrabi:
This PR will ulitmtively removes the legecy SearchEngine and replace it with a propper SearchExtractor. Also It refactors URLIdHandlers. They are short UIHs (UrlIdHanlder) and QIHs (QuerryIdHandler) now. These  are Immutable and must be created by UIHFactories QIHFactories which must be overwritten by the service. QIHs can store more information than just the Id. They can also accept sort and filter parameter, which is why they are not simply called UIHs. The QIH where necessary to for supporting SearchEngineExtractor. They will also make it possible for other ListExtracotrs to let the content be sorted in the future.


extractor/src/main/java/org/schabi/newpipe/extractor/uih/UIHandler.java
```diff
@@ -0,0 +1,31 @@
+package org.schabi.newpipe.extractor.uih;
+
+import java.io.Serializable;
+
+public class UIHandler implements Serializable {
```
karyogamy: Please don't abbreviate class names. Since this is a library, we should try to make names clear, even if they are long so they don't confuse the users. This is especially the case for `UIHandler`, I thought we are doing something graphical in the extractor when I saw this initially.

In case of `UrlIdHandler`, I think we should think of a different name for it, such as `LinkHandler`, because this now also contain other data. Similarly, we can simply drop the `Handler` suffix since now the class is just a pojo and doesn't *handle* any logic. 

This way, `UIHandler` class becomes `Link` class, `ListUIHandler` to `ListLink` and `SearchQIHandler` to `SearchQuery`. What do you think?

extractor/src/main/java/org/schabi/newpipe/extractor/uih/UIHFactory.java
```diff
@@ -23,52 +23,46 @@
  * along with NewPipe.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-public abstract class UrlIdHandler {
+public abstract class UIHFactory {
```
karyogamy: Ideally, I would keep this as an interface and use default method for `fromUrl` and others which can be used by the implementor. Always go for composition over inheritence when the code is still simple to refactor ;)

This also means we might want to do to same to `UIHandler` and make it implement a `Linkable` that only has the `originalUrl` and `url` and `Identifiable` that contains the `id`. This way, `ListUIHandler` no longer have to extend `UIHandler`, and instead implements `Linkable`, `Identifiable` and `Filterable`.

This gives us more flexibility later down the road and also helps addressing potential misuses where the `id` in `ListUIHandler` is used as a holder for the search query string in `SearchQIHandler`, which would be decoupled from `ListUIHandler` and instead implements `Linkable` and `Filterable` and the search query becomes a field.

I understand this is probably a lot of work, so I'm fine with keeping the inheritance structure for now. However, I recommend having it done right now as there will likely be a time when more complex url uses arise when simple inheritance will not be enough.

theScrabi: > I understand this is probably a lot of work, so I'm fine with keeping the inheritance structure for now. However, I recommend having it done right now as there will likely be a time when more complex url uses arise when simple inheritance will not be enough.

Yea your ideas are quite good, but the refactor took me almost two months and now I'm already done. I don't want to throw all over and start again.

However we should consider something like this, because once we pull off multisservice support and everyone starts writing supporting services there is probably no turning back (or it will be hart).

extractor/src/main/java/org/schabi/newpipe/extractor/uih/UIHandler.java
theScrabi: I agree. I will rename them.

PR #96, opened 2018-07-25
Enable opus format 251

Bleuzen:
Enabled opus format 251 again
because format 249 and 250 are already enabled and work great. So 251 should also work, since it's the same just with higher bitrate.


wb9688:
Close in favor of #99?

Bleuzen:
@wb9688 
Hm, in my opinion we don't need the old comments and the webm things anymore.
But in the end it is the same change in code. So it doesn't matter which one gets merged.

wb9688:
Yep, I only saw your PR later and I agree with you, so I'll close mine.

PR #98, opened 2018-07-27
[WIP] Minified url of youtube videos NewPipe#1556

mjsir911:

I haven't gotten an opportunity to test this out due to not having a working android development environment yet, will get on that ASAP.

I don't see any pressing issues with change and I can't imagine how it would break anything.

I created the initial issue for this on the wrong repo: TeamNewPipe/NewPipe#1556

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/Info.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/Info.java
@@ -28,6 +28,13 @@ public abstract class Info implements Serializable {
      * @see Extractor#getOriginalUrl()
      */
     private final String originalUrl;
+    private final String shortUrl;
+    /**
+     * A shortened version of the Url pointing at the same media
+     * e.g. Youtube https://www.youtube.com/watch?v=RER5qCTzZ7 &gt; https://youtu.be/RER5qCTzZ7
+     *
+     * @see LinkHandler#getShortUrl()
+     */
     private final String name;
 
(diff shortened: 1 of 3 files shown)


wb9688:
I wouldn't do that, since the resulting URL would take a few ms more to load and Google will track that you've used youtu.be and there's no reason to do this.

mjsir911:
Hi,

Would there be a more fitting place to shorten the youtube URL when the user actually has to interact with it through the `share` menu? See TeamNewPipe/NewPipe#1556 for more details, but i'm happy to help if i'm given a direction to go.

wb9688:
Maybe you could create a new getShortenedUrl() function in StreamExtractor,
then implement it in YoutubeStreamExtractor and SoundcloudStreamExtractor
(just return null there, since SoundCloud doesn't have them). Then we also
have to add something to NewPipe to optionally share that minified link.

Op vr 27 jul. 2018 18:57 schreef Marco Sirabella <notifications@github.com>:

mjsir911:
Sorry, I won't be able to accomplish this PR successfully.

I've looked into modifying `LinkHandlerFactory` to pass the short url to `LinkHandler`, which then passes it to `Info`, which can be accessed through `StreamInfo` in NewPipe's `VideoDetailFragment` `shareUrl()`

Thanks for trying to help, but the flow of the codebase is a bit too complex for me to file a quick PR that doesnt break more than it fixes

PR #99, opened 2018-07-28
Fixed 160 kB/s Opus

wb9688:

PS: @theScrabi can I have write access to NewPipeExtractor please?


Bleuzen:
This is the same as my PR here: https://github.com/TeamNewPipe/NewPipeExtractor/pull/96
The only difference is that I also removed old comments which we don't need anymore.
@theScrabi should decide which one to merge.

PR #108, opened 2018-09-07
Fix signature extraction parse exception

karyogamy:
Copy of [fix PR on youtube-dl](https://github.com/rg3/youtube-dl/pull/17464) because of I'm too dumb to figure out a solution and decided to wait it out =P

The stream extraction tests are now passing.

Credits and many thanks to @Khang-NT.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
@@ -698,8 +698,10 @@ public class YoutubeStreamExtractor extends StreamExtractor {
             }
             String playerCode = downloader.download(playerUrl);
 
-            decryptionFuncName =
-                    Parser.matchGroup("([\"\\'])signature\\1\\s*,\\s*([a-zA-Z0-9$]+)\\(", playerCode, 2);
+            decryptionFuncName = Parser.matchGroup(
+                    // Look for a function with the first line containing pattern of: [var]=[var].split("")
+                    "(\\w+)\\s*=\\s*function\\((\\w+)\\)\\{\\s*\\2=\\s*\\2\\.split\\(\"\"\\)\\s*;",
+                    playerCode, 1);
 
             String functionPattern = "("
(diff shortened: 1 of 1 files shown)


wb9688:
Looks OK to me. Can we release soon after this is merged? (And can F-Droid please be fast for once? )

karyogamy:
Yeah, we should release ASAP. Though the solution is probably temporary and can be easily patched. We might need to find a better way to match this soon.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -698,8 +698,10 @@ private String loadDecryptionCode(String playerUrl) throws DecryptException {
             }
             String playerCode = downloader.download(playerUrl);
 
-            decryptionFuncName =
-                    Parser.matchGroup("([\"\\'])signature\\1\\s*,\\s*([a-zA-Z0-9$]+)\\(", playerCode, 2);
+            decryptionFuncName = Parser.matchGroup(
+                    // Look for a function with the first line containing pattern of: [var]=[var].split("")
+                    "(\\w+)\\s*=\\s*function\\((\\w+)\\)\\{\\s*\\2=\\s*\\2\\.split\\(\"\"\\)\\s*;",
```
theScrabi: Is this the solution form youtube-dl?

theScrabi: I see.

avently:
After extractor failed to decrypt a stream I went on NewPipe git and found this PR and patched local extractor lib. That's insane that you don't have normal embed player one year later:)
I'm sorry, guys, it was painful year for you...

The-Demon-Prosecutor:
Hi @karyogamy, youtube-dl has now merged another regex pattern which works as well. Thought you would be interested to know. It can be found in their latest release or alternatively [here](https://github.com/rg3/youtube-dl/commit/9a47fa35dd9dd2d53f4d3f088811ea29295991e5).

karyogamy:
@The-Demon-Prosecutor Thanks, I've created a new PR with the update. See #109.

PR #114, opened 2018-09-15
Localization

theScrabi:
This is my implementation of a proper localization.
There is a default localization that will be created when you call `NewPipe.init()`. However you can override it by creating extractors manually.

@mauriciocolli @karyogamy @nv95 @kapodamy please tell me what you think of this solution.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/Downloader.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/Downloader.java
@@ -1,6 +1,7 @@
 package org.schabi.newpipe.extractor;
 
 import org.schabi.newpipe.extractor.exceptions.ReCaptchaException;
+import org.schabi.newpipe.extractor.utils.Localization;
 
 import java.io.IOException;
 import java.util.Map;
@@ -32,11 +33,11 @@ public interface Downloader {
      * but set the HTTP header field "Accept-Language" to the supplied string.
      *
      * @param siteUrl  the URL of the text file to return the contents of
-     * @param language the language (usually a 2-character code) to set as the preferred language
+     * @param localization the language and country (usually a 2-character code for each)
      * @return the contents of the specified text file
      * @throws IOException
      */
-    String download(String siteUrl, String language) throws IOException, ReCaptchaException;
+    String download(String siteUrl, Localization localization) throws IOException, ReCaptchaException;
 
     /**
      * Download the text file at the supplied URL as in download(String),
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/Extractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/Extractor.java
@@ -3,6 +3,7 @@ package org.schabi.newpipe.extractor;
 import org.schabi.newpipe.extractor.exceptions.ExtractionException;
 import org.schabi.newpipe.extractor.exceptions.ParsingException;
 import org.schabi.newpipe.extractor.linkhandler.LinkHandler;
+import org.schabi.newpipe.extractor.utils.Localization;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -16,17 +17,19 @@ public abstract class Extractor {
     private final StreamingService service;
 
     private final LinkHandler linkHandler;
+    private final Localization localization;
 
     @Nullable
     private boolean pageFetched = false;
     private final Downloader downloader;
 
-    public Extractor(final StreamingService service, final LinkHandler linkHandler) {
+    public Extractor(final StreamingService service, final LinkHandler linkHandler, final Localization localization) {
         if(service == null) throw new NullPointerException("service is null");
         if(linkHandler == null) throw new NullPointerException("LinkHandler is null");
         this.service = service;
         this.linkHandler = linkHandler;
         this.downloader = NewPipe.getDownloader();
+        this.localization = localization;
         if(downloader == null) throw new NullPointerException("downloader is null");
     }
 
@@ -99,5 +102,10 @@ public abstract class Extractor {
 
     public Downloader getDownloader() {
         return downloader;
+    }
+
+    @Nonnull
+    public Localization getLocalization() {
+        return localization;
     }
 }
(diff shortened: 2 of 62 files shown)


mauriciocolli:
I was working on this right now to complete the dates parsing, right on time.

The thing is, we cannot enforce a language in the services so I think a more appropriated term would be `set/getPreferredLocalization`.

With this in mind, instead of passing the preferred localization to each extractor, we could create a `getLocalization` function in the `StreamingService` class and override in each service implementation (usage would be `getService().getLocalization()`). Each one of them would compare it with a list of its supported languages/countries, and if it isn't, a well-supported default (definitely english) would be returned. 

This would improve the parsing of things that depend on it and make it more consistent across all the extractors (e.g. the same title in the search results and when viewing the details). We would only add to the list languages that we already tested and covered all the different cases when parsing.

What do you think?

theScrabi:
> The thing is, we cannot enforce a language in the services so I think a more appropriated term would be `set/getPreferredLocalization`.

I agree. 

> With this in mind, instead of passing the preferred localization to each extractor, we could create a `getLocalization` function in the `StreamingService`

I thought about this at the beginning as well but than I decided because of this:
I had a more functional approach in mind, by means when the extractor got created the context will not change. This inflicts ListExtractors. Imagine you scroll through a search result, change the language and keep scrolling, when the next page has to be loaded the outcome might be undefined.

Also if we could select the language per extractor we would be more flexible. I had the time parser as an example in mind. Imagine someone set the language to Spanish, but than the feet would be loaded using English we could make the time parser only  use one language. @mauriciocolli I don't know about that though. 


> Each one of them would compare it with a list of its supported languages/countries, and if it isn't, a well-supported default (definitely english) would be returned. 

I would not define a default language since the services will already do that. Imagine you use a French service that should show French television, the default language might most likely be French.

> We would only add to the list languages that we already tested and covered all the different cases when parsing.

Sounds like a lot of work and maintaining I'd prefer to avoid. Especially with multi service in mind. However testing against languages/countries is a good idea.

mauriciocolli:
> This inflicts ListExtractors. Imagine you scroll through a search result, change the language and keep scrolling

Not relevant to this problem, entirely a front-end issue (a solution: detect a configuration change, clear the cache and load everything again, what else can we do? Also, a rare edge-case).


> Also if we could select the language per extractor we would be more flexible

But why? We _must_ be consistent across all extractors, we cannot use a per-extractor basis on this.


> Keep in mind thought that you must still set a NewPipe wide Localization which will by default be passed down to each extractor.

Falls on what I was talking about in my previous comment, it's a preferred localization, we cannot be sure that it has. For example, SoundCloud have much fewer languages available than YouTube.

Each service would check if it was supported before using it, ensuring that it'd work.


> I would not define a default language since the services will already do that.

This doesn't help as we cannot be sure that it'll parse correctly if a language is not supported (for example, some subtle changes between different languages).

And of course, the default language would be the one that we know it does. If the service is only in one language, that one would be returned.


> Sounds like a lot of work and maintaining I'd prefer to avoid. Especially with multi service in mind.

See below.

---

At the end of the day, the only thing that would change between each localization would most likely be a (creator optional) title and description. I'm not even sure if SoundCloud supports this.

I think the solution that I talked about is better: we _do not_ need to support every language that YouTube does, only the bigger ones. The title/description of languages with fewer speakers will most likely *not* be translated to english (or whatever language), so the original title will be fetched, and things will be parsed correctly.


Let's trying doing that, if we eventually decided that supporting even the fewer languages is too much of a maintenance burden, we could just force the english (or the main language) site and be sure that it'll work (like `youtube-dl` does).

TheAssassin:
Please beware that I don't know very much about the extractor code wise, the following text was written on the basis of my discussion with @the-scrabi who explained some backgrounds to me, and the comments in this PR.

Regarding how the extractor's language is set, it seems better to keep that in the extractor's constructor to me. That's more flexible. If someone opts to instantiate an extractor manually, they can simply pass their preferred language. (TL;DR: that's also required for some method I suggest below.)

> We would only add to the list languages that we already tested and covered all the different cases when parsing.

It doesn't seem very efficient to implement language specific parsers for all languages, like @theScrabi said. If a format changes, that change must be applied in the code quickly. This can't be guaranteed, especially if the list is long. Testing could be automated, sure, but that doesn't scale well, especialy if more services are added. We must make sure that parsing errors are recognized, and provide a fallback solution.

Regarding this "edge case": I think that if the language changes for instance, old extractors should be replaced by new ones if possible, or the old ones should be kept until they're done. Users might experience some data with the old language, sure, but I guess it's acceptable for them, as the language change will take effect shortly. I'm with @mauriciocolli here: this is not an issue of the extractors, but the frontend using them. If e.g., a language changes, the frontend should either live with the consequences (i.e., mixed language content until all caches expire), or fix it itself. Extractor instances should be "immutable" language wise.

Implementing things like date parsers for _every_ language is too much work and also too much maintenance. Also, removing languages which you think "aren't popular" is a bad idea. We don't have statistics (for good reasons) about which languages are used by how many users. It feels wrong to remove features like support for specific languages. I'd rather implement a fallback solution, _even if_ there is some minimal overhead.

My suggestion is to maybe implement a few "known good" metadata parsers (short list so that it's maintainable, perhaps <10?). If the language is not supported, all the parsers could be tried (only if parsing errors could be recognized; some language's date formats might be quite close).
If parsing _really_ is not possible (i.e., none of the existing parsers returns data), we could simply fetch the English version of the same page and parse the metadata from there (e.g., by implicitly creating a new instance of the extractor using the English locale _within_ the localized extractor instance, calling the right method there, and returning the result). As @mauriciocolli already implemented quite some amount of parsers, all of them could be kept, and for future services, writing a proper parser for all languages is not required either.
Sure, there _is_ some overhead, but it's rather small (only the page content). This should be signalized to the user by providing some sort of `[ ! ]` button next to some not-so-well-supported language in the language selection, opening a popup informing about the overhead, and asking for contributions for better date parsers. NewPipe is streaming media content, after all, a few kB overhead shouldn't hurt.

theScrabi:
@mauriciocolli I am going to merge this, however please feel free to take my solution, and implement something on top of it :)

PR #119, opened 2018-09-26
Misc utils

kapodamy:
This include:
* Mp4 DASH reader/writter
* WebM reader/writter
* a subtitle converter for **Timed Text Markup Language** v1 and **TranScript** (v1, v2 and v3)
* SharpStream to wrap IntputStream and OutputStream in one interface
* custom implementation of DataInputStream

- [✔] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [✔] I did test the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).
- [✔] I agree to ASAP create a PULL request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) for making in compatible when I changed the api.


kapodamy:
well, XmlPullParser is only available in the frontend side, later i look for a workaround

theScrabi:
I thought pull parser was available in backend as well. Let me have a look.

theScrabi:
@kapodamy I just noticed that this code uses Android native Elements. However Android is not allowed in the extractor, only pure java.
This is why tests are failing.

I know I said earlier that it was a good Idea to put these parts into the extractor, however now I am not sure about it any more. Please put them into a separate module in the fronted. A good name for a package to put these source files would be:
`org.schabi.newpipe.utils.streams`.

theScrabi:
I see, however i still think this would fit better into frontend after reading into this.

Is it a lot of work to put it back?

ravig252525:
YouTube home page scrap please

kapodamy:
@ravig252525 what?

PR #132, opened 2019-01-12
refactored YouTube-linkHandler

connectety:

This PR will replace most of the regex and string contains, which are used inside the Youtube-linkHandler, which allows for better URL-Handling. To do this, the PR heavily depends on javas URL-Class and adds the three utility-methods isYoutubeURL(), isYoutubeALikeURL() & getQueryValue(). Most of the existing regex is still used to check whether or not the ID, which should be returned, is correct, although this could be changed easily.

Things to consider/do:
- Both getID and onAccept should make sure an ID can be extracted without errors. To reduce the amount of duplicate code, onAccept should call getID. Because there only are few exceptions to this rule (the linkHandler for trending), onAccept should return if getID runs with the given arguments without any exception and this behavior should be given in ListLinkHandlerFactory, if not overridden.
- ~~At the moment some tests make sure IDs are extractable from URLs without a protocol. This behavior is not implemented yet in this PR, because I personally do not like the solutions I figured for this. I guess NewPipe gets either called from the browser (which corrects URLs) or from other apps (which should only have valid URLs) and each time we add error-correction to the extractor, NewPipe gets bigger and slower. Is this feature being used?~~


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java
```diff
@@ -28,6 +30,42 @@
...
+    }
+
+    public static boolean isYoutubeALikeURL(URL url) {
+        // make sure its http or https
+        if (!isHTTP(url))
+            return false;
+
+        // make sure its a known youtube url
+        String host = url.getHost();
+        return host.equalsIgnoreCase("youtube.com") || host.equalsIgnoreCase("www.youtube.com")
```
theScrabi: You could add support for invidious.io as well.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeStreamLinkHandlerFactory.java
```diff
@@ -49,78 +41,139 @@ public static YoutubeStreamLinkHandlerFactory getInstance() {
...
         }
-        if (lowercaseUrl.contains("hooktube")) {
-            if (lowercaseUrl.contains("&v=")
-                    || lowercaseUrl.contains("?v=")) {
-                return Parser.matchGroup1("[?&]v=" + ID_PATTERN, url);
+
+        // using uppercase instead of lowercase, because toLowercase replaces some unicode characters
+        // with their lowercase ASCII equivalent. Using toLowercase could result in faultily matching unicode urls.
+        switch (host.toUpperCase()) {
+            case "WWW.YOUTUBE-NOCOOKIE.COM": {
```
theScrabi: This will require a change in the androidmanifest of the frontend.

connectety: I will ad the url to manifest later, but originally i added this code to get throught the test:
https://github.com/connectety/NewPipeExtractor/blob/19288c1456d667644e3d0ec12bf15d3d21b88447/extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeStreamLinkHandlerFactoryTest.java#L71
https://github.com/connectety/NewPipeExtractor/blob/19288c1456d667644e3d0ec12bf15d3d21b88447/extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeStreamLinkHandlerFactoryTest.java#L93

theScrabi: all right :)

connectety: did so in https://github.com/TeamNewPipe/NewPipe/pull/2092

PR #145, opened 2019-03-09
Basic PeerTube support (without P2P)

yausername:

* No P2P (video playback is via direct https link).
* No multi-instance support. Cannot use multiple instances simultaneously.
* Pre selected default instance - http://framatube.org/
* No login/accounts.


extractor/src/main/java/org/schabi/newpipe/extractor/ServiceList.java
```diff
@@ -37,17 +37,19 @@ private ServiceList() {
...
      * and give it the next free id.
      */
-    private static final List<StreamingService> SERVICES = unmodifiableList(
-            asList(
+    private static final List<StreamingService> SERVICES = Collections.unmodifiableList(
+            Arrays.asList(
                     YouTube = new YoutubeService(0),
                     SoundCloud = new SoundcloudService(1),
-                    MediaCCC = new MediaCCCService(2)
+                    PeerTube = new PeertubeService(2),
```
theScrabi: This move is Illegal, as it will mess up the database. Please never change the order of the numbers.

extractor/src/main/java/org/schabi/newpipe/extractor/stream/Stream.java
```diff
@@ -1,13 +1,14 @@
 package org.schabi.newpipe.extractor.stream;
 
-import org.schabi.newpipe.extractor.MediaFormat;
-
 import java.io.Serializable;
 import java.util.List;
 
+import org.schabi.newpipe.extractor.MediaFormat;
+
 public abstract class Stream implements Serializable {
```
theScrabi: Could you plese add a jdoc documentation to this file, and explain the use of torrenturl :)

yausername: done.
added example of the optional parameter `torrentUrl` (https://webtorrent.io/torrents/big-buck-bunny.torrent), supplied if the stream can be played via bittorrent.

Serkan-devel:
Has the peertube api actually been changed since?

theScrabi:
You've got my go. Code looks fine, and I didn't find anything where the api got broken :)

TobiGr:
The validation process looks cleaner now :+1:

TobiGr:
Yeah. Finally :tada: Thank you :heart:

PR #159, opened 2019-04-25
Improve thumbnail quality for Soundcloud service

ritiek:
The thumbnail URL provided when using the Soundcloud API defaults to 100x100 pixels image. It seems we could also fetch for better resolution thumbnails by replacing a small part in the URL. This stackoverflow post contains some valuable information on Soundcloud thumbnail quality: https://stackoverflow.com/a/13989369/.

I've added changes to the code to replace the part of the URL to fetch a 400x400 pixels thumbnail image. I didn't use the 500x500 pixels image because for some reason the part of the images were cut from the sides when displayed in NewPipe while 400x400 pixels image didn't have this problem.

Also, some Soundcloud tracks have the `artwork_url` key as null in the API response. Previously, the thumbnail for such tracks wouldn't show up in NewPipe. This PR will fallback to the `avatar_url` if `artwork_url` is `null` (this is what Soundcloud does when accessing it through a web browser, so these changes are consistent with that observed with a web browser).

----------

Some before-After NewPipe screenshots (click to zoom):
(At left: Before, At right: This PR)

 
 
 
----------

- [No breaking changes] I agree to ASAP create a PULL request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) for making in compatible when I changed the api.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudChannelInfoItemExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudChannelInfoItemExtractor.java
@@ -24,7 +24,9 @@ public class SoundcloudChannelInfoItemExtractor implements ChannelInfoItemExtrac
 
     @Override
     public String getThumbnailUrl() {
-        return itemObject.getString("avatar_url", "");
+        String avatarUrl = itemObject.getString("avatar_url", "");
+        String avatarUrlBetterResolution = avatarUrl.replace("large.jpg", "crop.jpg");
+        return avatarUrlBetterResolution;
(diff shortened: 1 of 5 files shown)


ritiek:
Some tests are failing but those don't seem to be related to this PR.

theScrabi:
Thanks :)

Although I think the big issue here right now is that soundcloud support seems broken right now.

TobiGr:
@theScrabi What do you mean by soundcloud is broken? I just fixed playlist extraction in 0d09a9fd61b1a02d858a6522dd5c09b54bb5a037 and 93d4299f075c3defcb465fc8709594017f2216cf should have fixed the other problems. We might want to add those commits to the next hotfix release.

PR #160, opened 2019-04-26
Fixed youtube subscription import: ignore ones with invalid url and keep ones with empty title.

Stypox:
- [no breaking change] I agree to ASAP create a PULL request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) for making in compatible when I changed the api.

Now when the YoutubeSubscriptionExtractor parses the YouTube subscription_manager file, it ignores channels that have no title, instead of throwing an error: that file can sometimes natively contain channels with no title (i.e. deleted channels).

I tested the changes inside the NewPipe app against [the file](https://gist.githubusercontent.com/wolttam/d56586d2c677565071c4e935812d39c9/raw/dba803bc29f694afaf6bc51faaeffcdf2f5866bf/subscription_manager) provided in TeamNewPipe/NewPipe#2226. I also added a unit test.

I left untouched all the other "document has invalid entries" exceptions, since they are not involved in the empty-title issue. I think this is ok.

Fix TeamNewPipe/NewPipe#2226

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSubscriptionExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSubscriptionExtractor.java
@@ -63,20 +63,10 @@ public class YoutubeSubscriptionExtractor extends SubscriptionExtractor {
             String title = outline.attr("title");
             String xmlUrl = outline.attr("abs:xmlUrl");
 
-            if (title.isEmpty()) {
-                continue;
-            }
-
-            if (xmlUrl.isEmpty()) {
-                throw new InvalidSourceException("document has invalid entries");
-            }
-
             try {
                 String id = Parser.matchGroup1(ID_PATTERN, xmlUrl);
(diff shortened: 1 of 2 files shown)


mauriciocolli:
Thanks for the pull request!

I think we should remove the check for the `title`, and leave only the `url` one (the essential part).

Like I commented in the issue, I think YouTube does this for channels deleted OR suspended. The channels that are suspended can come back (rarely), so what if the user export _before_ it comes back and tries to import _after_ it comes back? It just won't show up.

I say we should let this entry go and let the importing service handle it (which is simply ignoring the failing entry at this moment), what do you think?

TobiGr:
>After thinking a little more about it, should we fail immediately after we don't find a invalid url? Or be more lenient about it (just ignore the entry)?

Yes, we should ignore and not fail to import. Maybe return the number of imports/urls which caused a problem and then display a toast / dialog which or how many subscriptions could not be imported.

Stypox:
I don't think it would be much useful for the user to know how many subscriptions couldn't be imported. And that number will be 0 most of the times (if not always: a subscription with an empty title still has a valid url). I would just do what mauriciocolli suggested: ignoring the invalid entries.

Issue #155, opened 2019-04-13
Can't play/download certain YouTube videos


Phoenix616:
Trying to play or download certain videos seems to be failing. They player just loads for a long time and the downloader fails with a 404 error.

The only video with this issue that I've encountered so far is this one: https://www.youtube.com/watch?v=jdELRPEbX4A

It seems to play fine on the website or official app but other third party players (noticed it in Kodi first) seem to fail too.

theScrabi:
Noticed invideo.us has similar issues. Could you test if you can download the video using youtube-dl? Maybe they have a solution.

omarroth:
Invidious will sometimes have issues with playback since users download the video from a different IP from where the video is requested. See omarroth/invidious#92, you should be able to fix this with `&local=true`.

I'm able to play/download the video linked in Invidious, youtube-dl, and NewPipe. I expect though the issue in this case is similar to fent/node-ytdl-core#417. To paraphrase: the video signature sometimes needs to be added as a different param name than `&signature=`.  See also ytdl-org/youtube-dl#18927.

[This line](https://github.com/TeamNewPipe/NewPipeExtractor/blob/560c648e92476cc220a036f1668c4335b98ceece/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java#L891) likely needs to be changed to something like:
```java
streamUrl = streamUrl + tags.get("sp") + decryptSignature(tags.get("s"), decryptionCode);
```

Similar to [this line](https://github.com/omarroth/invidious/blob/0.16.0/src/invidious/signatures.cr#L62) in Invidious.

hp8wvvvgnj6asjm7:
Same here since two weeks.. please when do you release an update?

TiA4f8R:
This issue should have been fixed (non-progressive/unavailable streams have been hidden since 0.20.0 and extraction of streams with `signatureCipher`s is handled properly), that's why I am closing this issue.

If that's not the case, please let us know.

Thank you!

PR #163, opened 2019-05-13
Fix invalid yt url: signature tag name is not always "signature"

### LINKED ISSUE #162: cannot decrypt some video
hello,
some video such as https://www.youtube.com/watch?v=DTjdcgSXMwQ
when I repeat call, sometime cannot decrypt signature( use latest code), 
**** decryption code****
`var uv={lL:function(a){a.reverse()},NB:function(a,b){var c=a[0];a[0]=a[b%a.length];a[b%a.length]=c},nu:function(a,b){a.splice(0,b)}};var vv=function(a){a=a.split("");uv.NB(a,70);uv.NB(a,67);uv.NB(a,2);uv.NB(a,19);uv.lL(a,6);uv.NB(a,45);uv.NB(a,56);uv.nu(a,3);uv.lL(a,49);return a.join("")};function decrypt(a){return vv(a);}`

this info, can decrypt
`{itag=18, s=3B603054D4A80E0356E04F7D4E5EF5C396A271FE.6C50B059F5A51CCEA1A168CB473287E54BC1C82E821, quality=medium, type=video/mp4; codecs="avc1.42001E, mp4a.40.2", sp=signature, url=XXXX`

this info, cannot decrypt
`{itag=18, s=TL5xI2wwRQIhANJE_8nAuKPeSKt26hPGlE5TFmBCId8-7dU4hO=o8gkDAiBQXHjXWMXgELnNT17VH_fCgKF3pYpuPIbpU10SBUybvQ==Q=3, quality=medium, type=video/mp4; codecs="avc1.42001E, mp4a.40.2", sp=sig, url=XXXX`

****now i fixed it with
https://github.com/TeamNewPipe/NewPipeExtractor/issues/155#issuecomment-483033261

Stypox:
- [no breaking change] I agree to ASAP create a PULL request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) for making in compatible when I changed the api.

Thanks to @omarroth for the suggestion. For more info see #155 #162 TeamNewPipe/NewPipe#2337

This should fix many problems users keep having in the last few weeks: some videos in NewPipe don't load, at random. Invalid urls of those videos are then cached, and the app won't load them anymore until the cache is wiped. So even after this pr is merged, users would still have to manually wipe metadata from the app.

I tested the changes in the NewPipe app and it worked normally with all the videos I tried. I also added a `Log` instruction to have the url printed in the android logcat, and I saw that both types* of YouTube links work (one of them wouldn't have worked previously).
###### * One type (the one that has always worked) has "id" as the first url tag and signature tag "signature", the other (that created problems with the signature) has "expire" as the first url tag and signature tag "sig". Or at least it seems to be like this, to me.

Please merge this as soon as possible, since this issue is affecting many people and is very annoying :-) .

Fixes #162
Fixes (partially) #155 
Fixes TeamNewPipe/NewPipe#2336
Fixes TeamNewPipe/NewPipe#2337
Fixes TeamNewPipe/NewPipe#2324
Fixes TeamNewPipe/NewPipe#2304
Fixes TeamNewPipe/NewPipe#2275
Probably also fixes TeamNewPipe/NewPipe#867
Probably also fixes TeamNewPipe/NewPipe#2297

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
@@ -888,7 +888,7 @@ public class YoutubeStreamExtractor extends StreamExtractor {
                         String streamUrl = tags.get("url");
                         // if video has a signature: decrypt it and add it to the url
                         if (tags.get("s") != null) {
-                            streamUrl = streamUrl + "&signature=" + decryptSignature(tags.get("s"), decryptionCode);
+                            streamUrl = streamUrl + "&" + tags.get("sp") + "=" + decryptSignature(tags.get("s"), decryptionCode);
                         }
                         urlAndItags.put(streamUrl, itagItem);
                     }
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
@@ -888,7 +888,13 @@ public class YoutubeStreamExtractor extends StreamExtractor {
                         String streamUrl = tags.get("url");
                         // if video has a signature: decrypt it and add it to the url
                         if (tags.get("s") != null) {
-                            streamUrl = streamUrl + "&" + tags.get("sp") + "=" + decryptSignature(tags.get("s"), decryptionCode);
+                            if (tags.get("sp") == null) {
+                                // fallback for urls not conaining the "sp" tag
+                                streamUrl = streamUrl + "&signature=" + decryptSignature(tags.get("s"), decryptionCode);
+                            }
+                            else {
+                                streamUrl = streamUrl + "&" + tags.get("sp") + "=" + decryptSignature(tags.get("s"), decryptionCode);
+                            }
                         }
                         urlAndItags.put(streamUrl, itagItem);
                     }


Stypox:
The unsuccessful checks are due to a video used to test comments that is not available anymore.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -888,7 +888,7 @@ private static String getVideoInfoUrl(final String id, final String sts) {
                         String streamUrl = tags.get("url");
                         // if video has a signature: decrypt it and add it to the url
                         if (tags.get("s") != null) {
-                            streamUrl = streamUrl + "&signature=" + decryptSignature(tags.get("s"), decryptionCode);
+                            streamUrl = streamUrl + "&" + tags.get("sp") + "=" + decryptSignature(tags.get("s"), decryptionCode);
```
TobiGr: Thanks! I'd like to include this into 0.16.2 so we can get this fix out asap. Can we ensure that `tags.get("sp") != null`? If this is the case, we are good to go. Otherwise, I'd suggest to use `streamUrl = streamUrl + "&signature=" + decryptSignature(tags.get("s"), decryptionCode);` as fallback if `tags.get("sp") == null`.

hp8wvvvgnj6asjm7:
asap pls!

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
Stypox: Ok, I'll implement and test it this evening

Stypox:
Implemented and tested the suggestion from @TobiGr 
Even if we can ensure that `tags.get("sp") != null`, there is no way to guarantee that YouTube will never remove the signature name.

PR #182, opened 2019-08-08
Add Invidious instances

2secslater:
Added all publicly listed Invidious instances from omarroth/invidious wiki page to the link handler factory for YouTube.


TobiGr:
Hm...
I can now open the videos, but I get a warning: 
https://github.com/TeamNewPipe/NewPipeExtractor/blob/8de53111d9ca94abda47b196eaaed6331e793df9/extractor/src/main/java/org/schabi/newpipe/extractor/linkhandler/LinkHandlerFactory.java#L43
Can you please try to track it down?

2secslater:
what URL did you try to open?

TobiGr:
I get the error regardless which instance I use... But I just recognized this happens when using invidio.us, too. We can fix this in a different pull request as it is not caused by your changes.

PR #184, opened 2019-08-12
 Fix inconsistency in youtube channel urls

Stypox:
- [yes] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [yes] I did test the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).
- [no need] I agree to ASAP create a PULL request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) for making in compatible when I changed the api.

That inconsistency created problems in NewPipe's database. It lead to wrong results when comparing two channels, since they could have been the same even with different urls. This pr makes all youtube channel urls provided by the extractor be in the form "https://www.youtube.com/channel/ID".

Fixes TeamNewPipe/NewPipe#2167

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
@@ -47,6 +47,7 @@ import java.util.ArrayList;
 
 @SuppressWarnings("WeakerAccess")
 public class YoutubeChannelExtractor extends ChannelExtractor {
+    /*package-private*/ static final String CHANNEL_URL_BASE = "https://www.youtube.com/channel/";
     private static final String CHANNEL_FEED_BASE = "https://www.youtube.com/feeds/videos.xml?channel_id=";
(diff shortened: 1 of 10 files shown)


TobiGr:
Some tests are failing due to your changes, can you please fix them?

Maybe we need to update all subscription urls in the app database to use the `https://www.youtube.com/channel/ID` format. Or did we force the app to use this form earlier in the subscription and history db?

Stypox:
@TobiGr I fixed the tests and the regex, sorry for that.
Newpipe's database contains both "/user/" and "/channel/", depending on the url provided by the extractor. That's why at the beggining I couldn't reproduce TeamNewPipe/NewPipe#2167
A migration would be a good idea, but it would require network requests to get the channel id for the url.

Stypox:
There are other extractors that need to be changed: this pr is not ready

Stypox:
I have gone through all `getUploaderUrl()` functions and changed them. In `InfoStreamItem`s I couldn't find a way to reliably get a url containin "/channel/...", so I added a comment. This is not good, but does not cause problems in the current newpipe implementation, as `InfoStreamItem`'s `uploaderUrl`s are never queried (there is no "subscribe to the uploader" button in the long press menu).

PR #189, opened 2019-08-20
Retrieve recent videos ISO datetime strings from subscribed channels

timfbfbfbfb:
### Features:
1. Retrieve videos publish dates (in ISO time format) from the Youtube RSS feed XML file

### Remarks:
This PR should be merged together with TeamNewPipe/NewPipe#2547.


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
```diff
@@ -62,6 +64,29 @@ public void onFetchPage(@Nonnull Downloader downloader) throws IOException, Extr
         String channelUrl = super.getUrl() + CHANNEL_URL_PARAMETERS;
         String pageContent = downloader.download(channelUrl);
         doc = Jsoup.parse(pageContent, channelUrl);
+
+        String feedUrl = getFeedUrl();
+        retrieveVideoTimestampFromXml(Jsoup.parse(
+                downloader.download(feedUrl),
+                feedUrl,
+                org.jsoup.parser.Parser.xmlParser()));
```
Stypox: Content should not be extracted in the `onFetchPage` method. This method should only fetch pages, so you should save the result of `Jsoup.parse(downloader.download(feedUrl), feedUrl, org.jsoup.parser.Parser.xmlParser())` in a variable. This way data can be extracted **on demand**, making the extractor faster when not all info is needed. So you should call `retrieveVideoTimestampFromXml` in `getPublishIsoTimeStrLookup` with the saved variable from above.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
```diff
@@ -88,7 +113,7 @@ public String getId() throws ParsingException {
 
             return element.attr("data-channel-external-id");
         } catch (Exception e) {
-            throw new ParsingException("Could not get channel id", e);
+            return super.getId();
```
Stypox: Are you sure this is ok? You should probably check that `super.getId()` (that is equivalent to `linkHandler.getId()`, see [Extractor.java, line 74](https://github.com/TeamNewPipe/NewPipeExtractor/blob/f250db521e490c6b45059463ac26e4137aaccb44/extractor/src/main/java/org/schabi/newpipe/extractor/Extractor.java#L74) and [LinkHandler.java, line 29](https://github.com/TeamNewPipe/NewPipeExtractor/blob/f250db521e490c6b45059463ac26e4137aaccb44/extractor/src/main/java/org/schabi/newpipe/extractor/linkhandler/LinkHandler.java#L29)) really returns something. If it returns nothing (i.e. `null` or an empty string), just throw to the exception (`throw new ParsingException("Could not get channel id", e);`).

wb9688:
Closed in favor of @mauriciocolli's solution.

PR #190, opened 2019-08-26
Fetching frames preview from youtube

nv95:
- [ ] I agree to ASAP create a PULL request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) for making in compatible when I changed the api.

I found the way to get frames of youtube videos (for example, https://i9.ytimg.com/sb/HoK9shIJ2xQ/storyboard3_L0/default.jpg?sqp=ovOX_wMGCNawlNgF&sigh=rs$AOn4CLA-CIXBDpHJ_2hQzANfY6D568Ez4Q) and tried to implement it. I don't understand well architecture of extractor, so, please, help me to integrate this correctly)

If you shall found it useful, I can implement this feature in NewPipe.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
@@ -1035,4 +1035,29 @@ public class YoutubeStreamExtractor extends StreamExtractor {
             }
         };
     }
+
+    @Nullable
+    public StreamFrames getFrames() {
+        try {
+            final String script = doc.select("#player-api").first().siblingElements().select("script").html();
+            int p = script.indexOf("ytplayer.config");
+            if (p == -1) {
+                return null;
+            }
+            p = script.indexOf('{', p);
+            int e = script.indexOf("ytplayer.load", p);
+            if (e == -1) {
+                return null;
+            }
+            JsonObject jo = JsonParser.object().from(script.substring(p, e - 1));
+            final String resp = jo.getObject("args").getString("player_response");
+            jo = JsonParser.object().from(resp);
+            final String[] spec = jo.getObject("storyboards").getObject("playerStoryboardSpecRenderer").getString("spec").split("\\|");
+            final String url = spec[0];
+            final List<String> opts = Arrays.asList(spec).subList(1, spec.length);
+            return new StreamFrames(url, opts);
+        } catch (Exception e) {
+            return null;
+        }
+    }
 }
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamFrames.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamFrames.java
@@ -1,0 +1,113 @@
+package org.schabi.newpipe.extractor.stream;
+
+import javax.annotation.Nullable;
+import java.util.ArrayList;
+import java.util.List;
+
+public class StreamFrames {
+
+	private final List<Frameset> frames;
+
+	public StreamFrames(String baseUrl, List<String> params) {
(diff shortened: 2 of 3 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamFrames.java
```diff
@@ -0,0 +1,113 @@
...
+import javax.annotation.Nullable;
+import java.util.ArrayList;
+import java.util.List;
+
+public class StreamFrames {
+
+	private final List<Frameset> frames;
+
+	public StreamFrames(String baseUrl, List<String> params) {
+		frames = new ArrayList<>(params.size());
```
Stypox: This link generation should be done in the YouTubeStreamExtractor above, since it is a youtube-specific thing. If, for example, PeerTube provides something similar, it wouldn't be possible to use this StreamFrames class.
Services implementations lie in the `extractor/services/` subfolder, and service-specific things should ONLY be there. All of the files not in that subfolder are: utilities common to every service (like `MediaFormat`), interfaces to be implemented by services (like `InfoItemExtractor`), interfaces to be implemented by NewPipe or users (like `Downloader`), classes wrapping around extracted data (like `InfoItem`).

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamFrames.java
```diff
@@ -0,0 +1,113 @@
+package org.schabi.newpipe.extractor.stream;
+
+import javax.annotation.Nullable;
+import java.util.ArrayList;
+import java.util.List;
+
+public class StreamFrames {
```
Stypox: This should probably be just a class wrapping around extracted data (see below).

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -1035,4 +1035,29 @@ public String getThumbnailUrl() throws ParsingException {
             }
         };
     }
+
+    @Nullable
+    public StreamFrames getFrames() {
```
Stypox: Add a method in the base class `StreamExtractor` that returns `null` by default, then add `@Override`.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeStreamExtractorDefaultTest.java
```diff
@@ -231,4 +231,29 @@ public void testGetFullLinksInDescription() throws ParsingException {
...
+        @Test
+        public void testGetFrames() {
+            final StreamFrames frames = extractor.getFrames();
+            assertNotNull(frames);
+            assertNotNull(frames.getDefaultVariant());
+            for (int i=0;i<frames.getVariantsCount();i++) {
+                final StreamFrames.Frameset frameset = frames.getVariant(i);
+                final int pages = frameset.getTotalPages();
+                final String url = pages == 0 ? frameset.getUrl() : frameset.getUrl(pages - 1);
+                assertNotNull(url);
```
Stypox: You should probably put `assertIsSecureUrl` here (from `ExtractorAsserts`: https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/test/java/org/schabi/newpipe/extractor/ExtractorAsserts.java)

hardBSDk:
It would be a HUGE feature if merged.
@nv95 if merged, the frontend would need a PR for rendering the squares?
Sometimes I think that wonders like this need some YouTube API, am I right?

nv95:
I can also implement it in Android app

hardBSDk:
@nv95 Do you think that it will be merged in the front?

TobiGr:
@nv95 Thanks!
@hardBSDk It will.

PR #191, opened 2019-08-27
Fix id extraction of youtube channels that have no "Subscribe" button

Stypox:
- [yes] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [yes] I did test the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).
- [no need] I agree to ASAP create a PULL request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) for making in compatible when I changed the api.

In 0.17.1 searching for "eminemvevo" led to a crash: in #184 I changed the way to get the channel url from channel info items, but I left a fallback to the old inconsistent method. But I forgot to correctly handle exceptions so the fallback was not called in case of failure. ;-) The channel "EminemVEVO" apparently has no ".../channel/..." url in search results.

Some youtube channels (such as EminemVEVO) have no "Subscribe" button, so the id could not be extracted leading to a ParsingException. Now the id is extracted in a different way.

Fixes TeamNewPipe/NewPipe#1477

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
@@ -83,6 +83,11 @@ public class YoutubeChannelExtractor extends ChannelExtractor {
     @Override
     public String getId() throws ParsingException {
         try {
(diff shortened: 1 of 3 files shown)


TobiGr:
I didn't recognize this PR.

PR #196, opened 2019-09-16
Fix YouTube subscriber count

TobiGr:
Modify test to fail on too small subscriber count

Fixes TeamNewPipe/NewPipe#2632

CC @theScrabi @Stypox

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
@@ -28,6 +28,35 @@ public class Utils {
     }
 
     /**
+     * <p>Convert a mixed number word to a long.</p>
+     * <p>Examples:</p>
+     * <ul>
+     *     <li>123 -&gt; 123</li>
+     *     <li>1.23K -&gt; 1230</li>
+     *     <li>1.23M -&gt; 1230000</li>
+     * </ul>
+     * @param numberWord string to be converted to a long
+     * @return a long
+     * @throws NumberFormatException
+     * @throws ParsingException
+     */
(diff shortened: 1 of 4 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
```diff
@@ -27,6 +27,35 @@ public static String removeNonDigitCharacters(String toRemove) {
...
+     *     <li>123 -&gt; 123</li>
+     *     <li>1.23K -&gt; 1230</li>
+     *     <li>1.23M -&gt; 1230000</li>
+     * </ul>
+     * @param numberWord string to be converted to a long
+     * @return a long
+     * @throws NumberFormatException
+     * @throws ParsingException
+     */
+    public static long mixedNumberWordToLong(String numberWord) throws NumberFormatException, ParsingException {
```
Stypox: Since this is a public method you could add a test only for this one

TobiGr: Done. just noticed that there are no tests for the util methods.

mauriciocolli:
Just a question: Shouldn't we have a way to indicate when we _don't_ have a precise number available? It's a shame if someone is using this library as their source of information for some project that depends on it (e.g. track subscribers over time).

Keep in mind that this change would just make the wording/results more accurate, the user would still have to live with that.

Maybe including a warning in the documentation is enough?

TobiGr:
> Shouldn't we have a way to indicate when we _don't_ have a precise number available?

Yes, you are right. A warning in the docs is the minimum, maybe printing a warning to the console would be a good compromise, but even a new method (something like `isSubscriberCountAccurate()`) is possible.
@Stypox @theScrabi

PR #209, opened 2019-10-23
[SoundCloud] Fix for client id extraction

### LINKED ISSUE #205: SoundCloud broken
Whenever I open the SoundCloud tab it says no videos and crashes and afterwards It won't load YouTube or MediaCCC

mauriciocolli:
- Added `HEAD` request to the current downloader implementation.
- Try to use the hardcoded client id first, falling back to search in scripts the old way.
- Added a test to notify us when it becomes invalid (based on current knowledge: points towards 3~4 months).

---

Closes #205.


mauriciocolli:
@TobiGr notice that the commits have lost their signature, unfortunate, but no big deal.

I think it's an issue with GitHub, the _author_ should merge it when using the "Rebase and Merge", at least according to some issues that I found about it. Something to keep in mind I guess.

PR #210, opened 2019-10-23
Quick fix for failing test

mauriciocolli:
Tests should be refactored as soon as possible. I mean [look at their state](https://github.com/TeamNewPipe/NewPipeExtractor/blob/d14c45c948d3f0641cd41961dca93c7c860d3018/extractor/src/test/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudPlaylistExtractorTest.java#L276..L277), it's sad, really.

For example, from time to time (read often) we get rate limited and receive a 500 status from SoundCloud, we should implement retry and back off in the downloader used in tests. They are very brittle in the current state, and yes, we should expect them to be because we don't control the source, but we can at least improve the situation.

This is just a temporary quick fix.


Stypox:
I tried fixing tests in #180, but I needed some feedback to continue... can you provide some, please? ;-)

Stypox:
This pr should be closed in favour of #234, where the entire YoutubeSearchCountTest class is removed

PR #215, opened 2019-11-21
Fix search result paging

gzsombor:
Due to the way InfoItemsSearchCollector are re-used, the returned item list just grows, which cause that same videos are returned.

Fixes https://github.com/TeamNewPipe/NewPipe/issues/2580

### DIFF
--- a/extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/search/YoutubeSearchPagingTest.java
+++ b/extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/search/YoutubeSearchPagingTest.java
@@ -24,7 +24,8 @@ public class YoutubeSearchPagingTest {
     private static ListExtractor.InfoItemsPage<InfoItem> page2;
     private static Set<String> urlList1;
     private static Set<String> urlList2;
-    private static int pageSize;
+    private static int page1Size;
+    private static int page2Size;
 
     @BeforeClass
     public static void setUpClass() throws Exception {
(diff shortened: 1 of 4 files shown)


Stypox:
Does this fix TeamNewPipe/NewPipe#2580 ?

gzsombor:
I wasn't aware of this bug is already reported, but yes, I think, this is the same issue. 
The problem is that InfoItemsSearchCollector is reused, and with every new page, new items are added, but the old ones are not removed

Stypox:
As far as I understand TeamNewPipe/NewPipe#2580 says that some videos are duplicated, but not entire portions of them, while this pr fixes repeated patches of videos (which I have never encountered). Have you tested with a seach term that gives duplicate results in the current app, but does not with the fix in this pr?

gzsombor:
This error can be seen easily, I've added a simple unit test, to demonstrate it :
* the first page size increase to 40 from 20, after asking for the second page. 
* the second page contains 40 items
* the second page contains 20 items from the first page

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/search/YoutubeSearchPagingTest.java
```diff
@@ -0,0 +1,78 @@
...
+    @Test
+    public void secondPageUniqueVideos() {
+        assertEquals("distinct videos", 20, urlList2.size());
+    }
+
+    @Test
+    public void noRepeatingVideosInPages() {
+        Set<String> intersection = new HashSet<>(urlList2);
+        intersection.retainAll(urlList1);
+        assertEquals("empty intersection", 0, intersection.size());
```
TobiGr:
```suggestion
        assertEquals("Duplicated video on second search page" + , 0, intersection.size());
```

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/search/YoutubeSearchPagingTest.java
```diff
@@ -0,0 +1,78 @@
...
+    public static void setUpClass() throws Exception {
+        NewPipe.init(DownloaderTestImpl.getInstance());
+
+        YoutubeSearchExtractor extractor = (YoutubeSearchExtractor) YouTube.getSearchExtractor("cirque du soleil",
+                singletonList(YoutubeSearchQueryHandlerFactory.VIDEOS), null);
+
+        extractor.fetchPage();
+        page1 = extractor.getInitialPage();
+        urlList1 = extractUrls(page1.getItems());
+        assertEquals("page with items loaded", 20, page1.getItems().size());
```
TobiGr: Can we always assume, that the item count is 20? I'd suggest to choose something like > 15 to not fail when YouTube changes algos again.

gzsombor: I don't know, if this 20 will be 20 forever, so I could rewrite the tests to check for page sizes are equals, and the unique videos are returned.

gzsombor:
I've implemented a change for the tests

gzsombor:
Actually, before this fix, the paging was so broken, that when the second page was loaded, the first page was increased as well. So even if  the  page1.getItems().size() check went fine in the 'setupClass()', when the firstPageOk() is executed, page1.getItems().size() become more than 15, so that checks had meaning. (You can verify it, that without the fix, the setupClass went fine, however all the test method failed)

TobiGr:
Oh yes, I'll re-add the tests tomorrow. Thanks for the heads up.

PR #217, opened 2019-11-28
Fix SoundCloud client_id extraction and update SoundCloud client_id

TobiGr:
Don't throw exception when hard coded client_id is wrong.
Addresses TeamNewPipe/NewPipe#2823

There is still one problem left: the audioStreamTest fails with a 401 response. Can someone please look into this?

CC @mauriciocolli @Stypox @theScrabi

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
@@ -31,7 +31,7 @@ import static org.schabi.newpipe.extractor.ServiceList.SoundCloud;
 import static org.schabi.newpipe.extractor.utils.Utils.replaceHttpWithHttps;
 
 public class SoundcloudParsingHelper {
-    private static final String HARDCODED_CLIENT_ID = "LHzSAKe8eP9Yy3FgBugfBapRPLncO6Ng"; // Updated on 22/10/19
+    private static final String HARDCODED_CLIENT_ID = "bkcJLoXNaiFlsLaKBQXOxO5FhW0NJVnu"; // Updated on 29/11/19
     private static String clientId;
     
     private SoundcloudParsingHelper() {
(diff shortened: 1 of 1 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
```diff
@@ -73,10 +73,16 @@ public static String clientId() throws ExtractionException, IOException {
...
 
-    static boolean checkIfHardcodedClientIdIsValid(Downloader dl) throws IOException, ReCaptchaException {
+    static boolean checkIfHardcodedClientIdIsValid(Downloader dl) {
         final String apiUrl = "https://api.soundcloud.com/connect?client_id=" + HARDCODED_CLIENT_ID;
-        // Should return 200 to indicate that the client id is valid, a 401 is returned otherwise.
-        return dl.head(apiUrl).getResponseCode() == 200;
+        try {
+            // Should return 200 to indicate that the client id is valid, a 401 is returned otherwise.
+            return dl.head(apiUrl).getResponseCode() == 200;
+        } catch (Exception ignored) {
```
TobiGr: It might be good to throw the ReCaptcha exception. What do you think?

Stypox: If I understand correctly this function only has to check if the current client_id is correct, and if it is not it should return false and let `clientId()` fetch another client_id. So it is ok to just return false

TobiGr:
@Stypox What can we do about the 401 response when accessing a stream? With that PR we are just able to view pages, don't ask me why.

TobiGr:
I am not able to fix the stream issue... That's why I will not prepare a hotfix release and this will be merged to the dev branch

pihug12:
I saw this fail in the Travis CI log:
```
Caused by: java.io.IOException: Server returned HTTP response code: 401 for URL: https://api.soundcloud.com/i1/tracks/276206960/streams?client_id=bkcJLoXNaiFlsLaKBQXOxO5FhW0NJVnu
```

Here are the stream URLs for this track:
* Info: https://api-v2.soundcloud.com/tracks/276206960?client_id=bkcJLoXNaiFlsLaKBQXOxO5FhW0NJVnu
* Stream 1: https://api-v2.soundcloud.com/media/soundcloud:tracks:276206960/6f584a2d-87c2-451b-90df-158efcc42680/stream/hls?client_id=bkcJLoXNaiFlsLaKBQXOxO5FhW0NJVnu
* Stream 2: https://api-v2.soundcloud.com/media/soundcloud:tracks:276206960/6f584a2d-87c2-451b-90df-158efcc42680/stream/progressive?client_id=bkcJLoXNaiFlsLaKBQXOxO5FhW0NJVnu

https://github.com/ytdl-org/youtube-dl/blob/master/youtube_dl/extractor/soundcloud.py helped me a bit. I used `youtube-dl -s -v --print-traffic https://soundcloud.com/liluzivert/do-what-i-want-produced-by-maaly-raw-don-cannon` to see the requests done.

I hope it can help you. :)

TobiGr:
Hey @pihug12,
thanks for your help :) I already figured out, that we can need to use the v2 api and how to get no 401 response, but the valid stream. You can find a first, but messy implementation **[here]( 
https://github.com/TeamNewPipe/NewPipeExtractor/blob/fix/SC_stream/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudStreamExtractor.java#L143)**.
However, the player does not open the stream / m3u file correctly.

PR #230, opened 2019-12-25
ChannelInfo: if getName() fails, try to check if there is a user readable error message

### LINKED ISSUE #228: ParsingException is thrown on deleted or non-existant channels
Try to open [this deleted channel](https://www.youtube.com/channel/UChlFShtShMTo9PUHRkasR6A) in NewPipe: instead of getting a message explaining that the channel was deleted, the app crashes. We should add a method that checks whether the channel exists and is valid or not.

## Exception
* __Request:__ https://www.youtube.com/channel/UChlFShtShMTo9PUHRkasR6A
* __Content Language:__ GB
* __Service:__ YouTube
* __Version:__ 0.17.4
* __OS:__ Linux Android 7.0 - 24

```
org.schabi.newpipe.extractor.exceptions.ParsingException: Could not get channel name
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeChannelExtractor.getName(YoutubeChannelExtractor.java:106)
	at org.schabi.newpipe.extractor.channel.ChannelInfo.getInfo(ChannelInfo.java:62)
	at org.schabi.newpipe.extractor.channel.ChannelInfo.getInfo(ChannelInfo.java:49)
	at org.schabi.newpipe.util.-$$Lambda$ExtractorHelper$u5W7VszTe8AoEexIsFM9huQfbkM.call(lambda)
	at java.util.concurrent.FutureTask.run(FutureTask.java:237)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:272)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
	at java.lang.Thread.run(Thread.java:776)
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String org.jsoup.nodes.Node.attr(java.lang.String)' on a null object reference
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeChannelExtractor.getName(YoutubeChannelExtractor.java:104)
	... 28 more

```
</p></details>
<hr>

Redirion:

This fixes #228 

But in NewPipe we should also add a handling that will show the proper Youtube Error Message. At least it won't crash anymore if getName fails.

P.S.: don't judge me for committing on Christmas please. I have slow internet and thought I could do something useful while waiting.


extractor/src/main/java/org/schabi/newpipe/extractor/channel/ChannelInfo.java
```diff
@@ -57,9 +57,31 @@ public static ChannelInfo getInfo(StreamingService service, String url) throws I
...
+        try {
+            info = new ChannelInfo(extractor.getServiceId(),
+                    extractor.getLinkHandler(),
+                    extractor.getName());
+        } catch (ParsingException e) {
+            info = new ChannelInfo(extractor.getServiceId(),
+                    extractor.getLinkHandler(),
+                    "");
+            Throwable nameFailure;
+            if (extractor instanceof YoutubeChannelExtractor) {
```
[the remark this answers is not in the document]
Redirion: the abstract class "Info" could have the new method "public abstract String obtainErrorMessage". It wouldn't do anything yet for other extractors but the YoutubeExtractor, but this would be an approach that wouldn't require casting to the YoutubeChannelExtractor then.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
```diff
@@ -162,6 +164,15 @@ public String getDescription() throws ParsingException {
         }
     }
 
+    public String getYoutubeError() throws ParsingException {
```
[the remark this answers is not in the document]
Redirion: it did initially, but I changed my mind before committing it. I have now added Javadoc and removed the throws

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
```diff
@@ -162,6 +164,23 @@ public String getDescription() throws ParsingException {
...
 
+    /**
+     * Youtube might show a red warning text for example
+     * "This account has been terminated due to multiple or severe violations of
+     * YouTube's policy prohibiting hate speech."
+     * In that case parsing of the channel informations will fail. This error text should
+     * be displayed to the user instead of a cryptic ParsingException.
+     * @return the text as String or null if no warning text could be found
+     */
+    public String getYoutubeError() {
```
Stypox: Add @Nullable

extractor/src/main/java/org/schabi/newpipe/extractor/channel/ChannelInfo.java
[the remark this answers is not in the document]
Stypox: Yeah, you could add `obtainErrorMessage` to the base `Info` class and make it return `null` by default. When extractors override it, they can return a `String` to indicate there was an error or return null otherwise (as you did with youtube).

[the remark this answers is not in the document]
Redirion: sorry, Info is not the correct inheritance hierarchy. I have made a new commit where I have added the method obtainErrorMessage in the abstract class ChannelExtractor, returning null by default, as you have suggested @Stypox

extractor/src/main/java/org/schabi/newpipe/extractor/channel/ChannelInfo.java
```diff
@@ -57,9 +56,27 @@ public static ChannelInfo getInfo(StreamingService service, String url) throws I
...
+            String errorMessage = extractor.obtainErrorMessage();
+            if (errorMessage != null) {
+                nameFailure = new ParsingException(errorMessage);
+            } else {
+                nameFailure = e;
+            }
+            info.addError(nameFailure);
+            // this is so severe, that other extractions most likely fail anyway,
+            // so skip them
+            return info;
```
TobiGr: That's not good and will fail in a `NullPointerExcepetion`. The frontend assumes, that when a `ChannelInfo` is returned there is information which can be processed. We need to fail here because we cannot get any information. 
Just throw the `ParsingException` when `obtainErrorMessage()` brings no result. When `obtainErrorMessage()` returned an error message, I suggest to throw a `ContentNotAvailableException` with the message and then adjust the frontend handlers to display it. Can you check how long the other messages are, which are passed to other `ContentNotAvailableException`'s? If they are too long or not helpful, we might need to add a new `displayMessage` attribute.

Redirion: > That's not good and will fail in a `NullPointerException`. The frontend assumes, that when a `ChannelInfo` is returned there is information which can be processed.

but this is inline with the other cases. For instance
 ```
try {
            info.setOriginalUrl(extractor.getOriginalUrl());
        } catch (Exception e) {
            info.addError(e);
        }
```
would also render a NullPointerException if the originalUrl would be obtained without checking in frontend as there is no setting of an empty url in error case.

Comment (Redirion) [reply extractor/src/main/java/org/schabi/newpipe/extractor/channel/ChannelInfo.java]: I am now throwing the ContentNotAvailableException as requested, which is basically a crash. So I have renamed the PR as it now basically just gets a user readable error message.

File: extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java [hunk 5]
```diff
@@ -162,6 +164,24 @@ public String getDescription() throws ParsingException {
...
+     * "This account has been terminated due to multiple or severe violations of
+     * YouTube's policy prohibiting hate speech."
+     * In that case parsing of the channel information will fail. This error text should
+     * be displayed to the user instead of a cryptic ParsingException.
+     * @return the text as String or null if no warning text could be found
+     */
+    @Override
+    public String obtainErrorMessage() {
+        try {
+            return doc.select("div.yt-alert-message").first().text();
```
TobiGr: Please exclude the old browser message. `:not(#old-browser-alert)`

<details><summary>HTML</summary>

``` HTML
<div class="alerts-wrapper">
  <div id="alerts" class="content-alignment">
   <div class="yt-alert yt-alert-default yt-alert-warn  " id="old-browser-alert">
    <div class="yt-alert-icon">
     <span class="icon master-sprite yt-sprite"></span>
    </div>
    <div class="yt-alert-content" role="alert">
     <div class="yt-alert-message" tabindex="0">
       We'll stop supporting this browser soon. For the best experience, please
      <a href="/supported_browsers">update your browser</a>.
     </div>
    </div>
    <div class="yt-alert-buttons"></div>
   </div>
   <div class="yt-alert yt-alert-default yt-alert-error  ">
    <div class="yt-alert-icon">
     <span class="icon master-sprite yt-sprite"></span>
    </div>
    <div class="yt-alert-content" role="alert">
     <div class="yt-alert-message" tabindex="0">
       This account has been terminated due to multiple or severe violations of YouTube's policy prohibiting hate speech.
     </div>
    </div>
    <div class="yt-alert-buttons">
     <button class="yt-uix-button yt-uix-button-size-default yt-uix-button-close close yt-uix-close" type="button" onclick=";return false;" aria-label="Close" data-close-parent-class="yt-alert"><span class="yt-uix-button-content">Close</span></button>
    </div>
   </div>
  </div>
 </div>
```
</details>

extractor/src/main/java/org/schabi/newpipe/extractor/channel/ChannelInfo.java
TobiGr: I thought about extending the `ContentNotAvailableException` like this:
``` Java
public class ContentNotAvailableException extends ParsingException {
    /**
     * Reason why content is not available.
     * Can be displayed to the user
     */
    private String reason;

   ....

    public ContentNotAvailableException(String message, String reason) {
        super(message);
        this.reason = reason;
    }

   public ContentNotAvailableException(String message, String reason, Throwable cause) {
        super(message, cause);
        this.reason = reason;
    }

    ....

    /**
     * {@link ContentNotAvailableException#reason}
     * @return null when no reason is available
     */
    public String getReason() {
        return reason;
    }
```
This allows us to display the reason in the app by checking if the reason is not null, otherwise display the generic "content unavailable" message.
We could also create an `ContentNotAvailableErrorInfo` Class which holds this information. @yausername @Stypox @Redirion What do you think?

Redirion: maybe instead of adding a new method (getReason) we could use an Java API method that should be fitting:
getLocalizedMessage ?

Stypox: I think @Redirion idea is good, since it follows the answer provided here: https://stackoverflow.com/questions/24988491/difference-between-e-getmessage-and-e-getlocalizedmessage
The issue is: what would getMessage() return? In theory it should return a non-localized reason, but obviously it can't, so is it OK for it to also return a localized message?

Redirion: If I understood @TobiGr correctly, he wants to have getMessage() return the none localized text "Could not get channel name" when getName() fails. Which was the state prior to this PR. With this PR getMessage() would return the localized error message from youtube that is now extracted with "obtainErrorMessage".

getLocalizedMessage just returns what getName returns if not overridden. So if we want to have both we could use getLocalizedMessage() (as an alternative to adding a new method like Tobi suggested)

Stypox: Oh that makes sense, then I agree with you, getLocalizedMessage is a good idea

gillhash:
@TobiGr @Stypox this also needs to be implemented. Just a reminder.

TobiGr:
@gillhash Thanks.
@Redirion We might need to update this after we solved #258

Redirion:
https://github.com/TeamNewPipe/NewPipeExtractor/issues/228#issuecomment-613410434

PR #231, opened 2019-12-31
[Soundcloud] Update client_id

pihug12:
Blind attempt to fix the current Soundcloud issue (https://github.com/TeamNewPipe/NewPipe/issues/2915).

The previous client_id seems now invalid.

**Old:** https://api-v2.soundcloud.com/tracks/276206960?client_id=bkcJLoXNaiFlsLaKBQXOxO5FhW0NJVnu
**New:** https://api-v2.soundcloud.com/tracks/276206960?client_id=r5ELVSy3RkcjX7ilaL7n2v1Z8irA9SL8

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
@@ -31,7 +31,7 @@ import static org.schabi.newpipe.extractor.ServiceList.SoundCloud;
 import static org.schabi.newpipe.extractor.utils.Utils.replaceHttpWithHttps;
 
 public class SoundcloudParsingHelper {
-    private static final String HARDCODED_CLIENT_ID = "bkcJLoXNaiFlsLaKBQXOxO5FhW0NJVnu"; // Updated on 29/11/19
+    private static final String HARDCODED_CLIENT_ID = "r5ELVSy3RkcjX7ilaL7n2v1Z8irA9SL8"; // Updated on 31/12/19
     private static String clientId;
     
     private SoundcloudParsingHelper() {


ADepic:
Can you compile an apk to see if this works?

theScrabi:
Are we officially using the api? If not can we make it official, and why are thrse ids oftenly not working all of a sudden?

placaze:
I could be wrong but I don't think so. It's kinda of a hack to use the ones used by the website (that change everytime they update their website).

I think we have to deal with authentification to officially use their API (https://developers.soundcloud.com/docs/api/guide)?

In the [2020.01.01 youtube-dl](https://github.com/ytdl-org/youtube-dl/releases/tag/2020.01.01) release, they added the ability to automatically search for the client_id when the requests fail:
* https://github.com/ytdl-org/youtube-dl/commit/3bed621750b7fe25afc04a0131664bbbc610c563
* https://github.com/ytdl-org/youtube-dl/commit/de7aade2f872d6de2dbd0d82624e51c24968e057

pihug12:
I could be wrong but I don't think so. It's kinda of a hack to use the ones used by the website (that change everytime they update their website).

I think we have to deal with authentification to officially use their API (https://developers.soundcloud.com/docs/api/guide)?

In the [2020.01.01 youtube-dl](https://github.com/ytdl-org/youtube-dl/releases/tag/2020.01.01) release, they added the ability to automatically search for the client_id when the requests fail:
* https://github.com/ytdl-org/youtube-dl/commit/3bed621750b7fe25afc04a0131664bbbc610c563
* https://github.com/ytdl-org/youtube-dl/commit/de7aade2f872d6de2dbd0d82624e51c24968e057

TobiGr:
> Are we officially using the api? If not can we make it official, and why are thrse ids oftenly not working all of a sudden?

@theScrabi The website uses the Api. SoundCloud changes the `client_id` from time to time. We both have a hardcoded client_id and a method which fetches the cliend_id  which is currently used on the website. We test the hardcoded id if it is still valid before fetching the one from the website. It seems like we need to update the [method which detects whether a client_id is valid](https://github.com/TeamNewPipe/NewPipeExtractor/blob/d83787a5ca308c4ca4e86e63a8b63c5e7c4708d6/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java#L76) to check whether we can access a stream and not some unimportant data.

pihug12:
I'm currenty at work and cannot confirm this with the company proxy, but it looks like there is a HTTP 302 (redirect) first in the method you mentionned (api.soundcloud.com --> secure.soundcloud.com). So it's not HTTP 200.

Maybe use `api-v2` (`final String apiUrl = "https://api-v2.soundcloud.com/connect?client_id=" + HARDCODED_CLIENT_ID;`) instead?

TobiGr:
> Maybe use `api-v2` (`final String apiUrl = "https://api-v2.soundcloud.com/connect?client_id=" + HARDCODED_CLIENT_ID;`) instead?

This v2 url does not work. It seems like `"https://api.soundcloud.com/connect?client_id=" + HARDCODED_CLIENT_ID` is accepted for some more time although there is a new client_id. Otherwise I don't have an explanation why you got no result in https://github.com/TeamNewPipe/NewPipe/issues/2915, but now the request is not answered with a `200` status.

pihug12:
I'm back home where Soundcloud isn't blocked. :stuck_out_tongue: 

`/connect` doesn't seem the best check to me. It seems better with `/tracks`.

https://api.soundcloud.com/connect?client_id=bkcJLoXNaiFlsLaKBQXOxO5FhW0NJVnu --> 401
https://api.soundcloud.com/connect?client_id=r5ELVSy3RkcjX7ilaL7n2v1Z8irA9SL8 --> 302

image

https://api.soundcloud.com/tracks?client_id=bkcJLoXNaiFlsLaKBQXOxO5FhW0NJVnu --> 401
https://api.soundcloud.com/tracks?client_id=r5ELVSy3RkcjX7ilaL7n2v1Z8irA9SL8 --> **200**

> It seems like "https://api.soundcloud.com/connect?client_id=" + HARDCODED_CLIENT_ID is accepted for some more time although there is a new client_id. Otherwise I don't have an explanation why you got no result in TeamNewPipe/NewPipe#2915, but now the request is not answered with a 200 status.

I'm no good to read Java. I don't know what is expected with:
https://github.com/TeamNewPipe/NewPipeExtractor/blob/d83787a5ca308c4ca4e86e63a8b63c5e7c4708d6/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java#L44-L47
https://github.com/TeamNewPipe/NewPipeExtractor/blob/d83787a5ca308c4ca4e86e63a8b63c5e7c4708d6/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java#L76-L85

pihug12:
Unbelievable. I reinstall the app and the problem is now gone. -_-

The code of the extractor is fine. I debugged the method with IntelliJ (when the *client_id* is OK and when it is not).

Could it be a cache issue? @ADepic Could you try after clearing the metadata cache in the 4th menu of the settings?

ADepic:
Clearing cache seemed to make SoundCloud work.

HOWEVER, I am using !2309 as my default version of newpipe, and that isn't fixed by clearing cache. But clearing user data seems to make that work...

ADepic:
EDIT2: Importing content databse and PRESSING CANCEL on "import settings" allows SoundCloud to work.

This time I had to clear user data 2-3 times just for the bug to go...

pihug12:
I think you have the bug you mentionned with the "unofficial" https://github.com/TeamNewPipe/NewPipe/pull/2309 build because https://github.com/TeamNewPipe/NewPipeExtractor/commit/8e53fdad16c3e687f6f88ea1328a788e62ea13a5 is missing the *mauriciocolli* build.

____

**Official v0.18.0 build:**

https://github.com/TeamNewPipe/NewPipe/blob/v0.18.0/app/build.gradle#L65:
```
implementation 'com.github.TeamNewPipe:NewPipeExtractor:8e53fda'
```
Extractor commits list: https://github.com/TeamNewPipe/NewPipeExtractor/commits/8e53fda

____

***mauriciocolli* build:**
https://github.com/mauriciocolli/NewPipe/blob/feed/app/build.gradle#L75:
```
implementation 'com.github.mauriciocolli:NewPipeExtractor:bf7e4b82'
```
Extractor commits list: https://github.com/mauriciocolli/NewPipeExtractor/commits/bf7e4b82 (the last commit of the official build is missing and it's a Soundcloud fix)

ADepic:
Yh, okay fine. Then clearing data does seem to work of the official newpipe, HOWEVER importing database and saying YES to the import settings brings back the bug. Saying NO to import settings removes the bug

ADepic:
So my assumption is that some settings from an older version of newpipe carried over to 0.18.0 and caused this bug

PR #233, opened 2020-01-06
Youtube stream extractor improvements by using playerResponse json

Stypox:

## Summary
Some time ago me and @theScrabi agreed that it would be better to extract data using the playerResponse json data, since it is faster to query, less error-prone and less subject to changes.
So with this PR the extractor will use playerResponse whenever possible as the first method of obtaining data. Fallback methods based on HTML querying have been kept. In case of error the thrown exception contains the data about the playerResponse failure, that should be clearer than a NPE caused by not-found HTML tags.
Made error messages more consistant and fixed some related copy-paste mistakes.
Added tests for video ratings.

## Function-specific changes
- `getName`, `getTextualUploadDate`, `getViewCount`, `getUploaderName`: add playerResponse method as the primary one;
- `getThumbnailUrl`: add playerResponse method as the primary one; remove `playerArgs` method that would never work;
- `getDescription`: the HTML-based approach is the only one that provides an HTML-formatted description so it has to be kept as the primary method; add playerResponse method as fallback that provides raw text;
- `getLength`: first try to get the length from playerResponse["videoDetails"]
- `getLikeCount`, `getDislikeCount`: assert that ratings are disabled when the button is missing
- `getUploaderUrl`: add playerResponse method as the primary one; the returned channel url could be different from the one returned before with music channels, but I don't think this is an issue since YouTube picks one at random, too.
- `getUploaderAvatarUrl`: assure `@Nonnull` is respected;
- `getStreamType`: first check inside playerResponse, only if it failed use playerArgs

## ParsingException vs ExtractionException
I have a question: what is the difference between `ParsingException` and `ExtractionException`? I see that they are used interchangeably in all services, though `ParsingException`  is used more often.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
@@ -296,7 +296,7 @@ public class YoutubeStreamExtractor extends StreamExtractor {
                         .getArray("formats")
                         .getObject(0)
                         .getString("approxDurationMs");
-                return Math.round(Long.parseLong(durationMs)/1000.0f);
+                return Math.round(Long.parseLong(durationMs) / 1000f);
             } catch (Exception ignored) {
                 throw new ParsingException("Could not get duration", e);
             }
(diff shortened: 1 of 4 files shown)


TobiGr:
I took a first look at that looks good :+1: 

The `ExtractionException` tells us that something went wrong while extracting information and that this info could not be fetched.
There are two reason for this:
- we could not get the document containing the info => `IOException` or `ReCaptchaException`
- there was an error while parsing the document to retrieve the requested info -> `ParsingException`; reasons for not being able to parse:
  - we got something to parse, but cannot retrieve the info requested -> raw `ParsingException`
  - the content has been removed or blocked: `ContentNotAvailableException`
  - we got an advertisement, but not the info -> `FoundAdException`

Exceptions

Stypox:
Ok, thanks for the explanation 😊

Stypox:
@mauriciocolli when you have some time could you take a look at this, too? In TeamNewPipe/NewPipe#2309 you said "[Maybe] Use the new structure YouTube, it uses JSON which is easier to parse and less error prone", so I'd like to hear you opinion :-)

Stypox:
Done ;-)

PR #253, opened 2020-02-02
Extractor for youtube mix (auto-generated playlist)

XiangRongLin:

closes  TeamNewPipe/NewPipe#2895

## Problem
As describes in the issue in the NewPipe repo, when sharing a youtube mix to newpipe, it crashes. Reason for that is that the linkhandler receives (1)"https://www.youtube.com/watch?v=QXCE3Betyug&list=RDQXCE3Betyug" and converts it to (2)"https://www.youtube.com/playlist?list=RDQXCE3Betyug", which loads a basicly empty page.

## Changes
- The LinkHandler doesn't convert mix urls.
- New PlaylistExtractor to extract playlist information from the url (1).
- The YoutubeService uses the new extractor if it is a mix.

## Testing
I tested the changes on my android 10 and 7.1 devices and a 5.0 emulator, by:
- sharing a link from my browser to NewPipe
- saving the playlist and opening it from the bookmarks fragment
- playling the playlist (didn't play on 5.0 emulator, but i can't play ANY video here)

This doesn't require changes in the app, besides updating the extractor dependency.

[app-debug.zip](https://github.com/TeamNewPipe/NewPipeExtractor/files/4153412/app-debug.zip)

## What's missing
- Mixes are generated based on the user. Meaning if I open the site once while logged in and once while in incognito mode, it gives me different playlists.
- Mixes on youtube are endless but i couldn't find a link for the continuation
- A lot of other fields are not present in the html, so i just return empty strings or 0

## Notes
- In the mix url (1) above you can see the playlist id is just the video id with "RD" at the start.
All mix urls that i found so far are like that, where the playlist id is "RD" + video id of the first video. I assume RD stands for radio (Playlist ids start with "PL" instead)

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
@@ -18,6 +18,10 @@ import org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeParsingH
 import org.schabi.newpipe.extractor.stream.StreamInfoItem;
 import org.schabi.newpipe.extractor.stream.StreamInfoItemsCollector;
 
+/**
+ * A YoutubePlaylistExtractor for a mix (auto-generated playlist).
+ * It handles urls in the format of "youtube.com/watch?v=videoId&list=playlistId"
+ */
 public class YoutubeMixPlaylistExtractor extends PlaylistExtractor {
 
   private Document doc;
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
@@ -1,12 +1,8 @@
 package org.schabi.newpipe.extractor.services.youtube.extractors;
 
-import com.grack.nanojson.JsonObject;
-import com.grack.nanojson.JsonParser;
-import com.grack.nanojson.JsonParserException;
 import java.io.IOException;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import org.jsoup.Jsoup;
 import org.jsoup.nodes.Document;
 import org.jsoup.nodes.Element;
 import org.schabi.newpipe.extractor.StreamingService;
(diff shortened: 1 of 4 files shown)


B0pol:
Instead of ``Auto-generated (no uploader found)``, you should put ``YouTube``
YouTube_mix (look at top left)

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,205 @@
...
+  @Override
+  public String getUploaderUrl() {
+    //Youtube mix are auto-generated
+    return "";
+  }
+
+  @Override
+  public String getUploaderName() {
+    //Youtube mix are auto-generated
+    return "";
```
B0pol: return YouTube instead of nothing.
See this screenshot, on YouTube website (look at top left):
YouTube_mix

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  @Override
+  public String getUploaderAvatarUrl() {
```
B0pol: As they are generated by YouTube, shouldn't you return YouTube logo (or censored version)?

XiangRongLin: I would probably need to hardcode the url, since i can't find one in the crawled html.
Another way would be to save it as a resource in the app and load from there, but that will probably cause license issues.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
XiangRongLin: I don't have that YouTube string marked in you image when i open the site.

But that could also be changed in the app, since there was already a PR to deal with playlists without uploader. Just the String resource would need to be changed.
TeamNewPipe/NewPipe#2724

XiangRongLin:
With #258 coming soon, I will probably need to adjust my code to the new user agent. Before that I would like to know if the general approach is ok.

Stypox:
@XiangRongLin the general approach seems good to me. But ask @wb9688 to be sure ;-)

XiangRongLin:
An idea on how to make it endless would be requesting a new mix from YouTube with the last video as the starting point.

I'll check out on the weekend, if it works and whether there are too many duplicates.

PR #256, opened 2020-02-12
Support for non-api mediaccc links, invidious shortened links, refactor mediaccc tests 

B0pol:

See TeamNewPipe/NewPipe#3090

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCStreamExtractor.java
@@ -170,15 +170,13 @@ public class MediaCCCStreamExtractor extends StreamExtractor {
         return null;
     }
 
-    @Nonnull
     @Override
     public List<SubtitlesStream> getSubtitlesDefault() throws IOException, ExtractionException {
         return null;
     }
 
-    @Nonnull
     @Override
-    public List<SubtitlesStream> getSubtitles(MediaFormat format) throws IOException, ExtractionException {
+    public List<SubtitlesStream> getSubtitles(final MediaFormat format) throws IOException, ExtractionException {
         return null;
     }
 
@@ -212,7 +210,6 @@ public class MediaCCCStreamExtractor extends StreamExtractor {
         } catch (JsonParserException jpe) {
             throw new ExtractionException("Could not parse json returned by url: " + getLinkHandler().getUrl(), jpe);
         }
-
     }
 
     @Nonnull
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCStreamExtractor.java
@@ -16,6 +16,7 @@ import org.schabi.newpipe.extractor.stream.*;
 import javax.annotation.Nonnull;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 
@@ -104,13 +105,13 @@ public class MediaCCCStreamExtractor extends StreamExtractor {
     @Nonnull
     @Override
     public String getDashMpdUrl() throws ParsingException {
-        return null;
+        return "";
     }
 
     @Nonnull
     @Override
     public String getHlsUrl() throws ParsingException {
-        return null;
+        return "";
     }
 
     @Override
@@ -170,14 +171,16 @@ public class MediaCCCStreamExtractor extends StreamExtractor {
         return null;
     }
 
+    @Nonnull
     @Override
     public List<SubtitlesStream> getSubtitlesDefault() throws IOException, ExtractionException {
-        return null;
+        return Collections.emptyList();
     }
 
+    @Nonnull
     @Override
     public List<SubtitlesStream> getSubtitles(final MediaFormat format) throws IOException, ExtractionException {
-        return null;
+        return Collections.emptyList();
     }
 
     @Override
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/linkHandler/MediaCCCConferenceLinkHandlerFactory.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/linkHandler/MediaCCCConferenceLinkHandlerFactory.java
@@ -1,13 +1,9 @@
 package org.schabi.newpipe.extractor.services.media_ccc.linkHandler;
 
-import org.schabi.newpipe.extractor.exceptions.FoundAdException;
 import org.schabi.newpipe.extractor.exceptions.ParsingException;
 import org.schabi.newpipe.extractor.linkhandler.ListLinkHandlerFactory;
 import org.schabi.newpipe.extractor.utils.Parser;
-import org.schabi.newpipe.extractor.utils.Utils;
 
-import java.net.MalformedURLException;
-import java.net.URL;
 import java.util.List;
 
 public class MediaCCCConferenceLinkHandlerFactory extends ListLinkHandlerFactory {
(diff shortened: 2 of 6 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/linkHandler/MediaCCCConferenceLinkHandlerFactory.java
```diff
@@ -14,19 +18,43 @@ public String getUrl(String id, List<String> contentFilter, String sortFilter) t
...
+            throw new IllegalArgumentException("The given URL is not valid");
+        }
+
+        String path = url.getPath();
+        // remove leading "/" of URL-path if URL-path is given
+        if (!path.isEmpty()) {
+            path = path.substring(1);
+        }
+
+        if (path.contains("b/")) {
```
TobiGr: It looks like you want to check if the path starts with `b/`. if this is the case use
```suggestion
        if (path.startsWith("b/")) {
```
for better performance. In which case can an url path start with `b/`? Do we need to check the host here or do we have only urls like `/b/myConferenceId`?

extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/linkHandler/MediaCCCStreamLinkHandlerFactory.java
```diff
@@ -1,16 +1,39 @@
...
+            throw new IllegalArgumentException("The given URL is not valid");
+        }
+
+        String path = url.getPath();
+        // remove leading "/" of URL-path if URL-path is given
+        if (!path.isEmpty()) {
+            path = path.substring(1);
         }
+
+        if (path.contains("v/")) {
```
TobiGr:
```suggestion
        if (path.startsWith("v/")) {
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/linkHandler/MediaCCCConferenceLinkHandlerFactory.java
wb9688: Instead of this, just do `urlString.startsWith("https://media.ccc.de/b/")` (and preferably change the name of the `urlString` variable back to just `url`)

extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/linkHandler/MediaCCCConferenceLinkHandlerFactory.java
```diff
@@ -14,19 +18,43 @@ public String getUrl(String id, List<String> contentFilter, String sortFilter) t
...
     }
 
     @Override
     public boolean onAcceptUrl(String url) throws ParsingException {
-        return url.startsWith("https://api.media.ccc.de/public/conferences/")
-                || url.startsWith("https://media.ccc.de/c/");
+        try {
+            getId(url);
+            return true;
+        } catch (FoundAdException fe) {
```
wb9688: Why are you catching that FoundAdException? `getId` doesn't even throw that.

extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/linkHandler/MediaCCCConferenceLinkHandlerFactory.java
B0pol: Yes it's conferences @TobiGr 
I changed back to how it was and added ``|| url.startsWith("https://media.ccc.de/b/")`` in the if condition

extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/linkHandler/MediaCCCConferenceLinkHandlerFactory.java
B0pol: my bad. because I used the same piece of code of youtubestreamlinkhandlerfactory and yeah, it's thrown on ad. I removed it.

extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCStreamExtractor.java
```diff
@@ -170,15 +170,13 @@ public String getHlsUrl() throws ParsingException {
         return null;
     }
 
-    @Nonnull
     @Override
     public List<SubtitlesStream> getSubtitlesDefault() throws IOException, ExtractionException {
         return null;
```
Stypox: As far as I know this is not good practice, an empty `List` should be returned instead. Though I am not sure if it is normal to see null instead of empty lists in NewPipeExtractor. Same 5 lines below

extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/linkHandler/MediaCCCStreamLinkHandlerFactory.java
```diff
@@ -1,16 +1,39 @@
...
 
     @Override
-    public String getId(String url) throws ParsingException {
-        if(url.startsWith("https://api.media.ccc.de/public/events/") &&
-            !url.contains("?q=")) {
-            return url.replace("https://api.media.ccc.de/public/events/", "");
+    public String getId(String urlString) throws ParsingException {
+        if (urlString.startsWith("https://api.media.ccc.de/public/events/") &&
+                !urlString.contains("?q=")) {
+            return urlString.replace("https://api.media.ccc.de/public/events/", "");
```
Stypox: Wouldn't it be a better idea to use substring? That's what we want, in the end

extractor/src/test/java/org/schabi/newpipe/extractor/services/media_ccc/MediaCCCStreamExtractorTest.java
```diff
@@ -21,81 +19,157 @@
...
+        }
+
+        @Test
+        public void testUploaderAvatarUrl() throws Exception {
+            assertEquals("https://static.media.ccc.de/media/events/gpn/gpn18/logo.png", extractor.getUploaderAvatarUrl());
+        }
+
+        @Test
+        public void testVideoStreams() throws Exception {
+            assertEquals(4, extractor.getVideoStreams().size());
```
Stypox: In these stream tests you should also check if all stream urls are secure urls (i.e. start with `https://`)

extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/linkHandler/MediaCCCStreamLinkHandlerFactory.java
B0pol: You can notice I didn't changed anything here, just the parameter name.
Anyway, I've done it, if I understood your request

extractor/src/test/java/org/schabi/newpipe/extractor/services/media_ccc/MediaCCCStreamExtractorTest.java
Stypox: for each stream run isSecureUrl(stream.getUrl())

extractor/src/test/java/org/schabi/newpipe/extractor/services/media_ccc/MediaCCCStreamExtractorTest.java
```diff
@@ -6,96 +6,182 @@
...
+        @Test
+        public void testUploaderAvatarUrl() throws Exception {
+            assertIsSecureUrl(extractor.getUploaderAvatarUrl());
+            assertEquals("https://static.media.ccc.de/media/events/gpn/gpn18/logo.png", extractor.getUploaderAvatarUrl());
+        }
+
+        @Test
+        public void testVideoStreams() throws Exception {
+            assertEquals(4, extractor.getVideoStreams().size());
+        }
```
Stypox: Please do something like this (also below and in all MediaCCC testVideoStreams and testAudioStreams functions).
The below code is untested
```suggestion
        public void testVideoStreams() throws Exception {
            List<VideoStream> streams = extractor.getVideoStreams();
            assertEquals(4, streams.size());
            for (VideoStream stream : streams) {
                assertIsSecureUrl(stream.getUrl())
            }
        }
```

TobiGr:
@wb9688 yes

PR #257, opened 2020-02-15
Re-introduce content language for YouTube

B0pol:

Reintroduced content language.
This fixes content language selector being useless, then titles and descriptions are now in the good language.
Fixes TeamNewPipe/NewPipe#3089

The only related problem is [channel subscription count](https://github.com/TeamNewPipe/NewPipe/issues/2632), so I fixed it this way: 
~~if the content language is not English and if the sub count is shortened, it makes a new request in English and get channel sub count.~~
We replace the abbreviation by its English equivalent with a HashMap, and then use the ``mixedNumberWordToLong`` function (as it is right now)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -107,8 +109,7 @@ public YoutubeStreamExtractor(StreamingService service, LinkHandler linkHandler)
     public String getName() throws ParsingException {
         assertPageFetched();
         try {
-            return playerResponse.getObject("videoDetails").getString("title");
-
+            return getString(playerResponse.getObject("microformat").getObject("playerMicroformatRenderer").getObject("title"), "simpleText");
```
TobiGr: what is the reason for changing the order?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -87,7 +89,7 @@
     private JsonObject playerArgs;
     @Nonnull
     private final Map<String, String> videoInfoPage = new HashMap<>();
-    private JsonObject playerResponse;
+    public JsonObject playerResponse;
```
TobiGr: why public?

B0pol: My bad, I used it for tests, forgot to let it private again

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
B0pol: > > The only related problem is channel subscription count, so I fixed it this way: if the content language is not English and if the sub count is shortened, it makes a new request in English and get channel sub count.
> 
> What is the exact problem? Do you get the wrong format or no result?

The problem is: Months ago, YouTube shortened sub count for channels, leading to no exact number. As it's shortened, it gives 250k, 1M… but in other languages, it could be 250 k (french), 250 somewhat (with space), and then we only gather the number, leading to https://github.com/TeamNewPipe/NewPipe/issues/2632
Enforcing english fixed this. But then broke title & description in the wrong language. Making a new request in English if the number is possibly wrong is the solution I thought

B0pol:
> > The only related problem is channel subscription count, so I fixed it this way: if the content language is not English and if the sub count is shortened, it makes a new request in English and get channel sub count.
> 
> What is the exact problem? Do you get the wrong format or no result?

The problem is: Months ago, YouTube shortened sub count for channels, leading to no exact number. As it's shortened, it gives 250k, 1M… but in other languages, it could be 250 k (french), 250 somewhat (with space), and then we only gather the number, leading to https://github.com/TeamNewPipe/NewPipe/issues/2632
Enforcing english fixed this. But then broke title & description in the wrong language. Making a new request in English if the number is possibly wrong is the solution I thought

TobiGr:
Ah yes, I remember. In this case, the best solution would be to create a list of the abbreviations for all supported languages and then convert the numbers correctly. Making a new request for a single value seems not the correct approach to me as we cause a bunch of traffic.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
B0pol: The reason for changing from videoDetails to microformat in JSON:
videoDetails gives the description in the original language.

The reason for changing to JSON first:
Bullet list are shown wrong with HTML.

See https://www.youtube.com/watch?v=nQTvw4_pzGk
Screenshot with official build:

and this screenshot (made with the test apk & this PR)

TobiGr: ok. thanks. Can you add this info as a comment?

B0pol: yeah

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
```diff
@@ -137,19 +142,32 @@ public String getFeedUrl() throws ParsingException {
...
-                return Utils.mixedNumberWordToLong(elTitle);
+                subCount = mixedNumberWordToLong(elTitle);
             } catch (NumberFormatException e) {
                 throw new ParsingException("Could not get subscriber count", e);
             }
-        } else {
-            // If the element is null, the channel have the subscriber count disabled
-            return -1;
+
+            if (!getExtractorLocalization().getLanguageCode().equals("en") && subCount < 1000 && subCount != -1) {
```
B0pol: > Ah yes, I remember. In this case, the best solution would be to create a list of the abbreviations for all supported languages and then convert the numbers correctly. Making a new request for a single value seems not the correct approach to me as we cause a bunch of traffic.

I move the discussion here @TobiGr 
How should we do?
I thought that:

1. Do a HashMap with
+ Key: abreviation in the language, with the space (or not if the language stick it)
+ value: English equivalent
2. Replace key by its value in the string
3. use mixedNumberWordToLong function.

Do you think of another better way?

TobiGr: > 1. Do a HashMap with
> - Key: abreviation in the language, with the space (or not if the language stick it)
> - value: English equivalent
>
>2. Replace key by its value in the string

I'd suggest to do this inside a second overloading `mixedNumberWordToLong(String numberWord, Locale locale)` method wich then calls the `mixedNumberWordToLong(String numberWord)` function.

B0pol: Done (wow that was hard and long)

extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
```diff
@@ -27,6 +33,24 @@ public static String removeNonDigitCharacters(String toRemove) {
         return toRemove.replaceAll("\\D+", "");
     }
 
+    /**
+     * <p>Remove a number from a string.</p>
+     * <p>Examples:</p>
+     * <ul>
+     *     <li>"123" -&gt; "123"</li>
```
TobiGr: I guess this is a typo and you mean
```suggestion
     *     <li>"123" -&gt; ""</li>
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeAbbreviationSubCountMap.java
```diff
@@ -0,0 +1,136 @@
...
+        abbreviationSubscribersCount.put(" مليون", englishMillionAbbreviation); //ar
+        abbreviationSubscribersCount.put(" mln", englishMillionAbbreviation); //az, et, lt, nl, pl, sq, uz
+        abbreviationSubscribersCount.put(" тыс", englishThousandAbbreviation); //be, ru
+        abbreviationSubscribersCount.put(" млн", englishMillionAbbreviation); //be, bg, kk, ky, ru, uk
+        abbreviationSubscribersCount.put(" хил", englishThousandAbbreviation); //bg
+        abbreviationSubscribersCount.put(" হা", englishThousandAbbreviation); //bn
+        abbreviationSubscribersCount.put(" hilj", englishThousandAbbreviation); //bs, sr
+        abbreviationSubscribersCount.put(" mil", englishMillionAbbreviation); //bs, cs, hr, ro, sk, sr-Latn
+        abbreviationSubscribersCount.put("m", englishThousandAbbreviation); //ca
+        abbreviationSubscribersCount.put(" M", englishMillionAbbreviation); //ca, es, eu, and many more
```
TobiGr: Why have some keys a leading whitespace and some not?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeAbbreviationSubCountMap.java
```diff
@@ -0,0 +1,136 @@
+package org.schabi.newpipe.extractor.services.youtube.extractors;
+
+import java.util.HashMap;
+
+public class YoutubeAbbreviationSubCountMap {
+
+    //should be safe until someone has 1 billion subscribers on YouTube
+    public static final HashMap<String, String> abbreviationSubscribersCount = new HashMap<>();
+    public static String englishMillionAbbreviation = "M";
```
TobiGr: I guess you can make those final, too

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
```diff
@@ -54,10 +54,6 @@
 
     private Document doc;
 
-    public Document getDoc() {
```
TobiGr: The question do we want to run the tests? If yes, we need this here.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeSubscriberTest.java
```diff
@@ -0,0 +1,1178 @@
...
+    @Test
+    public void testzum() throws IOException, ExtractionException {
+        YoutubeChannelExtractor extractor = getExtractor("m", new Localization("zu"));
+        ut(extractor);
+        assertEquals(countMillion, extractor.getSubscriberCount());
+    }
+
+    @Test
+    public void build() throws IOException, ExtractionException, InterruptedException {
+        Localization current = YoutubeService.SUPPORTED_LANGUAGES.get(79);
```
TobiGr:
```suggestion
        Localization current = YouTube.getSupportedLocalizations().get(79);
```

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeSubscriberTest.java
```diff
@@ -0,0 +1,1178 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.schabi.newpipe.extractor.ServiceList.YouTube;
+import static org.schabi.newpipe.extractor.utils.Utils.removeNumber;
+
+/**
+ * Test for {@link }
+ * and specifically YoutubeChannelExtractor.getSubscriberCount()
+ * in all the languages supported by YouTube.
+ * Takes a long time because we need to test make 146 requests to YouTube
+ * DON'T RUN ON MOBILE DATA
```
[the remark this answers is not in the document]
B0pol: I don't know what you want me to do here.
And yeah I'll write more why I let commented

extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
```diff
@@ -43,22 +68,66 @@ public static String removeNonDigitCharacters(String toRemove) {
...
-        } catch(ParsingException ignored) {}
+            multiplier = Parser.matchGroup("[\\d]+([\\.,][\\d]+)?([KMBkmb万লক億])+", numberWord, 2);
+        } catch (ParsingException ignored) {
+        }
         double count = Double.parseDouble(Parser.matchGroup1("([\\d]+([\\.,][\\d]+)?)", numberWord)
                 .replace(",", "."));
         switch (multiplier.toUpperCase()) {
             case "K":
                 return (long) (count * 1e3);
+            case "万": //10K
```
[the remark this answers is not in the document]
B0pol: You can watch the hashmap (the value says which abbreviation it is, and I commented which languages use it at the rigth)
For 10K (tenThousandAbbreviation as value on the hashmap), east-asian languages (ko = korean, ja = japanese, zh = chinese).
For 100K, indo-arabic languages (hundredThousandAbbreviation as value on the hashmap).
For 10M, indo-arabic languages
For 100M, east-asian languages

Here is the [Wikipedia page for the language codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeAbbreviationSubCountMap.java
B0pol: Because abbreviations are the same as Youtube give, except the final dot that is removed by the function removeNumber
Eg,
English 100K, French 100 K, German 100 Mio., so in the map there is also the whitespace.
And I guess and hope YouTube respects the punctuation, and punctuation changes with languages (for example in French I would put space before !, but in English it's stickied)

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeSubscriberTest.java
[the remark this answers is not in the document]
TobiGr: I didn't want you to do anything here. Just wanted to say that is better to not run the tests every time, but also keep the file in case we encounter a problem.

extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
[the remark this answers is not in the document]
TobiGr: Thanks can add this info here, too. And maybe add the Wikipedia list to the Map class

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
B0pol: We definitely don't want to run the tests everytime, because it takes 3mins with fiber for me, then the CI bot will be slow down for nothing.
This is the crash report we have if something is wrong:

## Exception
* __Request:__ https://www.youtube.com/channel/UChmpTxQ0Gp8KEJKsqzeg9Lg
* __Content Language:__ system
* __Service:__ YouTube
* __Version:__ 0.18.3


```
java.lang.NullPointerException: replacement == null
	at java.lang.String.replace(String.java:2239)
	at org.schabi.newpipe.extractor.utils.Utils.mixedNumberWordToLong(Utils.java:101)
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeChannelExtractor.getSubscriberCount(YoutubeChannelExtractor.java:155)
	at org.schabi.newpipe.extractor.channel.ChannelInfo.getInfo(ChannelInfo.java:87)
	at org.schabi.newpipe.extractor.channel.ChannelInfo.getInfo(ChannelInfo.java:47)
	at org.schabi.newpipe.util.-$$Lambda$ExtractorHelper$u5W7VszTe8AoEexIsFM9huQfbkM.call(Unknown Source:4)

```
</p></details>
<hr>
and it doesn't make the app crash, it just shows an error at the bottom and shows subscriber count unavailable.

(This was during test phase, I wanted to see what happens if something's broken).

And if something is broken, we could create this method (I added a comment mentioning this method)., run the test, easily get the abbreviation with build method at the bottom.

I think mentioning it in comments in enough, but tell me if I should recreate it.

extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
B0pol: thank you

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeAbbreviationSubCountMap.java
```diff
@@ -0,0 +1,136 @@
+package org.schabi.newpipe.extractor.services.youtube.extractors;
+
+import java.util.HashMap;
+
+public class YoutubeAbbreviationSubCountMap {
```
TobiGr: Can you please add a JavaDoc for this class?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
TobiGr: Mentioning the method in the comments should be enough. Thank you

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeAbbreviationSubCountMap.java
TobiGr: I am not sure about the class name, too. I assume that these abbreviations can be used by other services, too. Is this correct? If yes, please move this file to the extractor utils.

B0pol: Yes, it can be used if one wants to parse abbreviations from languages (and if some are missing for a service, it can easily be added).

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeSubscriberTest.java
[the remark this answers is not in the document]
B0pol: Should I comment the imports too? Or it makes no change?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeAbbreviationSubCountMap.java
B0pol: Is the JavaDoc good?

extractor/src/main/java/org/schabi/newpipe/extractor/utils/AbbreviationHashMap.java
```diff
@@ -1,13 +1,36 @@
-package org.schabi.newpipe.extractor.services.youtube.extractors;
+package org.schabi.newpipe.extractor.utils;
 
 import java.util.HashMap;
 
-public class YoutubeAbbreviationSubCountMap {
+
+/**
+ * Map matching abbreviations with their English equivalents
```
TobiGr: The class only contains maps

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeAbbreviationSubCountMap.java
TobiGr: yes, looks good. however, the class is not a map. we should rename it and change that part in the doc, too.

extractor/src/main/java/org/schabi/newpipe/extractor/utils/AbbreviationHashMap.java
B0pol: You wanna rename it to what?

TobiGr: Sorry, utils is the wrong package here. Should be placed in localization and probably called AbbreviationHelper. @Stypox Do you have an idea for a proper name?

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeSubscriberTest.java
```diff
@@ -0,0 +1,1180 @@
...
+Commenting the whole file because otherwise it will slow down too much the CI test.
+And also, often one up to three tests fail if you launch the whole tests, because some requests fail
+(it could be reCAPTCHAs) but they, as of today (2020-02-16) success if you run each one.
+
+To run the test (to investigate maybe future problems),
+You should temporarily create « public Document getDoc() { return this.doc;} » method in YoutubeChannelExtractor.
+ */
+
+/*
+public class YoutubeSubscriberTest {
```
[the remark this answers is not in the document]
B0pol: > It takes every language and compares it to the english case (the default one), this way we can guarantee that the values parsed by the others are correct.

That's exactly what this file does.

> That way we can use less requests and test multiple things at once

Having one test for one thing is great for this situation, because you instantly know what failed with the pattern for function names I gave.
But I don't know what you did, maybe indeed it's better.

> Using the other approach, I already find some missing patterns here as well.

Can you share the file you used for the other approach?

> PS: Please don't commit a commented file. Consider the annotation Ignore passing the correct reason.
Thank you

extractor/src/test/java/org/schabi/newpipe/extractor/utils/UtilsTest.java
```diff
@@ -15,4 +19,10 @@ public void testMixedNumberWordToLong() throws JsonParserException, ParsingExcep
         assertEquals(10.5e6, Utils.mixedNumberWordToLong("10,5M"), 0.0);
         assertEquals(1.5e9, Utils.mixedNumberWordToLong("1,5B"), 0.0);
     }
+
+    public static void createFile(String path, String content) throws IOException {
```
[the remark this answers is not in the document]
B0pol: No. I'm using often a function of this kind when testing things in NewPipeExtractor.
I thought it could be useful for everyone and then I put it here.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
```diff
@@ -137,19 +142,18 @@ public String getFeedUrl() throws ParsingException {
 
     @Override
     public long getSubscriberCount() throws ParsingException {
-
+        long subCount = -1;
         final Element el = doc.select("span[class*=\"yt-subscription-button-subscriber-count\"]").first();
         if (el != null) {
+            // If the element is null, the channel have the subscriber count disabled
```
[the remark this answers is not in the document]
B0pol: No it is at the good place now. Look at that:
image
In class EminemVEVO in YoutubeChannelExtractorTest

extractor/src/main/java/org/schabi/newpipe/extractor/utils/AbbreviationHashMap.java
```diff
@@ -0,0 +1,156 @@
...
+        abbreviationSubscribersCount.put(" milj", englishMillionAbbreviation); //fi, lv
+        abbreviationSubscribersCount.put(" մլն", englishMillionAbbreviation); //hy
+        abbreviationSubscribersCount.put(" jt", englishMillionAbbreviation); //id
+        abbreviationSubscribersCount.put(" Mln", englishMillionAbbreviation); //it
+        abbreviationSubscribersCount.put(" მლნ", englishMillionAbbreviation); //ka
+        abbreviationSubscribersCount.put(" លាន", englishMillionAbbreviation); //km
+        abbreviationSubscribersCount.put("ಮಿ", englishMillionAbbreviation); //kn
+        abbreviationSubscribersCount.put(" ລ້ານ", englishMillionAbbreviation); //lo
+        abbreviationSubscribersCount.put(" М", englishMillionAbbreviation); //mk. It isn't a space but a
+        // narrow non-breaking space, ie U+202F
```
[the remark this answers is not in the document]
B0pol: Done

extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
```diff
@@ -43,22 +68,66 @@ public static String removeNonDigitCharacters(String toRemove) {
...
+
+        try { //special cases where it gives a number directly for some languages, or with a dot or a comma, or space
+            String maybeAlreadyNumber = numberWord.replaceAll("([ .,])", ""); //dot, comma or narrow non-breaking space, ie U+202Fw
+            return Long.parseLong(maybeAlreadyNumber);
+        } catch (NumberFormatException e) {
+            //the number had an abbreviation, so it will be handled below
+        }
+
+        if (!langCode.equals("en")) {
+            numberWord = numberWord.replace(abbreviation, abbreviationSubscribersCount.get(abbreviation));
```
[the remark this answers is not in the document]
B0pol: Done. I will edit with the crash report we would have if one fails

extractor/src/main/java/org/schabi/newpipe/extractor/utils/AbbreviationHashMap.java
```diff
@@ -0,0 +1,156 @@
...
+ * hundred thousand by {@link #hundredThousandAbbreviation},
+ * ten million by {@link #tenMillionAbbreviation},
+ * hundred million by {@link #hundredMillionAbbreviation}.
+ * </p>
+ * <p>
+ * The languages using the abbreviation is commented with the language code at the left.
+ * @see <a href="https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes">Wikipedia page of language codes</a>
+ * </p>
+ */
+public class AbbreviationHashMap {
```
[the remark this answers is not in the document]
B0pol: I don't really understand. You want me to gather data from Unicode, instead of tests?
So I would have to check Unicode, find if it's not already in the map, add them if, and do that 80 times?
It would be faster to find channels at each range, run tests, see failed one, if we get error rerun / add in the map.

extractor/src/main/java/org/schabi/newpipe/extractor/utils/AbbreviationHashMap.java
```diff
@@ -0,0 +1,156 @@
+package org.schabi.newpipe.extractor.utils;
+
+import java.util.HashMap;
+
+
+/**
+ * Map matching abbreviations with their English equivalents
+ *
+ * @author B0pol
```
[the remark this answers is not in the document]
B0pol: done

extractor/src/main/java/org/schabi/newpipe/extractor/utils/AbbreviationHashMap.java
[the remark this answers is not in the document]
B0pol: done

extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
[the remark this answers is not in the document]
B0pol: Done. Here is the crash report we would have
## Exception
* __Request:__ https://www.youtube.com/channel/UCJtvlB_Z-Sge7Ok-NxCjbqg
* __Content Language:__ system
* __Service:__ YouTube
* __Version:__ 0.18.3


```
org.schabi.newpipe.extractor.exceptions.ParsingException: The abbreviation "M" is missing in AbbreviationHelper map
	at org.schabi.newpipe.extractor.utils.Utils.mixedNumberWordToLong(Utils.java:126)
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeChannelExtractor.getSubscriberCount(YoutubeChannelExtractor.java:148)
	at org.schabi.newpipe.extractor.channel.ChannelInfo.getInfo(ChannelInfo.java:87)
	at org.schabi.newpipe.extractor.channel.ChannelInfo.getInfo(ChannelInfo.java:47)
	at org.schabi.newpipe.util.-$$Lambda$ExtractorHelper$u5W7VszTe8AoEexIsFM9huQfbkM.call(Unknown Source:4)

```
</p></details>
<hr>

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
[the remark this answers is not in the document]
B0pol: See this screenshot:
image
So no, it is at the right place.

extractor/src/main/java/org/schabi/newpipe/extractor/utils/AbbreviationHashMap.java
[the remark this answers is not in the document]
B0pol: I don't understand. You want me to check unicode, check if something is missing, and add it to the map, and **do that 80 times?**
At this point, it would be faster to run tests with more channels, and I will do it (until you explain me what you mean by the other approach.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeSubscriberTest.java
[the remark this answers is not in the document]
B0pol: For example, yeah ``hi`` fails, but you have 29 tests each time, when I run the test it's not clear which one failed, or even if the whole failed.

Also, of course you can test more, because you test 29 videos with a request, because you request the feed. That makes 2320 test with only 80 request.
For the subscriber count, I'm using two request for each language.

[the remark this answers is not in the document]
B0pol: Hum, found how to test 112 times for a request.
https://www.youtube.com/feed/guide_builder

B0pol:
So, I made another test file with [YouTube discover page](https://www.youtube.com/feed/guide_builder), we test about 112 channels for each 80 languages. It just test abbreviations though, as it's not a channel page.

I included below tests for the 80 languages, you can easily test a channel. The downside is that there are more false negatives.

The map is fullfilled now, the crash report is straightforward ([here](https://github.com/TeamNewPipe/NewPipeExtractor/pull/257#discussion_r380130886)), it doesn't make the app crash, and there is an easy workaround for users if somehow one language fails: switch content language to English (until next update).

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeSubscriberTest.java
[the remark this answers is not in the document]
B0pol: Look at my last comment & commit, I refactored the test, it's should be better know.

extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
```diff
@@ -44,22 +68,77 @@ public static String removeNonDigitCharacters(String toRemove) {
...
+            case "億": //100M, used by east-asian languages
+                return (long) (count * 1e8);
             case "B":
                 return (long) (count * 1e9);
             default:
                 return (long) (count);
         }
     }
 
+    public static String cleanWhiteSpaces(String s) {
```
TobiGr: removeWhiteSpaces

extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
```diff
@@ -44,22 +68,77 @@ public static String removeNonDigitCharacters(String toRemove) {
...
+        return s.replaceAll("(\\s| | )", "");
+    }
+
+    //does the same as the function above, but for the 80 languages supported by YouTube.
```
TobiGr: Please make this a JDoc

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeSubcriberTest.java
```diff
@@ -0,0 +1,209 @@
...
+import static org.schabi.newpipe.extractor.ServiceList.YouTube;
+import static org.schabi.newpipe.extractor.localization.AbbreviationHelper.abbreviationSubscribersCount;
+import static org.schabi.newpipe.extractor.utils.Utils.cleanWhiteSpaces;
+import static org.schabi.newpipe.extractor.utils.Utils.removeNumber;
+
+/**
+ * A class that tests abbreviations and subscriber counts for all the languages YouTube supports.
+ */
+@Ignore("Should be ran manually from time to time, as it's too time consuming.")
+public class YoutubeSubcriberTest {
```
TobiGr: small typo. this should be YoutubeSub**s**criberTest

extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
```diff
@@ -44,22 +68,84 @@ public static String removeNonDigitCharacters(String toRemove) {
...
-            multiplier = Parser.matchGroup("[\\d]+([\\.,][\\d]+)?([KMBkmb])+", numberWord, 2);
-        } catch(ParsingException ignored) {}
+            multiplier = Parser.matchGroup("[\\d]+([\\.,][\\d]+)?([KMBkmb万লক億])+", numberWord, 2);
+        } catch (ParsingException ignored) {
+        }
         double count = Double.parseDouble(Parser.matchGroup1("([\\d]+([\\.,][\\d]+)?)", numberWord)
                 .replace(",", "."));
         switch (multiplier.toUpperCase()) {
             case "K":
                 return (long) (count * 1e3);
```
Stypox: `1e3` is more readable than `1000`, but it could cause issues since it is a floating point number. Floats are not perfectly precise, so we could get that `(long) (x * 1e3) != x * 1000`, because `x * 1e3` could be a little bit less (i.e. 0.0000001 less) than `x * 1000` so when converting it to long, that just floors the value, we would get `x * 1000 - 1`. So I think it would be better here and in the following lines to use `1000`, `10000`, `100000`, etc.

B0pol: Ah. I got once while testing a value using  
```java
case "ল": //100K, used by indo-arabic languages
                return (long) (count * 1e5);
```
or 10M, I don't know, with only 1 difference between the English count and the language count (it was like a number, let's say 16M, with English I got 16M, with these languages I got 16M - 1).

I think the best for readability is to use underscores:
https://docs.oracle.com/javase/7/docs/technotes/guides/language/underscores-literals.html

TobiGr:
@Stypox @mauriciocolli When you think that this is good to go, please merge.

Stypox:
@B0pol [Tarvis](https://travis-ci.org/TeamNewPipe/NewPipeExtractor/builds/652566427?utm_source=github_status&utm_medium=notification) gives two warnings. Could you fix them?
```
linkhandler/SearchQueryHandlerFactory.java:48: warning - @return tag has no arguments.
utils/Utils.java:108: warning - @param argument "loc:" is not a parameter name.
```


Also, there is an error with full links in description, could this be related to your changes?
```
org.schabi.newpipe.extractor.services.youtube.stream.YoutubeStreamExtractorDefaultTest$DescriptionTestUnboxing > testGetFullLinksInDescription FAILED
```

B0pol:
Because I switched to raw text, instead of HTML, so full links are not provided.

Stypox:
That's bad, we need the full links provided in the html, otherwise long links in the description won't work... Are you sure there is no way to fix HTML formatting?

Stypox:
Oh, I saw you edited your comment... Yeah, that would be ok ;-)
Are you sure other urls are not abbreviated?

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/stream/YoutubeStreamExtractorDefaultTest.java
```diff
@@ -249,15 +250,18 @@ public void testGetDescription() throws ParsingException {
...
-            assertFalse(extractor.getDescription().getContent().contains("https://youtu.be/XxaRBPyrnBU?list=PL7..."));
-            assertFalse(extractor.getDescription().getContent().contains("https://youtu.be/U-9tUEOFKNU?list=PL7..."));
+            Description description = extractor.getDescription();
+            String content = description.getContent();
+            if (description.getType() == Description.HTML) {
+                assertTrue(content.contains("https://www.youtube.com/watch?v=X7FLCHVXpsA&list=PL7u4lWXQ3wfI_7PgX0C-VTiwLeu0S4v34"));
+                assertTrue(content.contains("https://www.youtube.com/watch?v=Lqv6G0pDNnw&amp;list=PL7u4lWXQ3wfI_7PgX0C-VTiwLeu0S4v34"));
+                assertTrue(content.contains("https://www.youtube.com/watch?v=XxaRBPyrnBU&amp;list=PL7u4lWXQ3wfI_7PgX0C-VTiwLeu0S4v34"));
+                assertTrue(content.contains("https://www.youtube.com/watch?v=U-9tUEOFKNU&amp;list=PL7u4lWXQ3wfI_7PgX0C-VTiwLeu0S4v34"));
+            }
```
Stypox: A test for when the description is PLAIN_TEXT should be made:
```suggestion
            } else if (description.getType() == Description.PLAIN_TEXT) {
                assertTrue(content.contains("https://youtu.be/X7FLCHVXpsA?list=PL7u4lWXQ3wfI_7PgX0C-VTiwLeu0S4v34"));
                assertTrue(content.contains("https://youtu.be/Lqv6G0pDNnw?list=PL7u4lWXQ3wfI_7PgX0C-VTiwLeu0S4v34"));
                assertTrue(content.contains("https://youtu.be/XxaRBPyrnBU?list=PL7u4lWXQ3wfI_7PgX0C-VTiwLeu0S4v34"));
                assertTrue(content.contains("https://youtu.be/U-9tUEOFKNU?list=PL7u4lWXQ3wfI_7PgX0C-VTiwLeu0S4v34"));
            }
```

B0pol:
The main reason I switched to HTML if error was it's not the good language, not really formatting.
I tested, and links in description are OK.
I didn't find other shortened link.

But this is weird. Look at [the video](https://www.youtube.com/watch?v=cV5TjZCJkuA), it's only youtu.be links.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/stream/YoutubeStreamExtractorDefaultTest.java
B0pol: I don't understand. They are already tested below, at the end of the if. And the HTML way pass the test with them at the bottom.

Stypox:
Ok, then maybe the issue has to do with the video (if that's the case ignore my review and please revert the changes in the description test and then replace youtube.com with youtu.be).
Could you test with this video? It has a long link in the description: https://www.youtube.com/watch?v=gd5bynDvDUw

B0pol:
on YouTube website it has no full links:
image

Stypox:
The HTML description converts abbreviated links into the correct ones (see `parseHtmlAndGetFullLinks`: https://github.com/TeamNewPipe/NewPipeExtractor/blob/ea68770ee1d5ed8367081257274137ddb5371938/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java#L219)

B0pol:
What's wrong with having shortened links?

Stypox:
I think we are misunderstanding ourselves ;-)
By "full links" I mean "links that are not abbreviated using `...`". Those cannot be clicked in NewPipe descriptions since the full url is missing. Shortened youtube urls (i.e. "youtu.be"), on the other hand, have nothing wrong and work without problems.

With the new JSON method I am not sure if full links are provided or if there are `...`. Could you check that the JSON microformat description for video "https://www.youtube.com/watch?v=gd5bynDvDUw" contains the full link "https://www.youtube.com/channel/UCf5q0cbFOLbphljteZ9d4Pw" and not "https://www.youtube.com/channel/UCf5q..."

Sorry for my misunderstanding :man_facepalming:

B0pol:
Oh ok I understand. Yeah they are ok,

B0pol:
Yes they are ok. (Why close button is exactly at the same place as cancel on issues???)

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/stream/YoutubeStreamExtractorDefaultTest.java
Stypox: If I understood correctly, the HTML description has the same shortened links as the JSON microformat description. So  this should be (note there is no `if`, i.e. no need to distinguish between formats):
```
                assertTrue(content.contains("https://youtu.be/X7FLCHVXpsA?list=PL7u4lWXQ3wfI_7PgX0C-VTiwLeu0S4v34"));
                assertTrue(content.contains("https://youtu.be/Lqv6G0pDNnw?list=PL7u4lWXQ3wfI_7PgX0C-VTiwLeu0S4v34"));
                assertTrue(content.contains("https://youtu.be/XxaRBPyrnBU?list=PL7u4lWXQ3wfI_7PgX0C-VTiwLeu0S4v34"));
                assertTrue(content.contains("https://youtu.be/U-9tUEOFKNU?list=PL7u4lWXQ3wfI_7PgX0C-VTiwLeu0S4v34"));
```

B0pol: test this:
```java
@Test
public void test() throws IOException, ReCaptchaException {
    Downloader dl = NewPipe.getDownloader();
    Response response = dl.get("https://www.youtube.com/watch?v=cV5TjZCJkuA");
    Document doc = Jsoup.parse(response.responseBody());
    System.out.println(doc.select("#eow-description"));
}
```
Basically, all links are like that
```<a href="/watch?v=X7FLCHVXpsA&amp;list=PL7u4lWXQ3wfI_7PgX0C-VTiwLeu0S4v34" class="yt-uix-sessionlink       spf-link " data-sessionlink="itct=CCwQ6TgYACITCNb42KP44ucCFZYWBgAdKaEKRA">https://youtu.be/X7FLCHVXpsA?list=PL7...</a>```
If we don't use ``parseHtmlAndGetFullLinks``, we get this: <a href="/watch?v=X7FLCHVXpsA&amp;list=PL7u4lWXQ3wfI_7PgX0C-VTiwLeu0S4v34" class="yt-uix-sessionlink       spf-link " data-sessionlink="itct=CCwQ6TgYACITCNb42KP44ucCFZYWBgAdKaEKRA">https://youtu.be/X7FLCHVXpsA?list=PL7...</a> broken because the link in href needs the domain, which is not youtube because we aren't youtube.

That's why we use ``parseHtmlAndGetFullLinks``, to gather what is in href, and fix the link.

For JSON, it's not needed, because it gives the proper link directly.

If I do your requested changes, it won't work because HTML gives full links, JSON shortened.
What we don't want, is to have broken shortened links, like this <a href="/watch?v=X7FLCHVXpsA&amp;list=PL7u4lWXQ3wfI_7PgX0C-VTiwLeu0S4v34" class="yt-uix-sessionlink       spf-link " data-sessionlink="itct=CCwQ6TgYACITCNb42KP44ucCFZYWBgAdKaEKRA">https://youtu.be/X7FLCHVXpsA?list=PL7...</a> and test if the links are good.

So we need a if.

B0pol:
Otherwise, about this pr: yt_new, ie #258 breaks again, by adding ``subscribers`` to the count, eg ``100K subscribers``.
I think it break could break more things because i've seen other places where they added like ``views`` (but I think remove non-digit number still do the job as it's not rounded).

As it's in March, so soon, don't merge this, it will be pretty useless if we have to then comment the supported languages because it breaks again.
I'll wait for yt_new to be merged, and try to fix again.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/stream/YoutubeStreamExtractorDefaultTest.java
wb9688: Can't you just do `new URL(new URL(whateverFunctionGetsTheUrlOfTheVideo()), whateverVariableHasTheHref)`? Or is that exactly what that function does?

B0pol: Which function? The test one? If yes, you are right

Stypox:
Ok

PR #261, opened 2020-02-25
Update YouTube to material version 

### LINKED ISSUE #258: Heads up, YouTube discontinuing older version on desktop soon
This might affect us.

> Enter 2020 and the older versions are missing many of the new features and design improvements we’ve introduced over the past 3 years, including top requests based on your feedback (see most recent update here). That's why the older version will be going away in March and you'll only be able to access newer desktop versions to enjoy the best of YouTube.

https://support.google.com/youtube/thread/27596769

TobiGr:
Fixes #258.
We will add more details later.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
@@ -52,7 +52,7 @@ public class YoutubeChannelExtractor extends ChannelExtractor {
     private static final String CHANNEL_URL_PARAMETERS = "/videos?view=0&flow=list&sort=dd&live_view=10000";
 
     private Document doc;
-    private JsonObject ytInitialData;
+    private JsonObject initialData;
 
     public YoutubeChannelExtractor(StreamingService service, ListLinkHandler linkHandler) {
         super(service, linkHandler);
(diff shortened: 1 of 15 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
```diff
@@ -59,11 +67,13 @@ public void onFetchPage(@Nonnull Downloader downloader) throws IOException, Extr
         String channelUrl = super.getUrl() + CHANNEL_URL_PARAMETERS;
         final Response response = downloader.get(channelUrl, getExtractorLocalization());
         doc = YoutubeParsingHelper.parseAndCheckPage(channelUrl, response);
+        initialData = YoutubeParsingHelper.getInitialData(response.responseBody());
```
mauriciocolli: @TobiGr @wb9688 

I noticed that this page is very heavy to download, about 300\~400KB per request.

What do you think of using the `pbj` parameter? Using it offers a big advantage: it's a JSON formatted response which is, as you might've imagine, much less error prone to parse (it is basically the `initialData` already extracted).

Another thing is that the size would be as tiny, or even tinier than it was, a `gzip`ed 15\~20KB response.

Here's an example using the `/videos` endpoint (could be another one as well or even the "root", but this one seems to be sufficient to parse all the things we need, like description, name, etc.):

```bash
curl 'https://www.youtube.com/channel/UCLA_DiR1FfKNvjuUpBHmylQ/videos?pbj=1' \
    -H 'x-youtube-client-name: 1' -H 'x-youtube-client-version: 2.20200224.11.00' \
    -H 'User-Agent: Mozilla/5.0 (Windows NT 10.0; WOW64; rv:68.0) Gecko/20100101 Firefox/68.0' \
    -H 'Accept-Encoding: gzip' --compressed -v | jq > videos.pbj1.json
```

It looks like the only headers that we need are those two, which maybe we can just hardcode it (who knows how long it remains valid) or even find some lightweight method to get it dynamically.

I'll investigate more about it later.

**PS**: This is used throughout YouTube, replacing the channel with a video url works like a charm. Worth investigating as well.

wb9688: We're using pbj for YoutubeSearchExtractor's next pages iirc, I just didn't realize that's possible with all pages. I don't think we could do that with YoutubeStreamExtractor though.

PR #262, opened 2020-02-26
Improve yt_new

### LINKED ISSUE #264: Broke Fix For Playlists generated without uploader (Regression)
around [here](https://github.com/TeamNewPipe/NewPipeExtractor/blob/8838e2d1361a24aa96ce5c39c8e45d6e4c3725bb/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubePlaylistExtractor.java#L52)

breaks #219 

this regression prevents youtube music playlists from working.

wb9688:

Things I want to do in this PR:

- [x] Use pbj=1 in YoutubeSearchExtractor
- [x] Use pbj=1 in YoutubeChannelExtractor
- [x] Use pbj=1 in YoutubePlaylistExtractor
- [x] Use pbj=1 in YoutubeTrendingExtractor
- [x] Use pbj=1 in YoutubeStreamExtractor ~(not possible due to age restriction stuff?)~
- [x] Improve getClientVersion() so that it's similar to the clientId() for SoundCloud
~Use new YouTube site for extracting comments~
- [x] Fix existing tests
- [x] Fix TeamNewPipe/NewPipe#3145 (trending page not containing everything)
- [x] Correctly parse links in e.g. https://www.youtube.com/watch?v=cV5TjZCJkuA
- [x] Fix TeamNewPipe/NewPipe#3144 (pictures of certain channels)
- [x] Fix #264 (YouTube Music playlists)
- [x] Fix exception for channels without Videos tab
- [x] Fix getUploaderUrl() in YoutubeStreamInfoItemExtractor
- [x] Fix video titles with @ (and probably #)
- [x] Make one helping function for text (as YouTube often provides that either as a `simpleText` or a `runs` array with `text` in them; this would prevent issues like those video titles or channel descriptions)
- [x] Fix channel descriptions
- [x] Fix subscribers count for channels with 0 subscribers
- [x] Fix channels with multiple views on Videos tab
- [x] Fix thumbnail URLs starting with //
- [x] Fix TeamNewPipe/NewPipe#3153 (some search queries having 0 results)

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
@@ -25,6 +25,7 @@ import java.util.Map;
 
 import javax.annotation.Nonnull;
 
+import static org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeParsingHelper.getTextFromObject;
 import static org.schabi.newpipe.extractor.utils.Utils.HTTP;
 import static org.schabi.newpipe.extractor.utils.Utils.HTTPS;
 
@@ -124,8 +125,20 @@ public class YoutubeChannelExtractor extends ChannelExtractor {
     @Override
     public String getAvatarUrl() throws ParsingException {
         try {
-            return initialData.getObject("header").getObject("c4TabbedHeaderRenderer").getObject("avatar")
+            String url = initialData.getObject("header").getObject("c4TabbedHeaderRenderer").getObject("avatar")
                     .getArray("thumbnails").getObject(0).getString("url");
+
+            // the first characters of the avatar URLs are different for each channel and some are not even valid URLs
+            if (url.startsWith("//")) {
+                url = url.substring(2);
+            }
+            if (url.startsWith(HTTP)) {
+                url = Utils.replaceHttpWithHttps(url);
+            } else if (!url.startsWith(HTTPS)) {
+                url = HTTPS + url;
+            }
+
+            return url;
         } catch (Exception e) {
             throw new ParsingException("Could not get avatar", e);
         }
@@ -172,7 +185,7 @@ public class YoutubeChannelExtractor extends ChannelExtractor {
         final JsonObject subscriberInfo = initialData.getObject("header").getObject("c4TabbedHeaderRenderer").getObject("subscriberCountText");
         if (subscriberInfo != null) {
             try {
-                return Utils.mixedNumberWordToLong(subscriberInfo.getArray("runs").getObject(0).getString("text"));
+                return Utils.mixedNumberWordToLong(getTextFromObject(subscriberInfo));
             } catch (NumberFormatException e) {
                 throw new ParsingException("Could not get subscriber count", e);
             }
@@ -301,10 +314,10 @@ public class YoutubeChannelExtractor extends ChannelExtractor {
         }
 
         try {
-            if (videoTab.getObject("content").getObject("sectionListRenderer").getArray("contents")
-                    .getObject(0).getObject("itemSectionRenderer").getArray("contents")
-                    .getObject(0).getObject("messageRenderer").getObject("text").getArray("runs")
-                    .getObject(0).getString("text").equals("This channel has no videos."))
+            if (getTextFromObject(videoTab.getObject("content").getObject("sectionListRenderer")
+                    .getArray("contents").getObject(0).getObject("itemSectionRenderer")
+                    .getArray("contents").getObject(0).getObject("messageRenderer")
+                    .getObject("text")).equals("This channel has no videos."))
                 return null;
         } catch (Exception ignored) {}
 
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
@@ -5,11 +5,9 @@ import com.grack.nanojson.JsonObject;
 import com.grack.nanojson.JsonParser;
 import com.grack.nanojson.JsonParserException;
 
-import org.jsoup.nodes.Document;
 import org.schabi.newpipe.extractor.StreamingService;
 import org.schabi.newpipe.extractor.channel.ChannelExtractor;
 import org.schabi.newpipe.extractor.downloader.Downloader;
-import org.schabi.newpipe.extractor.downloader.Response;
 import org.schabi.newpipe.extractor.exceptions.ExtractionException;
 import org.schabi.newpipe.extractor.exceptions.ParsingException;
 import org.schabi.newpipe.extractor.linkhandler.ListLinkHandler;
@@ -68,10 +66,8 @@ public class YoutubeChannelExtractor extends ChannelExtractor {
 
         Map<String, List<String>> headers = new HashMap<>();
         headers.put("X-YouTube-Client-Name", Collections.singletonList("1"));
-        // Use the hardcoded client version first to get JSON with a structure we know
-        // TODO: Use YoutubeParsingHelper.getClientVersion() as fallback
         headers.put("X-YouTube-Client-Version",
-                Collections.singletonList(YoutubeParsingHelper.HARDCODED_CLIENT_VERSION));
+                Collections.singletonList(YoutubeParsingHelper.getClientVersion()));
         final String response = getDownloader().get(url, headers, getExtractorLocalization()).responseBody();
         if (response.length() < 50) { // ensure to have a valid response
             throw new ParsingException("Could not parse json data for next streams");
@@ -222,10 +218,8 @@ public class YoutubeChannelExtractor extends ChannelExtractor {
 
         Map<String, List<String>> headers = new HashMap<>();
         headers.put("X-YouTube-Client-Name", Collections.singletonList("1"));
-        // Use the hardcoded client version first to get JSON with a structure we know
-        // TODO: Use YoutubeParsingHelper.getClientVersion() as fallback
         headers.put("X-YouTube-Client-Version",
-                Collections.singletonList(YoutubeParsingHelper.HARDCODED_CLIENT_VERSION));
+                Collections.singletonList(YoutubeParsingHelper.getClientVersion()));
         final String response = getDownloader().get(pageUrl, headers, getExtractorLocalization()).responseBody();
         if (response.length() < 50) { // ensure to have a valid response
             throw new ParsingException("Could not parse json data for next streams");
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
@@ -2,8 +2,6 @@ package org.schabi.newpipe.extractor.services.youtube.extractors;
 
 import com.grack.nanojson.JsonArray;
 import com.grack.nanojson.JsonObject;
-import com.grack.nanojson.JsonParser;
-import com.grack.nanojson.JsonParserException;
 
 import org.schabi.newpipe.extractor.StreamingService;
 import org.schabi.newpipe.extractor.channel.ChannelExtractor;
@@ -19,14 +17,11 @@ import org.schabi.newpipe.extractor.stream.StreamInfoItemsCollector;
 import org.schabi.newpipe.extractor.utils.Utils;
 
 import java.io.IOException;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
 
 import javax.annotation.Nonnull;
 
 import static org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeParsingHelper.fixThumbnailUrl;
+import static org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeParsingHelper.getJsonResponse;
 import static org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeParsingHelper.getTextFromObject;
 
 /*
@@ -62,22 +57,7 @@ public class YoutubeChannelExtractor extends ChannelExtractor {
     public void onFetchPage(@Nonnull Downloader downloader) throws IOException, ExtractionException {
         final String url = super.getUrl() + "/videos?pbj=1&view=0&flow=grid";
 
-        JsonArray ajaxJson;
-
-        Map<String, List<String>> headers = new HashMap<>();
-        headers.put("X-YouTube-Client-Name", Collections.singletonList("1"));
-        headers.put("X-YouTube-Client-Version",
-                Collections.singletonList(YoutubeParsingHelper.getClientVersion()));
-        final String response = getDownloader().get(url, headers, getExtractorLocalization()).responseBody();
-        if (response.length() < 50) { // ensure to have a valid response
-            throw new ParsingException("Could not parse json data for next streams");
-        }
-
-        try {
-            ajaxJson = JsonParser.array().from(response);
-        } catch (JsonParserException e) {
-            throw new ParsingException("Could not parse json data for next streams", e);
-        }
+        final JsonArray ajaxJson = getJsonResponse(url);
 
         initialData = ajaxJson.getObject(1).getObject("response");
     }
@@ -214,22 +194,7 @@ public class YoutubeChannelExtractor extends ChannelExtractor {
         fetchPage();
 
         StreamInfoItemsCollector collector = new StreamInfoItemsCollector(getServiceId());
-        JsonArray ajaxJson;
-
-        Map<String, List<String>> headers = new HashMap<>();
-        headers.put("X-YouTube-Client-Name", Collections.singletonList("1"));
-        headers.put("X-YouTube-Client-Version",
-                Collections.singletonList(YoutubeParsingHelper.getClientVersion()));
-        final String response = getDownloader().get(pageUrl, headers, getExtractorLocalization()).responseBody();
-        if (response.length() < 50) { // ensure to have a valid response
-            throw new ParsingException("Could not parse json data for next streams");
-        }
-
-        try {
-            ajaxJson = JsonParser.array().from(response);
-        } catch (JsonParserException e) {
-            throw new ParsingException("Could not parse json data for next streams", e);
-        }
+        final JsonArray ajaxJson = getJsonResponse(pageUrl);
 
         JsonObject sectionListContinuation = ajaxJson.getObject(1).getObject("response")
                 .getObject("continuationContents").getObject("gridContinuation");
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
@@ -85,6 +85,7 @@ public class YoutubeChannelExtractor extends ChannelExtractor {
 
     @Override
     public String getNextPageUrl() throws ExtractionException {
+        if (getVideoTab() == null) return "";
         return getNextPageUrlFrom(getVideoTab().getObject("content").getObject("sectionListRenderer")
                 .getArray("contents").getObject(0).getObject("itemSectionRenderer")
                 .getArray("contents").getObject(0).getObject("gridRenderer").getArray("continuations"));
@@ -195,10 +196,12 @@ public class YoutubeChannelExtractor extends ChannelExtractor {
     public InfoItemsPage<StreamInfoItem> getInitialPage() throws ExtractionException {
         StreamInfoItemsCollector collector = new StreamInfoItemsCollector(getServiceId());
 
-        JsonArray videos = getVideoTab().getObject("content").getObject("sectionListRenderer").getArray("contents")
-                .getObject(0).getObject("itemSectionRenderer").getArray("contents").getObject(0)
-                .getObject("gridRenderer").getArray("items");
-        collectStreamsFrom(collector, videos);
+        if (getVideoTab() != null) {
+            JsonArray videos = getVideoTab().getObject("content").getObject("sectionListRenderer").getArray("contents")
+                    .getObject(0).getObject("itemSectionRenderer").getArray("contents").getObject(0)
+                    .getObject("gridRenderer").getArray("items");
+            collectStreamsFrom(collector, videos);
+        }
 
         return new InfoItemsPage<>(collector, getNextPageUrl());
     }
@@ -241,9 +244,7 @@ public class YoutubeChannelExtractor extends ChannelExtractor {
 
 
     private String getNextPageUrlFrom(JsonArray continuations) {
-        if (continuations == null) {
-            return "";
-        }
+        if (continuations == null) return "";
 
         JsonObject nextContinuationData = continuations.getObject(0).getObject("nextContinuationData");
         String continuation = nextContinuationData.getString("continuation");
@@ -294,6 +295,14 @@ public class YoutubeChannelExtractor extends ChannelExtractor {
         if (videoTab == null) {
             throw new ParsingException("Could not find Videos tab");
         }
+
+        try {
+            if (videoTab.getObject("content").getObject("sectionListRenderer").getArray("contents")
+                    .getObject(0).getObject("itemSectionRenderer").getArray("contents")
+                    .getObject(0).getObject("messageRenderer").getObject("text").getArray("runs")
+                    .getObject(0).getString("text").equals("This channel has no videos."))
+                return null;
+        } catch (Exception ignored) {}
 
         return videoTab;
     }
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
@@ -53,6 +53,7 @@ public class YoutubeChannelExtractor extends ChannelExtractor {
     /*package-private*/ static final String CHANNEL_URL_BASE = "https://www.youtube.com/channel/";
 
     private JsonObject initialData;
+    private JsonObject videoTab;
 
     public YoutubeChannelExtractor(StreamingService service, ListLinkHandler linkHandler) {
         super(service, linkHandler);
@@ -276,6 +277,8 @@ public class YoutubeChannelExtractor extends ChannelExtractor {
     }
 
     private JsonObject getVideoTab() throws ParsingException {
+        if (this.videoTab != null) return this.videoTab;
+
         JsonArray tabs = initialData.getObject("contents").getObject("twoColumnBrowseResultsRenderer")
                 .getArray("tabs");
         JsonObject videoTab = null;
@@ -301,6 +304,7 @@ public class YoutubeChannelExtractor extends ChannelExtractor {
                 return null;
         } catch (Exception ignored) {}
 
+        this.videoTab = videoTab;
         return videoTab;
     }
 }
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
@@ -177,8 +177,12 @@ public class YoutubeChannelExtractor extends ChannelExtractor {
                 throw new ParsingException("Could not get subscriber count", e);
             }
         } else {
-            // If the element is null, the channel have the subscriber count disabled
-            return -1;
+            // If there's no subscribe button, the channel has the subscriber count disabled
+            if (initialData.getObject("header").getObject("c4TabbedHeaderRenderer").getObject("subscribeButton") == null) {
+                return -1;
+            } else {
+                return 0;
+            }
         }
     }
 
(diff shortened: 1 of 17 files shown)


wb9688:
Apparently YouTube videos have either no age restriction or are 18+. This means we could probably just assume a video is 18+ if there are no next streams.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java
```diff
@@ -162,55 +171,83 @@ public static JsonObject getInitialData(String html) throws ParsingException {
...
-            } else if (s.getString("service").equals("ECATCHER")) {
-                // fallback to get a shortened client version which does not contain the last do digits
-                JsonArray params = s.getArray("params");
-                for (Object param: params) {
-                    JsonObject p = (JsonObject) param;
-                    String key = p.getString("key");
-                    if (key != null && key.equals("client.version")) {
-                        shortClientVersion = p.getString("value");
+                } else if (s.getString("service").equals("ECATCHER")) {
+                    // fallback to get a shortened client version which does not contain the last do digits
```
B0pol: Did you mean two instead of do?

wb9688: [Blame](https://github.com/TeamNewPipe/NewPipeExtractor/blame/dev/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java#L188) @TobiGr ;)

B0pol:
Can you have a look at #263?

wb9688:
@B0pol: I fixed parsing the links.

gzsombor:
It seems, that YoutubeStreamInfoItemExtractor.getUploaderUrl returns incorrect URLs, like https://www.youtube.com/UCeoWpzqutlvnNKgIX0RialQ

wb9688:
@gzsomber: Could confirm. I'll fix that tomorrow as well. Thanks for reporting, as NewPipe doesn't use that anywhere, and our test suite for NewPipeExtractor is really not comprehensive enough, e.g. it doesn't test YoutubeStreamInfoItemExtractor at all.

gzsombor:
Thanks, really appreciated!

wb9688:
@gzsombor: No problem. I've fixed that now and the only bugs now are TeamNewPipe/NewPipe#3144 and video titles with @ (and probably # as well) in them (e.g. https://www.youtube.com/watch?v=srnxL6KX4n8).

PeterHindes:
Nice, I just threw a build together with this, works like a charm so far. Fixed my issue :)

PeterHindes:
Why does the Travis CI build fail? Does it need to be updated for the new YouTube updates?

wb9688:
@PeterHindes: Travis fails because some sites sometimes don't like servers fetching them. Additionally, there's a PeerTube test that fails, because it uses a video that doesn't exist anymore.

B0pol:
I had an error with debug build & latest commit only:

## Exception
* __Request:__ https://www.youtube.com/channel/UCAvCL8hyXjSUHKEGuUPr1BA
* __Content Language:__ FR
* __GMT Time:__ 2020-02-27 16:33
* __Package:__ org.schabi.newpipe.debug
* __Version:__ 0.18.5
* __Service:__ YouTube
* __Operating System:__ Linux Android 8.1.0 - 27


```
java.lang.NullPointerException: Attempt to invoke virtual method 'com.grack.nanojson.JsonArray com.grack.nanojson.JsonObject.getArray(java.lang.String)' on a null object reference
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeChannelExtractor.getInitialPage(YoutubeChannelExtractor.java:206)
	at org.schabi.newpipe.extractor.utils.ExtractorHelper.getItemsPageOrLogError(ExtractorHelper.java:19)
	at org.schabi.newpipe.extractor.channel.ChannelInfo.getInfo(ChannelInfo.java:82)
	at org.schabi.newpipe.extractor.channel.ChannelInfo.getInfo(ChannelInfo.java:47)
	at org.schabi.newpipe.util.-$$Lambda$ExtractorHelper$u5W7VszTe8AoEexIsFM9huQfbkM.call(Unknown Source:4)

```
</p></details>
<hr>

Also, no profile picture is shown on this video
https://www.youtube.com/watch?v=Pkh8UtuejGw , it just show default thumbnail, but if you go on channel page, there is a profile picture. On v0.18.5 it shows a random picture (the one from the suggested stream as you found)

wb9688:
@B0pol: Yeah, I kinda expected my latest commit to have some bugs. I'll fix them though.

gzsombor:
I've fixed one little thing, about the incorrect thumbnail URLs, please add it to your fixes - https://github.com/wb9688/NewPipeExtractor/pull/1

opusforlife2:
I think this should be merged to be released as hotfix 0.18.6 to restore functionality taken away by the website change. @TobiGr @Stypox

TobiGr:
@opusforlife2 You can see in the PR description that @wb9688's todo list for this PR is not yet completed. As soon @wb9688 wants to merge, I'll do so and release a new version.

wb9688:
@opusforlife2: If we do that, I want to resolve the last 2 points first. Plus for that new release we should imho also merge TeamNewPipe/NewPipe#3154.

opusforlife2:
Sure. All I meant was that this shouldn't wait till the issues in the milestone for 0.19.0 are done as well.

wb9688:
@TobiGr: I'm in fact not going to change the `YoutubeCommentsExtractor` in this PR, as when I want to do it properly, I would have to refactor the whole comment stuff, so that you'd call `getCommentsExtractor()` in a `StreamExtractor` instead of a `StreamingService`, so that a `CommentsExtractor` doesn't have to fetch the video page again. I now consider this PR finished.

Stypox:
@wb9688 could you provide a debug apk?

wb9688:
@Stypox: Sure… [NewPipe Debug_org.schabi.newpipe.debug.apk.zip](https://github.com/TeamNewPipe/NewPipeExtractor/files/4267585/NewPipe.Debug_org.schabi.newpipe.debug.apk.zip)

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
```diff
@@ -76,46 +79,46 @@ public String getUrl() throws ParsingException {
...
+                for (Object thumbnailOverlay : videoInfo.getArray("thumbnailOverlays")) {
+                    if (((JsonObject) thumbnailOverlay).getObject("thumbnailOverlayTimeStatusRenderer") != null) {
+                        duration = getTextFromObject(((JsonObject) thumbnailOverlay)
+                                .getObject("thumbnailOverlayTimeStatusRenderer").getObject("text"));
+                    }
+                }
+            } catch (Exception ignored) {}
         }
+        if (duration != null) return YoutubeParsingHelper.parseDurationString(duration);
+        throw new ParsingException("Could not get duration");
```
Stypox: This removes one check when the duration extractor was successful with first try
```suggestion
                }
            } catch (Exception ignored) {}
            
            if (duration == null) {
                throw new ParsingException("Could not get duration");
            }
        }
        return YoutubeParsingHelper.parseDurationString(duration);
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
```diff
@@ -125,30 +128,27 @@ public String getUploaderName() throws ParsingException {
     @Override
     public String getUploaderUrl() throws ParsingException {
```
Stypox: In order to remove some checks and make more clear that there are many methods to extract the url which are tried one after the other, I'd suggest doing something like this:

```java
// try method one

if (url == null) {
    // try method two

    if (url == null) {
        // try method three

        if (url == null || url.isEmpty()) {
            // throw exception
        }
    }
}
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java
```diff
@@ -162,56 +177,163 @@ public static JsonObject getInitialData(String html) throws ParsingException {
...
+            Map<String, List<String>> headers = new HashMap<>();
+            headers.put("X-YouTube-Client-Name", Collections.singletonList("1"));
+            headers.put("X-YouTube-Client-Version",
+                    Collections.singletonList(HARDCODED_CLIENT_VERSION));
+            final String response = getDownloader().get(url, headers).responseBody();
+            if (response.length() > 50) { // ensure to have a valid response
+                clientVersion = HARDCODED_CLIENT_VERSION;
+                return clientVersion;
+            }
+        } catch (Exception ignored) {}
```
TobiGr: Please do this in a separate method and add a test to check if the hard-coded client version is valid.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -116,7 +113,7 @@ public String getName() throws ParsingException {
         assertPageFetched();
         String title = null;
         try {
-            title = getVideoPrimaryInfoRenderer().getObject("title").getArray("runs").getObject(0).getString("text");
```
Stypox: Same as above: since you often use `getVideoPrimaryInfoRenderer()`, it would be a good idea to save it in a variable

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
```diff
@@ -64,16 +60,35 @@ public YoutubeChannelExtractor(StreamingService service, ListLinkHandler linkHan
...
+        JsonArray ajaxJson;
+
+        Map<String, List<String>> headers = new HashMap<>();
+        headers.put("X-YouTube-Client-Name", Collections.singletonList("1"));
+        headers.put("X-YouTube-Client-Version",
+                Collections.singletonList(YoutubeParsingHelper.getClientVersion()));
+        final String response = getDownloader().get(url, headers, getExtractorLocalization()).responseBody();
+        if (response.length() < 50) { // ensure to have a valid response
+            throw new ParsingException("Could not parse json data for next streams");
+        }
```
TobiGr: We might want to append response to the exception message. This could help us with debugging error reports later.

Apart from that, we could also add a 
`public static boolean isJsonResponseValid(String response)` to YouTubeParsingHelper.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -261,17 +212,14 @@ public Description getDescription() throws ParsingException {
     }
 
     @Override
-    public int getAgeLimit() throws ParsingException {
-        assertPageFetched();
-        if (!isAgeRestricted) {
+    public int getAgeLimit() {
+        if (initialData == null || initialData.isEmpty()) throw new IllegalStateException("initialData is not parsed yet");
```
Stypox: Why doesn't `assertPageFetched()` work anymore?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -381,17 +322,15 @@ public long getDislikeCount() throws ParsingException {
...
+            String uploaderId = playerResponse.getObject("videoDetails").getString("channelId");
+            if (uploaderId != null) return "https://www.youtube.com/channel/" + uploaderId;
         } catch (Exception ignored) {}
-        if (uploaderId == null) {
-            try {
-                uploaderId = playerResponse.getObject("videoDetails").getString("channelId");
-            } catch (Exception ignored) {}
-        }
-        if (uploaderId != null) return "https://www.youtube.com/channel/" + uploaderId;
         throw new ParsingException("Could not get uploader url");
```
Stypox: Same as above to save `if` checks and improve code readability, I'd move `if (uploaderId == null) throw new ParsingException("Could not get uploader url");` at the end of the `if` just above

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -261,17 +212,14 @@ public Description getDescription() throws ParsingException {
     }
 
     @Override
-    public int getAgeLimit() throws ParsingException {
-        assertPageFetched();
-        if (!isAgeRestricted) {
+    public int getAgeLimit() {
```
Stypox: The result of this function should, too, being cached, since it is used many times

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -578,7 +504,7 @@ public StreamType getStreamType() throws ParsingException {
     @Override
     public StreamInfoItem getNextStream() throws ExtractionException {
         assertPageFetched();
-        if (isAgeRestricted) {
+        if (getAgeLimit() == 18) {
```
Stypox:
```suggestion
        if (getAgeLimit() != NO_AGE_LIMIT) {
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -599,7 +525,7 @@ public StreamInfoItem getNextStream() throws ExtractionException {
     @Override
     public StreamInfoItemsCollector getRelatedStreams() throws ExtractionException {
         assertPageFetched();
-        if (isAgeRestricted) {
+        if (getAgeLimit() == 18) {
```
Stypox:
```suggestion
        if (getAgeLimit() != NO_AGE_LIMIT) {
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -869,7 +773,7 @@ private String getDecryptionFuncName(String playerCode) throws DecryptException
     @Nonnull
     private List<SubtitlesInfo> getAvailableSubtitlesInfo() {
         // If the video is age restricted getPlayerConfig will fail
-        if (isAgeRestricted) return Collections.emptyList();
+        if (getAgeLimit() == 18) return Collections.emptyList();
```
Stypox:
```suggestion
        if (getAgeLimit() != NO_AGE_LIMIT) return Collections.emptyList();
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubePlaylistExtractor.java
```diff
@@ -127,8 +149,7 @@ public String getBannerUrl() {
     @Override
     public String getUploaderUrl() throws ParsingException {
         try {
-            return YoutubeChannelExtractor.CHANNEL_URL_BASE +
-                    uploaderInfo.getObject("navigationEndpoint").getObject("browseEndpoint").getString("browseId");
+            return getUrlFromNavigationEndpoint(getUploaderInfo().getObject("navigationEndpoint"));
```
Stypox: Isn't there a way to always get the uploader url in the "/channel/ID" form? This would make it consistent with other urls in the app, preventing bugs such as duplicate subscriptions

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
wb9688: >We might want to append response to the exception message. This could help us with debugging error reports later.

I'll do that. The error message doesn't make sense (probably copied from somewhere), so I'll improve that as well.

>Apart from that, we could also add a
`public static boolean isJsonResponseValid(String response)` to YouTubeParsingHelper.

Is that really needed for just `response.length() < 50`?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
wb9688: Because getAgeLimit is (now indirectly) being called in `onFetchPage()`.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
wb9688: That's better indeed, though `getAgeLimit()` will only ever return `18`.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubePlaylistExtractor.java
wb9688: Yes, there is.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
wb9688: As in the following? Or what do you mean? I'm not sure if I think that's cleaner. @TobiGr: What do you think?

```
try {
    // Do whatever
} catch (Exception ignored) {}

if (url == null) {
    try {
        // Do whatever
    } catch (Exception ignored) {}

    if (url == null) {
        try {
            // Do whatever
        } catch (Exception ignored) {}

        if (url == null || url.isEmpty()) {
           throw new ParsingException("whatever");
        }
    }
}
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubePlaylistExtractor.java
```diff
@@ -41,11 +40,26 @@ public YoutubePlaylistExtractor(StreamingService service, ListLinkHandler linkHa
...
+            throw new ParsingException("Could not parse json data for next streams");
+        }
+
+        try {
+            ajaxJson = JsonParser.array().from(response);
+        } catch (JsonParserException e) {
+            throw new ParsingException("Could not parse json data for next streams", e);
+        }
+
+        initialData = ajaxJson.getObject(1).getObject("response");
```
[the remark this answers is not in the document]
wb9688: Yes, that was actually my plan, but I forgot about it. Thanks for reminding me.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
wb9688: True, this probably happens at a lot of other places as well.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
TobiGr: It is ok for now, but maybe we can validate a response in the future differently (and more accurate). We can leave it for now.

wb9688: YouTube returns `{"reload":"now"}` iirc if the client version is incorrect. I think this check isn't needed anymore now that we check whether the client version is valid and stuff in `getClientVersion()`.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
TobiGr: Well, the code above has less `url == null` checks when the first extraction method returns a valid result and is therefore "faster". On the other hand when the url is empty, we might not throw an Exception, but return an empty string. So we'd need something like this. 

```java
try {
    // Do whatever
} catch (Exception ignored) {}

if (url == null) {
    try {
        // Do whatever
    } catch (Exception ignored) {}

    if (url == null) {
        try {
            // Do whatever
        } catch (Exception ignored) {}
    }
}
if (url == null || url.isEmpty()) {
    throw new ParsingException("whatever");
}
return url;
```
However, I do not understand, why we only check once whether the url is empty. We should do this either every time or not at all. I suggest to remove this isEmpty check, because the JSON fields won't be empty. They either exist and contain a valid value or do not exist. That's why I'd use the following, which might be more efficient when having a valid result at the beginning. 

```java
try {
    // Do whatever
} catch (Exception ignored) {}

if (url == null) {
    try {
        // Do whatever
    } catch (Exception ignored) {}

    if (url == null) {
        try {
            // Do whatever
        } catch (Exception ignored) {}

        if (url == null) {
            throw new ParsingException("whatever");
        }
    }
}

return url;
```
However, these kind of changes won't speed up anything in a way a user can recognize that. If you want, you can change it, but maybe we should ask @Stypox first, if we understood correctly what he meant.

B0pol:
I found another bug:
## Exception
* __User Action:__ subscription
* __Request:__ https://www.youtube.com/channel/UCNiauGTV7XhkOpUAIXod4xA
* __Content Language:__ GB
* __Service:__ YouTube
* __Version:__ 0.18.5


```
org.schabi.newpipe.extractor.exceptions.ParsingException: Could not get client version
	at org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeParsingHelper.getClientVersion(YoutubeParsingHelper.java:265)
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeChannelExtractor.onFetchPage(YoutubeChannelExtractor.java:71)
	at org.schabi.newpipe.extractor.Extractor.fetchPage(Extractor.java:56)
	at org.schabi.newpipe.extractor.channel.ChannelInfo.getInfo(ChannelInfo.java:46)
	at org.schabi.newpipe.util.-$$Lambda$ExtractorHelper$u5W7VszTe8AoEexIsFM9huQfbkM.call(Unknown Source:4)
	at io.reactivex.internal.operators.maybe.MaybeSubscribeOn$SubscribeTask.run(MaybeSubscribeOn.java:54)
	at io.reactivex.internal.schedulers.ScheduledDirectTask.call(ScheduledDirectTask.java:38)
	at io.reactivex.internal.schedulers.ScheduledDirectTask.call(ScheduledDirectTask.java:26)

```
</p></details>
<hr>

Steps to reproduce: disable WI-FI and mobile data. If the error doesn't appear, close and relaunch the app.

What we expect: Network error, see that screenshot from official 0.18.5

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
Stypox: In terms of speed up this does not actually do much, but I think the second snippet provided by @TobiGr is more readable, since the indentation makes it clear that if the first method succeeds, the function will return that result immediately. With the current implementation you have to read and process more lines in order to understand that if the first method succeeds nothing else happens (i.e. there is no post processing or other things) @TobiGr @wb9688

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -381,17 +321,15 @@ public long getDislikeCount() throws ParsingException {
...
         try {
-            uploaderId = getVideoSecondaryInfoRenderer().getObject("owner").getObject("videoOwnerRenderer")
-                    .getObject("navigationEndpoint").getObject("browseEndpoint").getString("browseId");
+            String uploaderUrl = getUrlFromNavigationEndpoint(getVideoSecondaryInfoRenderer()
+                    .getObject("owner").getObject("videoOwnerRenderer").getObject("navigationEndpoint"));
+            if (uploaderUrl != null) return uploaderUrl;
+        } catch (Exception ignored) {}
+        try {
+            String uploaderId = playerResponse.getObject("videoDetails").getString("channelId");
+            if (uploaderId != null) return "https://www.youtube.com/channel/" + uploaderId;
```
TobiGr: This should be done using the 
[`YoutubeChannelLinkHandlerFactory.getUrl(iploaderId, null, null)`](https://github.com/TeamNewPipe/NewPipeExtractor/blob/fc465c8bf7b70d9052fc5bf1f524f1c97b330718/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeChannelLinkHandlerFactory.java#L39-L41) method. But it seems to be buggy to me. It does not return "https://www.youtube.com/channel/ID", but "https://www.youtube.com/ID". I searched for a small random channel (www.youtube.com/user/effectpl) and looked up its channelId (`UCuyC5c8JXzPx71aMg7rGFgQ`). While www.youtube.com/UCuyC5c8JXzPx71aMg7rGFgQ results in a 404 response,  www.youtube.com/channel/UCuyC5c8JXzPx71aMg7rGFgQ is valid.
Does anyone know, why it is implemented that way?

wb9688: That's something I noticed as well. It appears that someone wants to have `channel/UCwhatever` as input there, so that it would also accept e.g. `user/whatever` URLs without having to do a conversion of some kind.

TobiGr: Ah yes, correct. That was the reason. Can you please add this info as Jdoc and use the method?

opusforlife2:
What _is_ the pbj parameter, though? I can't find information on it anywhere.

wb9688:
@opusforlife2: YouTube is a SPA (Single Page Application). This means that it loads all the code and stuff when you go to youtube.com in your browser. Once you're on YouTube and click on e.g. a video, it only loads the neccessary (JSON) data to show that video. So that YouTube's server knows it should only serve the neccessary data, YouTube adds the pbj=1 parameter to the URL. Since we only need that data and not all the code and stuff, we now directly extract the data from the pbj=1 version, which thus also saves data and makes our parsing a bit more reliable.

opusforlife2:
Awesome! I just couldn't stop thinking of it as peanut butter and jam, so thanks for clearing that up. :P

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubePlaylistExtractor.java
wb9688: @TobiGr: I don't think we should do that, as there are many other ways to get to a /user/name channel in NewPipe. Plus NewPipe could do a simple call to `getId()` or `getUrl()` in `YoutubeChannelExtractor` to get a /channel/id URL, if NewPipe isn't doing that already.

wb9688:
@TobiGr: This PR should be ready to merge now I think.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/search/YoutubeSearchExtractorChannelOnlyTest.java
```diff
@@ -77,7 +92,7 @@ public void testChannelUrl() {
             if (item instanceof ChannelInfoItem) {
                 ChannelInfoItem channel = (ChannelInfoItem) item;
 
-                if (channel.getSubscriberCount() > 5e7) { // the real PewDiePie
+                if (channel.getSubscriberCount() > 5e7 && channel.getName().equals("PewDiePie")) { // the real PewDiePie
```
B0pol: Is that a good idea? Because we are talking about channel name, something youtubers (can) change often & easily. Check for id would be better (unchangeable)

wb9688: You're right.

B0pol: Then change 5e7 (50M) to 1e8, ie 100M

wb9688: The problem is T-Series showing up in the results, so that won't help. What would help is a `break` after the first channel is found that matches the subscriber criteria.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
```diff
@@ -76,89 +79,95 @@ public String getUrl() throws ParsingException {
...
-                name = videoInfo.getObject("title").getArray("runs").getObject(0).getString("text");
-            } catch (Exception ignored) {}
-        }
+        String name = getTextFromObject(videoInfo.getObject("title"));
         if (name != null && !name.isEmpty()) return name;
         throw new ParsingException("Could not get name");
     }
 
     @Override
     public long getDuration() throws ParsingException {
```
mauriciocolli: Doesn't handle video premieres.

PR #265, opened 2020-02-27
Test if services recognizes their own items urls

mauriciocolli:
There's still some work to do on the tests, let's begin improving it.

- Test if services recognizes their own items urls
- Improve kiosk tests across services


mauriciocolli:
@TobiGr This is just a change in the tests, should be pretty straightforward to merge.

Can't do it because a review is required.

extractor/src/test/java/org/schabi/newpipe/extractor/services/DefaultTests.java
```diff
@@ -37,34 +51,56 @@ public static void defaultTestListOfItems(int expectedServiceId, List<? extends
                     assertTrue("Upload date not in the past", uploadDate.date().before(Calendar.getInstance()));
                 }
 
+            } else if (item instanceof ChannelInfoItem) {
+                final ChannelInfoItem channelInfoItem = (ChannelInfoItem) item;
+                assertExpectedLinkType(expectedService, channelInfoItem.getUrl(), LinkType.CHANNEL);
+
+            } else if (item instanceof PlaylistInfoItem) {
+                final PlaylistInfoItem playlistInfoItem = (PlaylistInfoItem) item;
+                assertExpectedLinkType(expectedService, playlistInfoItem.getUrl(), LinkType.PLAYLIST);
```
Stypox: Shouldn'y the thumbnail url be tested here, too? I'd suggest creating a function in test utils that does this, so that it can be re-used anywhere:
```java
void assertThumbnailUrlOk(@Nullable String thumbnailUrl) {
    if (thumbnailUrl != null && !thumbnailUrl.isEmpty()) {
        assertIsSecureUrl(thumbnailUrl);
    }
}
```

mauriciocolli: They are already being tested, look at [this check above](https://github.com/TeamNewPipe/NewPipeExtractor/pull/265/files/f51da738370954d5e925adaa7192bdf18568fd61#diff-249e17d5a4640bc1be5171fc1853fddbR31-R33).

Issue #266, opened 2020-02-28
Improve tests


wb9688:
While implementing `yt_new`, I found out a lot of things just aren't tested at all, for example the InfoItemExtractors. I'll update this issue with more and concrete examples later.

PeterHindes:
Yes, I noticed a lot of regressions in the new versions. It would be nice to have tests for all the possible null values. For example properly handling unlikely cases like no upload date.

PR #267, opened 2020-03-01
Hotfixes for YouTube and improve unavailable cases

mauriciocolli:
Fix the issues that I mentioned in #262, and some more that I found along the way.

---

### YouTube

- Take into account videos that have their views hidden.
- Handle videos with no views or with "Recommended to you" text.
- Handle video premiere's date and duration.
- Avoid crashing by letting exceptions bubble up.
  - reCAPTCHAS, for example, were not reaching where they should, resulting in a crash instead of opening it for the user to solve.
  - Maybe there's a better way/approach to how the exceptions are handled in the extractor (e.g. communicate error to front end, error propagation), I'll take a look some time.
- Fix bug when url isn't present in the browseEndpoint object.
- Detect simple 404s in the standard fetch method.
- Detect deleted/nonexistent/invalid channels and playlists.
- Detect when a stream is deleted or doesn't exist.

### Downloader

- Make test downloader return a response instead of throwing an exception.
- Add latest url to the response to make detection of a redirect possible.
- Ignore null-keyed entries when iterating through the response headers.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
@@ -203,6 +203,12 @@ public class YoutubeStreamInfoItemExtractor implements StreamInfoItemExtractor {
 
             final String viewCount = getTextFromObject(viewCountObject);
 
+            if (viewCount.toLowerCase().contains("no views")) {
+                return 0;
+            } else if (viewCount.toLowerCase().contains("recommended")) {
+                return -1;
+            }
+
             return Long.parseLong(Utils.removeNonDigitCharacters(viewCount));
         } catch (Exception e) {
             throw new ParsingException("Could not get view count", e);
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
@@ -2,7 +2,6 @@ package org.schabi.newpipe.extractor.services.youtube.extractors;
 
 import com.grack.nanojson.JsonArray;
 import com.grack.nanojson.JsonObject;
-
 import org.schabi.newpipe.extractor.exceptions.ParsingException;
 import org.schabi.newpipe.extractor.localization.DateWrapper;
 import org.schabi.newpipe.extractor.localization.TimeAgoParser;
@@ -13,10 +12,11 @@ import org.schabi.newpipe.extractor.stream.StreamType;
 import org.schabi.newpipe.extractor.utils.Utils;
 
 import javax.annotation.Nullable;
+import java.text.SimpleDateFormat;
+import java.util.Calendar;
+import java.util.Date;
 
-import static org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeParsingHelper.fixThumbnailUrl;
-import static org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeParsingHelper.getTextFromObject;
-import static org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeParsingHelper.getUrlFromNavigationEndpoint;
+import static org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeParsingHelper.*;
 
 /*
  * Copyright (C) Christian Schabesberger 2016 <chris.schabesberger@mailbox.org>
@@ -86,7 +86,9 @@ public class YoutubeStreamInfoItemExtractor implements StreamInfoItemExtractor {
 
     @Override
     public long getDuration() throws ParsingException {
-        if (getStreamType() == StreamType.LIVE_STREAM) return -1;
+        if (getStreamType() == StreamType.LIVE_STREAM || isPremiere()) {
+            return -1;
+        }
 
         String duration = null;
 
@@ -165,7 +167,16 @@ public class YoutubeStreamInfoItemExtractor implements StreamInfoItemExtractor {
 
     @Nullable
     @Override
-    public String getTextualUploadDate() {
+    public String getTextualUploadDate() throws ParsingException {
+        if (getStreamType().equals(StreamType.LIVE_STREAM)) {
+            return null;
+        }
+
+        if (isPremiere()) {
+            final Date date = getDateFromPremiere().getTime();
+            return new SimpleDateFormat("yyyy-MM-dd HH:mm").format(date);
+        }
+
         try {
             return getTextFromObject(videoInfo.getObject("publishedTimeText"));
         } catch (Exception e) {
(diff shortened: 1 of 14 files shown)


wb9688:
Thanks a lot for your PR! I wanted to implement the things you wrote about, but now that's not needed anymore. I'll take a closer look in a few hours.

wb9688:
I just found another bug, which could be fixed by looping over all badges in getStreamType() in YoutubeStreamInfoItemExtractor. Could you please do that?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
```diff
@@ -225,7 +255,26 @@ private boolean isPremium() {
...
                 }
             }
-        } catch (Exception ignored) {}
+        } catch (Exception ignored) {
+        }
         return false;
     }
+
+    private boolean isPremiere() {
+        return videoInfo.has("upcomingEventData");
```
wb9688: Are you sure this works? I had issues with using `has()` for JsonObjects, so I've been doing `getObject() != null`. Otherwise this PR looks fine to me.

TobiGr: What issues did you have with `has(String key)`? I've looked at the source code of that method and it returns `hashmap.contains(key);` So there should not be issues with that.
Btw. we use an outdated version of nanojson. We might want to update it as the new version promises to be faster.

wb9688: It returning `false` when that object exists, but from a quick local test, that doesn't seem to happen, strange…

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
```diff
@@ -194,11 +213,22 @@ public long getViewCount() throws ParsingException {
...
+            final JsonObject viewCountObject = videoInfo.getObject("viewCountText");
+            if (viewCountObject == null) {
+                // This object is null when a video has its views hidden.
+                return -1;
+            }
+
+            final String viewCount = getTextFromObject(viewCountObject);
+
+            if (viewCount.toLowerCase().contains("no views")) {
+                return 0;
```
B0pol: Are there no other ways than relying on English?

TobiGr:
>I just found another bug, which could be fixed by looping over all badges in getStreamType() in YoutubeStreamInfoItemExtractor. Could you please do that?

@wb9688 Do you mean something like

``` Java
JsonArray badges = videoInfo.getArray("badges");
for (Object badge : badges) {
    if (((JsonObject) badge).getObject("metadataBadgeRenderer").getString("label").equals("LIVE NOW")) {
        return StreamType.LIVE_STREAM;
    }
}
```

wb9688:
@TobiGr: That's exactly what I meant.

TobiGr:
@wb9688 I could not find any StreamInfoItem which has two badges while running the tests from #266. Do you have a link to a document which contains such a case?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
TobiGr: At the moment no.

B0pol:
> ~@wb9688 I could not find any StreamInfoItem which has two badges while running the tests from #266. Do you have a link to a document which contains such a case?~ I found one.

can you give me a link please?

TobiGr:
@B0pol It is a [search querry for `cirque du soleil`](https://www.youtube.com/results?search_query=cirque+du+soleil&sp=EgIQAQ%253D%253D&gl=GB)

mauriciocolli:
One more issue though: when opening a premiere video, it's not throwing the right exception.

We should add more status checks [here](https://github.com/mauriciocolli/NewPipeExtractor/blob/5686a6f562619c50bed53904eaafbc2aa85c8dbe/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java#L625).

mauriciocolli:
Also, @TobiGr what about we disable these protected branches shenanigans?

In a lot of pull requests I see, it's infested with "merge branch dev into \<branch\>", because people keep pressing the update button.

For example, this one and #265 should be perfectly able to perform a simple merge. Then you don't need to overwrite the commits and rewrite history (I like seeing those verified commits).

TobiGr:
@mauriciocolli 
 Should I add request.url() to https://github.com/TeamNewPipe/NewPipe/blob/b1eaf5616ad25abf2907d3283abae04334cbcdb2/app/src/main/java/org/schabi/newpipe/DownloaderImpl.java#L174? I think that is what you intended to have there. Is that correct?

TobiGr:
>In a lot of pull requests I see, it's infested with "merge branch dev into <branch>", because people keep pressing the update button.

Yes, I hate that, too. I'd like people to rebase their commits to a have a readable history. We need to find a solution for this.

wb9688:
@TobiGr: Does GitHub not have a rebase and merge button?

mauriciocolli:
@TobiGr no, it should be latest url that response was produced to. In case of a redirection, the response would be made to the redirected url.

Here's a diff of what should be done:

```diff
diff --git a/app/src/main/java/org/schabi/newpipe/DownloaderImpl.java b/app/src/main/java/org/schabi/newpipe/DownloaderImpl.java
index 0ae072c66..6591fcbda 100644
--- a/app/src/main/java/org/schabi/newpipe/DownloaderImpl.java
+++ b/app/src/main/java/org/schabi/newpipe/DownloaderImpl.java
@@ -171,7 +171,9 @@ public class DownloaderImpl extends Downloader {
             responseBodyToReturn = body.string();
         }
 
-        return new Response(response.code(), response.message(), response.headers().toMultimap(), responseBodyToReturn);
+        final String latestUrl = response.request().url().toString();
+
+        return new Response(response.code(), response.message(), response.headers().toMultimap(), responseBodyToReturn, latestUrl);
     }
 
     /**
```

The `Response#request` returns the request that it originated from, then we can get the url.

Also, there's the other issue that I mentioned, so please wait a little bit longer.

> Yes, I hate that, too. I'd like people to rebase their commits to a have a readable history. We need to find a solution for this.

I think this is about the protected branch:

protected_branch

TobiGr:
> Here's a diff of what should be done:

Thanks. I'll change this when updating the extractor version.

> One more issue though: when opening a premiere video, it's not throwing the right exception.

Ok. Are you on it, or should I take a look at it?

> I think if we disable that, we can merge requests without that merge hell. Imagine 10 pull requests and having to update each one. I think the update button would work, but then even more merge commits would be produced.

ok.

mauriciocolli:
> @wb9688: I just found another bug, which could be fixed by looping over all badges in getStreamType() in YoutubeStreamInfoItemExtractor. Could you please do that?

TobiGr pushed this https://github.com/TeamNewPipe/NewPipeExtractor/commit/35252235b0f88fcc44eb048be0228f94bcc36360.

>@TobiGr: Ok. Are you on it, or should I take a look at it?

I'm currently testing it., I'll open another pull request soon.

mauriciocolli:
Oh, and @TobiGr, sorry if it caused a misunderstanding, but _you_ would have to disable it, as I don't have permissions to change that setting (I don't even see here, screenshot from another repo).

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
B0pol: Well there is almost always a way.
If it has BADGE_STYLE_TYPE_SIMPLE and can't extract views, it means it has no views.
For recommend, I tested with russian french german chinese thai, it stays « recommended for you »

PR #269, opened 2020-03-01
Fix for Android versions earlier than KitKat

wb9688:

The StandardCharsets class was only introduced in KitKat, so this is only relevant for NewPipe Legacy.


wb9688:
@TobiGr: Yup… I don't think you need to open an issue there, as Android Studio would just say those classes don't exist.

PR #271, opened 2020-03-01
[YouTube] Improve response status checks

mauriciocolli:
### YouTube
- Fail-fast if status exist and is anything other than "OK"

### PeerTube
- Update video used in age limit test (the previous was 404)

---

Incredible, forgot how a green check mark for tests looked like (though some areas are lacking tests).

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
@@ -623,7 +623,8 @@ public class YoutubeStreamExtractor extends StreamExtractor {
 
         final JsonObject playabilityStatus = playerResponse.getObject("playabilityStatus", JsonUtils.DEFAULT_EMPTY);
         final String status = playabilityStatus.getString("status");
-        if (status != null && status.toLowerCase().equals("error")) {
+        // If status exist, and is not "OK", throw a ContentNotAvailableException with the reason.
+        if (status != null && !status.toLowerCase().equals("ok")) {
             final String reason = playabilityStatus.getString("reason");
             throw new ContentNotAvailableException("Got error: \"" + reason + "\"");
         }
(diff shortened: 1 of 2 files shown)


Stypox:
Would it be possible that YouTube changes the "ok" string to something else (e.g. "success"), so that the extractor fails even though the content is there?

wb9688:
"some areas"? A lot of areas ;) (reminds me that I still need to update #266…)

And yes, that green check mark is really nice.

mauriciocolli:
> Would it be possible that YouTube changes the "ok" string to something else (e.g. "success"), so that the extractor fails even though the content is there?

Yes, but I didn't encounter this response and I tested a lot of videos.

It's easier to make an exception for "OK", than include all the others. But maybe it's worth it?

---

@TobiGr @wb9688 @Stypox 

What about we upload some test videos like `youtube-dl` does, testing various parts? I think we could do some pretty reliable tests that way.

For example, tests for features, like premieres, could be done (checked and it can be set all the way to 2022, we could keep testing until then).

wb9688:
@mauriciocolli: I think that's a good idea. However, we couldn't test everything that way, as some features aren't available to small channels and it's hard to test e.g. date parsing of a premiere that happened less than 24 hours ago.

Stypox:
> Yes, but I didn't encounter this response and I tested a lot of videos.

Ok, then it's ok, and if youtube ever changes the "ok" string it would not take much to find the problem.

> What about we upload some test videos like youtube-dl does, testing various parts?

Yeah, I thought about doing that, too. It is a good idea imo. I could do that if needed, I have some "experience" with YouTube (i.e. I used to run a stupid minecraft channel some years back). But should we publish the login data to a place everybody in TeamNewPipe can access, or what?

Stypox:
@wb9688 I have a channel with more that 4000 subscribers if needed (it is completely useless now, so if needed we could take advantage of it)

TobiGr:
> What about we upload some test videos like youtube-dl does, testing various parts? I think we could do some pretty reliable tests that way.

Good idea, but please keep in mind that you agree to YouTube's TOS when having a channel.

PR #276, opened 2020-03-02
Don't accept YouTube Mix playlists

wb9688:

This will let NewPipe open the actual video from the YouTube Mix playlist you're watching, as NewPipeExtractor can't handle YouTube Mix playlists yet. This would vastly improve YouTube Music support for now, as when you watch a video there, it'll automatically go to the Mix.

Or we could wait until someone implemens properly extracting those Mix playlists.


B0pol:
There is a PR for youtube mix, #253 (but was not updated to new youtube website).
To me, handling mixes is better because the behavior part of youtube music totally relies on playlist (mix are playlist), and here you break that playlist for a single video.

Why spotify (as example) has so many users: first, unlimited listening to whatever music you want, but also and that's an important thing, you listen to playlist (or "radios" / daily mixes) continuously
And Youtube music was built like that, this idea of continuous music, with mixes that are endless.

wb9688:
@B0pol: I know, but this was intended as a temporary fix, until e.g. @XiangRongLin properly implements support for mixes. One downside of our current playlist implementation is that it doesn't store at what stream the user was though.

PeterHindes:
Interesting. Since Mixes are endless how do you que them all?

wb9688:
Closed in favor of #280.

PR #279, opened 2020-03-03
Add support for extracting multiple tabs from a channel

wb9688:

To do:

- [x] Add support for different tabs in a channel
- [x] Add support for YouTube's playlists tab (partially closes TeamNewPipe/NewPipe#2414)
- [x] Add support for SoundCloud's popular tracks tab
- [x] Add support for SoundCloud's albums tab
- [x] Add support for SoundCloud's playlists tab
- [x] Make the tests not fail (something with most channels not having a next page of playlists…)
- [x] Make it possible to have both streams and playlists in a tab
- [x] Add support for SoundCloud's reposts tab
- [x] Actually test it in NewPipe (e.g. I think I'm missing tabs stuff in ChannelInfo?)

Are there any more tabs I should add?


Stypox:
Maybe a tab with channel description and links

wb9688:
@Stypox: That's not possible, a 'tab' here has to be a list of streams, playlists or channels. Making a separate tab for the description and stuff would just be UI stuff in NewPipe.

Stypox:
You're correct, sorry ;-)

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
```diff
@@ -168,107 +158,41 @@ public String getDescription() throws ParsingException {
...
-
+    private JsonObject getTab(String tabName) throws ParsingException {
         JsonArray tabs = initialData.getObject("contents").getObject("twoColumnBrowseResultsRenderer")
                 .getArray("tabs");
         JsonObject videoTab = null;
 
         for (Object tab : tabs) {
             if (((JsonObject) tab).getObject("tabRenderer") != null) {
-                if (((JsonObject) tab).getObject("tabRenderer").getString("title").equals("Videos")) {
+                if (((JsonObject) tab).getObject("tabRenderer").getString("title").equals(tabName)) {
```
B0pol: Please use url in endpoint of tabrenderer instead of title, so that it won't break for other languages. See https://github.com/B0pol/NewPipeExtractor/commit/ffc22b798a2e82d5ffec1cc5b61ca214b5813a6a
You can use endsWith (maybe lowercase if needed) if you just wanna check, or split by "/" and pick last to gather it

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelPlaylistsExtractor.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    private JsonObject getTab(String tabName) {
+        JsonArray tabs = initialData.getObject("contents").getObject("twoColumnBrowseResultsRenderer")
+                .getArray("tabs");
+        JsonObject videoTab = null;
+
+        for (Object tab : tabs) {
+            if (((JsonObject) tab).getObject("tabRenderer") != null) {
+                if (((JsonObject) tab).getObject("tabRenderer").getString("title").equals(tabName)) {
```
B0pol: Same thing here

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
wb9688: Thanks! I didn't think of localization.

wb9688:
@TobiGr and @mauriciocolli: I think this PR is finished now, so I'd appreciate it if you review it (once you have time for that).

I have one point for discussion though. When should we add a tab? E.g. should we always add all tabs? In some cases it's only possible to know a tab is empty once we've requested the initial page.

Don't merge this until I've the NewPipe part ready though.

PeterHindes:
Hey, refer to [my issue](https://github.com/TeamNewPipe/NewPipe/issues/3173) requesting the adding of support for playlist tabs. Music channels often have an albums category tag that you can select for on their page. This will need to be supported in the ui and in the extractor.

PeterHindes:
https://drive.google.com/uc?id=1nI5EK6fYdIVEzm5J6nbGKbSo7rxa_KgZ

B0pol:
Soundcloud is fully broken, ParsirgException. Doesn't happen with v0.18.6

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelVideosExtractor.java
```diff
@@ -0,0 +1,118 @@
...
+import org.schabi.newpipe.extractor.linkhandler.ListLinkHandler;
+import org.schabi.newpipe.extractor.localization.TimeAgoParser;
+
+import java.io.IOException;
+
+import javax.annotation.Nonnull;
+
+import static org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeParsingHelper.getJsonResponse;
+
+public class YoutubeChannelVideosExtractor extends ChannelTabExtractor {
```
mauriciocolli: The id is being extracted using the default way (`ListLinkHandler`) which returns a string starting in `user/` or `channel/`. Was this intentional?

But I guess it doesn't matter that much anyway because, currently, a tab can only be created by having the `ChannelInfo` first, let's see how the sort order can be implemented later.

Also, I see that the name is being used as a [sort of an id](https://github.com/TeamNewPipe/NewPipe/blob/a1a9b54432ae922dad144c3319e37a42fc65ec5b/app/src/main/java/org/schabi/newpipe/fragments/detail/ChannelDetailFragment.java#L544-L567) right now.

extractor/src/main/java/org/schabi/newpipe/extractor/channel/ChannelTabInfo.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private transient ChannelTabExtractor channelTabExtractor;
+
+    public ChannelTabExtractor getChannelTabExtractor() {
+        return channelTabExtractor;
+    }
+
+    public void setChannelTabExtractor(ChannelTabExtractor channelTabExtractor) {
+        this.channelTabExtractor = channelTabExtractor;
+    }
```
mauriciocolli: I thought that `Info` objects were supposed to be plain java/data objects? Storing a extractor in it feels definitely wrong (not serializable as well).

Would like your opinion/suggestions on this.

wb9688: True, that's why the `channelTabExtractor` variable is `transient`. This is needed to be able to get more items. The `getMoreItems()` function sets the `channelTabExtractor` again if it's not there anymore. We're doing this in `CommentsInfo` as well.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelVideosExtractor.java
wb9688: No, I just couldn't think of any other way, as a `LinkHandler` isn't really needed, since it could only be get through the `ChannelInfo`/`ChannelExtractor`, which is intentional.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
```diff
@@ -168,107 +158,39 @@ public String getDescription() throws ParsingException {
...
-        // Unfortunately, we have to fetch the page even if we are only getting next streams,
-        // as they don't deliver enough information on their own (the channel name, for example).
-        fetchPage();
-
-        StreamInfoItemsCollector collector = new StreamInfoItemsCollector(getServiceId());
-        final JsonArray ajaxJson = getJsonResponse(pageUrl, getExtractorLocalization());
-
-        JsonObject sectionListContinuation = ajaxJson.getObject(1).getObject("response")
-                .getObject("continuationContents").getObject("gridContinuation");
+    public List<ChannelTabExtractor> getTabs() throws ParsingException {
```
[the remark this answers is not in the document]
wb9688: The feed will use the first tab (which is the reason for a27979794e9a0eb6cd2087d22dc8f0304073b3e9)  and we should document that somewhere.

wb9688:
>I think that's not a problem and adding that there is a tab is enough, let the client lazy-load it when needed, otherwise it may be too much for some services.

@mauriciocolli: How are we gonna lazy-load it? That's not possible with the current architecture, as all ChannelTabInfos are put into a List in the ChannelInfo.

Also in case of YouTube, I think we can't know what tabs are exactly available, unless we make a request to both /videos (Music, More from the artist and Uploads should all be tabs) and /playlists (Created playlists and Albums should be tabs). Currently this PR only support Uploads and Created playlists, but I'll leave other tabs and detection of which ones are available to a later PR.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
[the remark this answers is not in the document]
mauriciocolli: Is it good enough though, don't a27979794e9a0eb6cd2087d22dc8f0304073b3e9 just prove it that it isn't?

I think that by relying on the position, we just introduce an anti-pattern to the code.

extractor/src/main/java/org/schabi/newpipe/extractor/channel/ChannelTabInfo.java
mauriciocolli: > True, that's why the channelTabExtractor variable is transient

I noticed it, I was talking about the use of the extractor object at all in the data class.

> This is needed to be able to get more items.

Is it though? Aren't we able to just use the [nextPageUrl](url) to get the next page? We are able to that in `ChannelInfo` itself currently.

But now we don't even need to fetch the initial page because the continuation response already include all we need.

mauriciocolli:
> @wb9688: How are we gonna lazy-load it? That's not possible with the current architecture, as all ChannelTabInfos are put into a List in the ChannelInfo.

Maybe we should change that then?

I think we should be able to use the same concept of something like kiosks I guess, which can be freely instantiated and fetch more items on its own as well.

As for the lists, put a loaded tab when it doesn't need to be fetched again, and some kind of placeholder when it needs to (with its tab's id/url to instantiate later).

> Also in case of YouTube, I think we can't know what tabs are exactly available, unless we make a request to both /videos (Music, More from the artist and Uploads should all be tabs) and /playlists (Created playlists and Albums should be tabs).

Should the options really be separate tabs then? Seems like following YouTube's design would be better on this one. It'd also work well with lazy loading this way.

When the user changed an option, the front end would just reload the tab with the new parameters.

> Currently this PR only support Uploads and Created playlists, but I'll leave other tabs and detection of which ones are available to a later PR.

:+1:

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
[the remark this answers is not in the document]
wb9688: I agree that this isn't the best solution, but I think it's fine if we clearly document it, unless you have a better suggestion.

wb9688:
>I think we should be able to use the same concept of something like kiosks I guess, which can be freely instantiated and fetch more items on its own as well.

@mauriciocolli: I don't think so, since not all channels have the same tabs, so I think instantiating ChannelTabs from the Channels instead of the StreamingServices is the way to go. However, dynamic tabs and tabs we haven't implemented yet will be left to another PR.

###### I also think we should use the same approach to instantiate the comments from the streams, so we don't have to get the video page again in the comments, but could just pass the continuation and save a request (unless the user has switched to another app before requesting the comments.

I've been thinking about lazy-loading and I think we should add a function to ChannelTabInfo to populate fields other than the name and call that only once that tab is selected in NewPipe.

> Should the options really be separate tabs then? Seems like following YouTube's design would be better on this one. It'd also work well with lazy loading this way.

I don't think we should implement sub-tabs, as that's just unnecessary complexity imho.

I'll rebase on dev again and implement lazy-loading today, and maybe the better `LinkHandler` as well.

Stypox:
Why is the extractor based on Info containers, that collect all info at once? Wouldn't it be better to dynamically load everything? This would mean adding many ReactiveX queries in the client, one for every field extracted by the Extractor, but probably improve the way errors are handled, and would show progress to the user one piece at a time, thus giving the impression of faster loading and also enabling some options/buttons some time before (like the current behaviour with comments).

wb9688:
@Stypox: They're serializable and thus could be used to restore the state on Android.

Stypox:
That's something that should eventually be done on Android side (e.g. loading everything separately but still saving all in a Serializable variable)

wb9688:
@Stypox: Please create an issue to discuss that, since that's not relevant to this PR.

extractor/src/main/java/org/schabi/newpipe/extractor/channel/ChannelInfo.java
```diff
@@ -94,6 +81,17 @@ public static ChannelInfo getInfo(ChannelExtractor extractor) throws IOException
             info.addError(e);
         }
 
+        List<ChannelTabInfo> tabs = new ArrayList<>();
+        for (int i = 0; i < extractor.getTabs().size(); i++) {
+            try {
+                ChannelTabInfo tabInfo = ChannelTabInfo.getInfo(extractor.getTabs().get(i), i == 0);
```
Stypox: Is it hardcoded that the first tab will always be the feed one? Also, what happens if the first tab throws an error?

extractor/src/main/java/org/schabi/newpipe/extractor/channel/ChannelTabInfo.java
```diff
@@ -0,0 +1,95 @@
...
+
+    public static void assureChannelTabExtractor(ChannelTabInfo tabInfo) throws ExtractionException, IOException {
+        if (tabInfo.getChannelTabExtractor() == null) {
+            ChannelExtractor channelExtractor = NewPipe.getServiceByUrl(tabInfo.getUrl()).getChannelExtractor(tabInfo.getUrl());
+            channelExtractor.fetchPage();
+
+            List<ChannelTabExtractor> channelTabExtractors = channelExtractor.getTabs();
+            for (ChannelTabExtractor channelTabExtractor : channelTabExtractors) {
+                if (channelTabExtractor.getName().equals(tabInfo.getName())) {
+                    tabInfo.setChannelTabExtractor(channelTabExtractor);
```
Stypox: Then `break`

extractor/src/main/java/org/schabi/newpipe/extractor/channel/ChannelTabExtractor.java
```diff
@@ -0,0 +1,12 @@
+package org.schabi.newpipe.extractor.channel;
+
+import org.schabi.newpipe.extractor.InfoItem;
+import org.schabi.newpipe.extractor.ListExtractor;
+import org.schabi.newpipe.extractor.StreamingService;
+import org.schabi.newpipe.extractor.linkhandler.ListLinkHandler;
+
+public abstract class ChannelTabExtractor extends ListExtractor<InfoItem> {
```
Stypox: Shouldn't `getName()` represent the name of an item? In this case if I understand correctly it is being used only to pass information about the tab to the app. For that purpose using a shared enum or some integer constants would be a better fit, since those can't contain typos. So I would add a `getTabType()`  base function here. In case that's not done, I'd suggest adding a notice here about the different usage of `getName()`.

extractor/src/test/java/org/schabi/newpipe/extractor/services/DefaultTests.java
```diff
@@ -2,27 +2,31 @@
...
-import static org.junit.Assert.*;
-import static org.schabi.newpipe.extractor.ExtractorAsserts.*;
-import static org.schabi.newpipe.extractor.StreamingService.*;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.schabi.newpipe.extractor.ExtractorAsserts.assertEmptyErrors;
+import static org.schabi.newpipe.extractor.ExtractorAsserts.assertIsSecureUrl;
+import static org.schabi.newpipe.extractor.ExtractorAsserts.assertNotEmpty;
```
Stypox: Wildcard imports are ok here

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeChannelExtractorTest.java
```diff
@@ -5,15 +5,23 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 import static org.schabi.newpipe.extractor.ExtractorAsserts.assertIsSecureUrl;
 import static org.schabi.newpipe.extractor.ServiceList.YouTube;
-import static org.schabi.newpipe.extractor.services.DefaultTests.*;
+import static org.schabi.newpipe.extractor.services.DefaultTests.defaultTestGetPageInNewExtractor;
+import static org.schabi.newpipe.extractor.services.DefaultTests.defaultTestMoreItems;
+import static org.schabi.newpipe.extractor.services.DefaultTests.defaultTestRelatedItems;
```
Stypox: Also here, wildcard imports are ok

mauriciocolli:
> I don't think so, since not all channels have the same tabs, so I think instantiating ChannelTabs from the Channels instead of the StreamingServices is the way to go.

> I've been thinking about lazy-loading and I think we should add a function to ChannelTabInfo to populate fields other than the name and call that only once that tab is selected in NewPipe.

Isn't `Info` objects supposed to be completed objects, I can't think of a reason to not make it just like kiosks. I don't know if we even need the `LinkHandler` stuff though, some other strategy seems to be the way to go.

And why store a extractor instance and even a function to load it when we could use the current architecture? Unless you want to change all that? I think another PR would be  welcome to introduce such a change (the kiosks implementation is definitely not perfect).

For example, here:

https://github.com/TeamNewPipe/NewPipeExtractor/blob/a657d9763bcded5d40efc7f45ead234b63772398/extractor/src/main/java/org/schabi/newpipe/extractor/channel/ChannelTabInfo.java#L33-L37

Why do we need a loaded `ChannelExtractor` when YouTube gives us a complete continuation response? This is up to the individual tab to decide if needs or not (if it is needed, it'd just fetch the initial page again, like it was done in the old channel extractor).

> I don't think we should implement sub-tabs, as that's just unnecessary complexity imho.

That more like a filter than a sub-tab though, it'd reload just like it's done in search now.

extractor/src/main/java/org/schabi/newpipe/extractor/channel/ChannelInfo.java
wb9688: Yes, that's hardcodes, see [this conversation](https://github.com/TeamNewPipe/NewPipeExtractor/pull/279#discussion_r394446468), and I'll change that now.

wb9688:
@mauriciocolli: We need the `ChannelTabExtractor` to load a next page. In kiosks, we also need a `KioskExtractor` in `KioskInfo`, but they're just recreated every time in `getMoreItems()`, which is possible there. Since `ChannelTabExtractor`s are only retrievable through a `ChannelExtractor`, it wouldn't make sense that way, as we would need to fetch the channel page to know what tabs are available and then select the correct one. We don't want to fetch the channel page again every time we're scrolling down on a tab, so we store the `ChannelTabExtractor`. There's no other way to instantiate the `ChannelTabExtractor`, though we do indeed only need the `nextPageUrl`. Do you have any better solution that allows dynamic tabs for each channel, and passing data from the `ChannelExtractor` to the `ChannelTabExtractor` for the initial page?

Also, the Extractor class requires there to be a LinkHandler, which we're of course extending, so yes, we do need it.

extractor/src/test/java/org/schabi/newpipe/extractor/services/DefaultTests.java
wb9688: My IDE changes that automatically.

extractor/src/main/java/org/schabi/newpipe/extractor/channel/ChannelTabExtractor.java
wb9688: Isn't e.g. "Videos" the name of the tab?

mauriciocolli:
> We need the ChannelTabExtractor to load a next page. In kiosks, we also need a KioskExtractor in KioskInfo, **_but they're just recreated every time in getMoreItems(), which is possible there_**.

> Also, the Extractor class requires there to be a LinkHandler, which we're of course extending, so yes, we do need it.

Yep, that's what should be done here then.

> Since ChannelTabExtractors are only retrievable through a ChannelExtractor, it wouldn't make sense that way, as we would need to fetch the channel page to know what tabs are available and then select the correct one. We don't want to fetch the channel page again every time we're scrolling down on a tab, so we store the ChannelTabExtractor. There's no other way to instantiate the ChannelTabExtractor, though we do indeed only need the nextPageUrl. Do you have any better solution that allows dynamic tabs for each channel, and passing data from the ChannelExtractor to the ChannelTabExtractor for the initial page?

I suggested a solution above, https://github.com/TeamNewPipe/NewPipeExtractor/pull/279#issuecomment-600945958:
>> As for the lists, put a loaded tab when it doesn't need to be fetched again, and some kind of placeholder when it needs to (with its tab's id/url to instantiate later).

I did some rough tests and it worked out fine. Instead of returning a complete tab info, a placeholder would be returned instead. When the frontend checks that it is just a placeholder, it would just fetch the complete tab by using the placeholder info.

That way, we can still offer a way to not duplicate request for tabs that are available in the first request.

mauriciocolli:
@wb9688: About that example that you requested, I did a very rough proof of concept of what I was talking about, you can find it on [this branch at my fork](https://github.com/mauriciocolli/NewPipeExtractor/tree/pr279-tabs).

ingingin:
Hey @wb9688 @Stypox @TobiGr @B0pol please revive this, and include it in **0.20.0** again the great 0.20.0 release should be released with this one aswell. Honestly pls?

B0pol:
We have already big changes for 0.20 for NewPipe part (unified player, media notifications, SAF), then we don't plan to add it back in 0.20

MD77MD:
I agree 👍... we should save some feature for future releases... also to avoid delaing 0.20 any longer.

MD77MD:
I think this should be the feature highlight of newPipe 0.20.1

Stypox:
@MD77MD usually breaking changes happen in major releases, like 0.20.0 with unified ui and iirc 0.19.0 for the feed. Minor releases like 0.20.1 usually focus on bugfixing and small improvements. So this would go into 0.21.0, probably.

MD77MD:
I see... thank you for clarifying the difference 😄

MD77MD:
@wb9688 is there a reason we're closing this?

xibr:
Something wrong with @wb9688 has closed many pull requests.

selurvedu:
> wb9688 wants to merge 17 commits into TeamNewPipe:dev from unknown repository

Looks like wb's NewPipe repo got deleted and all corresponding PRs got closed.

MD77MD:
@Stypox @mhmdanas @TobiGr 

could we still use this

mhmdanas:
Yes, it is possible to base a PR on another one (in fact, that is what has been done for the SAF PR). That being said, please don't ping me when I am not participating unless for good reason. Other people already subscribed to this thread can answer your question too.

PR #268, opened 2020-03-01
Introduce Image class and provide all available images

wb9688:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I did test the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).
- [x] I agree to ASAP create a PULL request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) for making in compatible when I changed the api.

See TeamNewPipe/NewPipe#3121.

This PR will introduce an Image class, so that NewPipeExtractor will also return the width and height of an image, which NewPipe could use to determine which image to load. It'll also provide a list of all available images (i.e. different resolutions), but that's not implemented yet. Also, currebtly I've only implemented getting the actual width and height of the image for YouTube, I still have to look into the other services.

~~@TobiGr: Don't merge this for v0.18.6 though, this will be for v0.19.0 (or whatever we end up calling the next proper non-hotfix release).~~

PR #280, opened 2020-03-07
Extractor for youtube mix (auto-generated playlist)

XiangRongLin:

closes  TeamNewPipe/NewPipe#2895

## Problem
As describes in the issue in the NewPipe repo, when sharing a youtube mix to newpipe, it crashes. Reason for that is that the linkhandler receives (1) https://www.youtube.com/watch?v=QXCE3Betyug&list=RDQXCE3Betyug and converts it to (2) https://www.youtube.com/playlist?list=RDQXCE3Betyug , which loads a basicly empty page.

## Changes
- The LinkHandler doesn't convert mix urls.
- New PlaylistExtractor to extract playlist information from the url (1).
- The YoutubeService uses the new extractor if it is a mix.

## Testing
I tested the changes on my android 10 and 7.1 devices and a 5.0 emulator, by:
- sharing a link from my browser to NewPipe
- saving the playlist and opening it from the bookmarks fragment
- playling the playlist (didn't play on 5.0 emulator, but i can't play ANY video here)

@Iamdeadlyz could you test the apk.
[app-debug.zip](https://github.com/TeamNewPipe/NewPipeExtractor/files/5683657/app-debug.zip)

## What's missing
- Mixes are generated based on the user. Meaning if I open the site once while logged in and once while in incognito mode, it gives me different playlists.
- [x] The size of a mix is infinity, but getStreamCount returns a long. So i don't really know how to handle that. I would let that method return -1 and let the frontend interpret that as infinity.
- [x] Check if "My Mix" is getting handled correctly. Here the list id starts with "RDMM".

## Notes
- In the mix url (1) above you can see the playlist id is just the video id with "RD" at the start.
All mix urls that i found so far are like that, where the playlist id is "RD" + video id of the first video. I assume RD stands for radio (Playlist ids start with "PL" instead)
- Since i can't reopen PR (#253) because i force pushed, i'll add the comments from @B0pol here:
1. Uploader name should be YouTube
2. Uploader avatar url should point to the YouTube logo  
- I noticed that there are quite a few duplicates if you continue scrolling down in a mix, especially the first video of the mix is repeated often.

Related PR: #276
Depends on: TeamNewPipe/NewPipe#3243


Iamdeadlyz:
Hi @XiangRongLin 

Thanks for the ping! The debug app works well with the auto-generated playlist. I can confirm that the issue is now fixed.

PeterHindes:
Hey @XiangRongLin isn't "my mix" based on your youtube profile? How would that work in new-pipe at all?

XiangRongLin:
@PeterHindes It just queues videos in the same genre. It's not very good at selecting them but it still works.

And currently you can only share mix urls to newpipe and not start a mix from newpipe itself, so you will always have a starting video on which the mix is based on.

I also just realized that i can't install my debug apk for whatever reason. I'll look into it on the weekend.

Stypox:
> I also just realized that i can't install my debug apk for whatever reason. I'll look into it on the weekend

I don't know why, but sometimes it happens. You can fix this by clicking "Clean Project" and rebuilding normally.

wb9688:
@Stypox: I think this happens when you compile the app to run the activity again on your phone, if you just do a full build (no need to clean), it'll work again.

Stypox:
>  The size of a mix is infinity, but getStreamCount returns a long. So i don't really know how to handle that. I would let that method return -1 and let the frontend interpret that as infinity.

On IRC we are discussing if it would be a good idea to return `-2` in these cases, that would mean "unknown number bigger than 100" (in this case infinity)

XiangRongLin:
@Stypox
Thanks for the update.

In that case we could use static constants, so we don't have magic numbers floating around.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,182 @@
...
+            final String playlistId = playlistData.getString("playlistId");
+            final String videoId;
+            if (playlistId.startsWith("RDMM")) {
+                videoId = playlistId.substring(4);
+            } else {
+                videoId = playlistId.substring(2);
+            }
+            if (videoId.isEmpty()) {
+                throw new ParsingException("");
+            }
```
Stypox: I would move this piece of code into `getThumbnailUrlFromId`, and maybe throw a ParsingException with a description like "videoId is empty".

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractorTest.java
```diff
@@ -0,0 +1,225 @@
...
+        public void getName() throws Exception {
+            String name = extractor.getName();
+            assertEquals("My Mix", name);
+        }
+
+        @Test
+        public void getThumbnailUrl() throws Exception {
+            final String thumbnailUrl = extractor.getThumbnailUrl();
+            assertIsSecureUrl(thumbnailUrl);
+            assertThat(thumbnailUrl, startsWith("https://i.ytimg.com/vi/_AzeUSL9lZc"));
```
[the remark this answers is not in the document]
XiangRongLin: Because the extractor is set up with a starting video (watch?v=videoId). The rest of the mix is indeed different for everyone, but because there is a starting video, the first one will always be that one.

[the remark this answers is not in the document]
Stypox: Oh, that's right, sorry ;-)

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeServiceTest.java
```diff
@@ -54,4 +58,24 @@ public void testGetKioskAvailableKiosks() throws Exception {
...
+    @Test
+    public void getPlaylistExtractorIsMix() throws Exception {
+        String videoId = "_AzeUSL9lZc";
+        PlaylistExtractor extractor = YouTube.getPlaylistExtractor(
+            "https://www.youtube.com/watch?v=" + videoId + "&list=RD" + videoId);
+        assertTrue(extractor instanceof YoutubeMixPlaylistExtractor);
+
+        extractor = YouTube.getPlaylistExtractor(
+            "https://www.youtube.com/watch?v=" + videoId + "&list=RD" + videoId);
+        assertTrue(extractor instanceof YoutubeMixPlaylistExtractor);
```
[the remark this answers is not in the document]
XiangRongLin: Because i may have not paid attention and forgot to make the second one into RDMM for My Mix.

I can add the test.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,174 @@
...
+import org.schabi.newpipe.extractor.stream.StreamInfoItem;
+import org.schabi.newpipe.extractor.stream.StreamInfoItemsCollector;
+
+/**
+ * A YoutubePlaylistExtractor for a mix (auto-generated playlist). It handles urls in the format of
+ * "youtube.com/watch?v=videoId&list=playlistId"
+ */
+public class YoutubeMixPlaylistExtractor extends PlaylistExtractor {
+
+    public final static Long MORE_THAN_100 = -2L;
```
B0pol: Why are you using ``Long`` instead of ``long``?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,174 @@
...
+            .getObject(PLAYLIST).getObject(PLAYLIST);
+    }
+
+    @Nonnull
+    @Override
+    public String getName() throws ParsingException {
+        final String name = playlistData.getString("title");
+        if (name == null) {
+            throw new ParsingException("Could not get playlist name");
+            }
```
B0pol:
```suggestion
        }
```

XiangRongLin:
@Stypox 
I applied your suggestions and updated to apk.
I'll leave the changes in the frontend to you. I have no clue how to integrate the "+" in "100+", so that it works in all languages.

I just saw that my CI build failed because i got a reCaptcha challange. The "getPageMultipleTimes" test may need to be removed.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
XiangRongLin: Kotlin habit

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeService.java
```diff
@@ -111,7 +115,7 @@ public KioskList getKioskList() throws ExtractionException {
                 public KioskExtractor createNewKiosk(StreamingService streamingService,
                                                      String url,
                                                      String id)
-                        throws ExtractionException {
+                    throws ExtractionException {
```
Stypox: One more tab ;-)
Wraps should be indented with two more steps, otherwise they can be confused with the content below

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,174 @@
...
+import org.schabi.newpipe.extractor.stream.StreamInfoItem;
+import org.schabi.newpipe.extractor.stream.StreamInfoItemsCollector;
+
+/**
+ * A YoutubePlaylistExtractor for a mix (auto-generated playlist). It handles urls in the format of
+ * "youtube.com/watch?v=videoId&list=playlistId"
+ */
+public class YoutubeMixPlaylistExtractor extends PlaylistExtractor {
+
+    public final static long MORE_THAN_100 = -2L;
```
Stypox: This would go in the base playlist extractor, so that it can be used by every service. Also, I just realized maybe a better name is `MORE_THAN_100_ITEMS`.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeService.java
XiangRongLin: Done.

I really hope we can have a code style file, so I can press autoformat again without worrying.

Stypox: For some reason this in not yet done, even though GitHub told me the was a change on that line. But if on Android Studio you see everything as expected, it is ok. Is not so important ;-)

Stypox:
I opened the corresponding client PR: TeamNewPipe/NewPipe#3243
Debug apk: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4344911/app-debug.zip)

extractor/src/main/java/org/schabi/newpipe/extractor/playlist/PlaylistExtractor.java
```diff
@@ -8,6 +8,8 @@
 
 public abstract class PlaylistExtractor extends ListExtractor<StreamInfoItem> {
 
+    public final static long MORE_THAN_100_ITEMS = -2L;
```
Stypox:
```suggestion
    public final static long INFINITE_ITEMS = -2L;
    public final static long MORE_THAN_100_ITEMS = -3L;
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,173 @@
...
+    @Override
+    public String getUploaderAvatarUrl() {
+        //Youtube mix are auto-generated by YouTube
+        return "https://upload.wikimedia.org/wikipedia/commons/thumb/0/09/YouTube_full-color_icon_(2017).svg/320px-YouTube_full-color_icon_(2017).svg.png";
+    }
+
+    @Override
+    public long getStreamCount() {
+        // Auto-generated playlist always start with 25 videos and are endless
+        return MORE_THAN_100_ITEMS;
```
Stypox:
```suggestion
        return INFINITE_ITEMS;
```

XiangRongLin:
@Stypox Feel free to just commit small stuff like that in the future.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,173 @@
...
+    @Override
+    public String getUploaderName() {
+        //Youtube mix are auto-generated by YouTube
+        return "YouTube";
+    }
+
+    @Override
+    public String getUploaderAvatarUrl() {
+        //Youtube mix are auto-generated by YouTube
+        return "https://upload.wikimedia.org/wikipedia/commons/thumb/0/09/YouTube_full-color_icon_(2017).svg/320px-YouTube_full-color_icon_(2017).svg.png";
```
TobiGr: When using YouTube's logo here, we can also use in other places. We should check legal stuff first. CC @TheAssassin

TobiGr: I think this should be handled differently in the frontend by not displaying an avatar. @Stypox, @wb9688 What do you think?

TheAssassin: IANAL. However, using any brand icons is a bad idea, and might get us into trouble. I don't think it's worth the effort. I haven't had a look whether the linked file would be fine copyright wise either.

TheAssassin (dismissed):

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
Stypox: Ok, so either we put no avatar (which would probably look strange) or we choose an icon from https://material.io/icons that somewhat represents the nature of the playlist and the uploader. I thought of these icons:
- [Radio](https://material.io/resources/icons/?icon=radio): 
- [Auto renew](https://material.io/resources/icons/?icon=autorenew):

Stypox: @XiangRongLin what do you think about this?

XiangRongLin: I'm for either the Radio icon or no icon. Radio since the mix is basically a radio.
But that depends on which looks less out of place.

Stypox:
@XiangRongLin I moved things around once again ;-)

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractorTest.java
```diff
@@ -0,0 +1,241 @@
...
+
+public class YoutubeMixPlaylistExtractorTest {
+
+    private static YoutubeMixPlaylistExtractor extractor;
+    private static String videoId = "_AzeUSL9lZc";
+    private static String videoTitle = "Most Beautiful And Emotional  Piano: Anime Music Shigatsu wa Kimi no Uso OST IMO";
+
+    public static class Mix {
+
+        @Before
```
Stypox: These should be `@BeforeClass`, otherwise the testing time skyrockets (it just took me 2 minutes just to run this class).

XiangRongLin: Yeah, that makes a lot of sense. That probably is also the reason why the recaptcha are appearing.

Stypox: May I change it or will you do it?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
TobiGr: whatever icons we use, that should be set in the app. It might be okay to return the URL to YouTube's logo in the extractor. However, it should be replaced in the frontend. Our extractor is used by other applications, too. They might want to use the original logo.

Stypox: Ok, that seems a good idea

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractorTest.java
XiangRongLin: @Stypox Just changing it as is, produces NullPointerExceptions in some test. I'll have to look into it.

wb9688: I think you need to add it for all 3 tests.

XiangRongLin: @Stypox I changed it. Was because i had 2 tests for invalid urls, which set the data to null

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
wb9688: I don't think we should return a link to YouTube's logo.

wb9688:
Does this support channel mixes (e.g. https://www.youtube.com/watch?v=mnk6gnOBYIo&list=RDCMUCXuqSBlHAE6Xw-yeJA0Tunw&start_radio=1&t=0)? You don't have tests for them.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeService.java
```diff
@@ -111,7 +115,7 @@ public KioskList getKioskList() throws ExtractionException {
                 public KioskExtractor createNewKiosk(StreamingService streamingService,
                                                      String url,
                                                      String id)
-                        throws ExtractionException {
+                throws ExtractionException {
```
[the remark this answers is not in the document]
wb9688: This hasn't been resolved?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java
```diff
@@ -163,6 +163,16 @@ public static Calendar parseDateFrom(String textualUploadDate) throws ParsingExc
...
     }
 
+    /**
+     * Checks if the given playlist id is a mix (auto-generated playlist)
+     * Ids from a mix start with "RD"
+     * @param playlistId
+     * @return Whether given id belongs to a mix
+     */
+    public static boolean isYoutubeMixId(String playlistId) {
+        return playlistId.startsWith("RD");
```
wb9688: `&& !playlistId.startsWith("RDAMVM")`, so that it won't accept YouTube Music mixes. You'll also need to do that in the `LinkHandler`, so that NewPipe will open those as just a video until we implement YouTube Music mixes.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,175 @@
...
+ * "youtube.com/watch?v=videoId&list=playlistId"
+ */
+public class YoutubeMixPlaylistExtractor extends PlaylistExtractor {
+
+
+    private final static String CONTENTS = "contents";
+    private final static String RESPONSE = "response";
+    private final static String PLAYLIST = "playlist";
+    private final static String TWO_COLUMN_WATCH_NEXT_RESULTS = "twoColumnWatchNextResults";
+    private final static String PLAYLIST_PANEL_VIDEO_RENDERER = "playlistPanelVideoRenderer";
```
wb9688: Why are you storing those as constants here? We don't do that anywhere else either…

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,175 @@
...
+        //Because continuation requests are created with the last video of previous request as start
+        streams.remove(0);
+        collectStreamsFrom(collector, streams);
+        return new InfoItemsPage<>(collector, getNextPageUrl());
+    }
+
+    private void collectStreamsFrom(
+            @Nonnull StreamInfoItemsCollector collector,
+            @Nullable JsonArray streams) {
+        collector.reset();
```
wb9688: You don't need to call `collector.reset()` here as every time you call it, you already have a new empty collector.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
XiangRongLin: Because i would sometime make a typo in those strings and wonder for 10 minutes where my error was. With the constants i can let my IDE autocomplete for me.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,175 @@
...
+    public InfoItemsPage<StreamInfoItem> getPage(final String pageUrl)
+            throws ExtractionException, IOException {
+        if (pageUrl == null || pageUrl.isEmpty()) {
+            throw new ExtractionException(
+                new IllegalArgumentException("Page url is empty or null"));
+        }
+
+        StreamInfoItemsCollector collector = new StreamInfoItemsCollector(getServiceId());
+        final JsonArray ajaxJson = getJsonResponse(pageUrl, getExtractorLocalization());
+        playlistData =
```
wb9688: Why are you assigning `playlistData` again here? If you call `getInitialPage()` after that, it won't return the initial page.

XiangRongLin:
> 
> 
> Does this support channel mixes (e.g. https://www.youtube.com/watch?v=mnk6gnOBYIo&list=RDCMUCXuqSBlHAE6Xw-yeJA0Tunw&start_radio=1&t=0)? You don't have tests for them.

@wb9688 From a quick look, it seems to mostly work. Only getThumbnailUrl would need to be adjusted.

But how do you even start a channel mix? Are there other mixes besides this, yt music mix, and my mix

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,175 @@
...
+    @Override
+    public String getNextPageUrl() throws ExtractionException {
+        final JsonObject lastStream = ((JsonObject) playlistData.getArray(CONTENTS)
+            .get(playlistData.getArray(CONTENTS).size() - 1));
+        if (lastStream == null || lastStream.getObject(PLAYLIST_PANEL_VIDEO_RENDERER) == null) {
+            throw new ExtractionException("Could not extract next page url");
+        }
+        return "https://youtube.com" + lastStream.getObject(PLAYLIST_PANEL_VIDEO_RENDERER)
+            .getObject("navigationEndpoint").getObject("commandMetadata")
+            .getObject("webCommandMetadata").getString("url") + "&pbj=1";
```
wb9688: You should call `getUrlFromNavigationEndpoint()` and just append `&pbj=1`. You'll first need to modify `getUrlFromNavigationEndpoint()` so that it handles the `index` parameter.

wb9688:
@XiangRongLin: I saw a channel mix somewhere in "Up next". I have no idea if there are any other kinds of YouTube mixes.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
XiangRongLin: Because i needed the updated `playlistData` for when `getNextPageUrl()` gets called multiple times.
I'just add a `initialPlaylistData` variable to the class

wb9688: `getNextPageUrl()` should always be relative to the initial page: https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/main/java/org/schabi/newpipe/extractor/ListExtractor.java#L30

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
wb9688: Please just follow the style of the rest of the extractors.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
XiangRongLin: oh, I completly missed that. I'll fix it

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
XiangRongLin: I just saw that I already handled the index incorrectly, since i assumed the index in the url would count upwards. But it always stays at 25.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
opusforlife2: My vote goes to the Radio icon as well.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
XiangRongLin: Is there a reason for using `&amq;` instead of just `&` since this is throwing a malformed url exception for me when i create new extractor with the return value of `getUrlFromNavigationEndpoint()`

https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java#L302

wb9688: Yes, there's a reason, but it's wrong. That function should return it with `&`, which should then be replaced with `&amp;` at https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java#L319.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractorTest.java
```diff
@@ -0,0 +1,325 @@
...
+            assertThat(streams.hasNextPage(), is(true));
+        }
+
+        @Test
+        public void getPageMultipleTimes() throws Exception {
+            InfoItemsPage<StreamInfoItem> streams = extractor.getPage(extractor.getNextPageUrl());
+
+            //Should work infinitely, but for testing purposes only 3 times
+            for (int i = 0; i < 3; i++) {
+                assertThat(streams.hasNextPage(), is(true));
```
wb9688: Use `assertTrue` and `assertFalse`

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    private String getNextPageUrlFrom(JsonObject playlistData) throws ExtractionException {
+        final JsonObject lastStream = ((JsonObject) playlistData.getArray("contents")
+            .get(playlistData.getArray("contents").size() - 1));
+        if (lastStream == null || lastStream.getObject("playlistPanelVideoRenderer") == null) {
+            throw new ExtractionException("Could not extract next page url");
+        }
+        //Index of video in mix is missing, but adding it doesn't appear to have any effect.
+        //And since the index needs to be tracked by us, it is left out
```
wb9688: I still think `getUrlFromNavigationEndpoint` needs to put the index in the URL. You actually don't need to track it, you just need to do `++navigationEndpoint.getObject("watchEndpoint").getInt("index")`.

XiangRongLin: With tracking i mean the behaviour you get when you play a mix on youtube.com
If you always select the last video in the mix the index jumps from 1 to 25 to 49 to 50 and then stays at 50.
But the index value in watchEndpoint is always 24.

wb9688: I don't think so: 
yt-mix-json

XiangRongLin: Is that in the browser?
Since i couldn't get it to simulate that behaviour with the extractor

wb9688:
Also, why did you make an `extractors` package just for `YoutubeMixPlaylistExtractorTest`? Just put that test in `org.schabi.newpipe.extractor.services.youtube` like the others.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
wb9688: Yes, that's in Firefox. I'll try to reproduce.

XiangRongLin:
> 
> 
> Also, why did you make an `extractors` package just for `YoutubeMixPlaylistExtractorTest`? Just put that test in `org.schabi.newpipe.extractor.services.youtube` like the others.

@wb9688 I let IntelliJ autogenerate it and then it is put into the same package as the source class is, which makes sense in my view.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
XiangRongLin: Then i would be for returning a constant and each frontend can decide themself what they want to display based on the constant and legal stuff.
For NewPipe it would then be the radio icon.

TobiGr:
> I let Android Studio autogenerate it and then it is put into the same package as the source class is, which makes sense in my view.

I agree, but please stick to the current structure. We can improve it in a later PR.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractorTest.java
TobiGr: IMO that's okay. We can keep it

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
TobiGr: We should not use the wikimedia image though. Here are the icons which are used on the website when the browser does not support SVG:
https://s.ytimg.com/yts/img/favicon_144-vfliLAfaB.png
https://s.ytimg.com/yts/img/favicon_96-vflW9Ec0w.png
https://s.ytimg.com/yts/img/favicon_48-vflVjB_Qk.png
https://s.ytimg.com/yts/img/favicon_32-vflOogEID.png

> Then i would be for returning a constant and each frontend can decide themself what they want to display based on the constant and legal stuff.

@XiangRongLin What constant? Can you elaborate the idea? One can check whether the retrieved PlaylistInfo is a YTMix by using `YouTubeParsingHelper.isYouTubeMixId(id)`

XiangRongLin: I meant as an arbitary string like `public static final String YOUTUBE_LOGO = "youtube_logo"`

Nice finding the png, i somehow only found svg.

wb9688: @gzsombor: We would like to know your opinion on this as well.

I personally think it would be best to return "" here, but @TobiGr thinks we should return a link to YouTube's logo, as we always return a URL for all other kind of playlists, however @TheAssassin thinks we should avoid logos/brands.

Either way, in the NewPipe front-end, we're gonna check if service ID equals YouTube's service ID and call `YoutubeParsingHelper`'s function to check whether the playlist ID is a mix. When that's the case, NewPipe will replace it with a radio icon embedded in our app.

mauriciocolli (dismissed):
> I noticed that there are quite a few duplicates if you continue scrolling down in a mix, especially the first video of the mix is repeated often.

Yeah, I was going to point it out.

I'm thinking towards cookies, maybe we should actually extend the options to get next pages, not every service will need only the url (although all of them work this way currently), this is a case where this depends on it.

But here's the thing right now: it's pretty much useless to return next pages due to so many duplicates.

I was doing tests with the cookies option, and actually the next page returns more items: the ones above and below the one that it's currently being fetched (so the next part start through the middle).

Impossible to fix it currently, only option is to either implement it or just disable the next page feature for now.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
mauriciocolli: Why even return anything here?

This should return an empty string just like the uploader url and name, and [this is actually already handled by the front end](https://github.com/TeamNewPipe/NewPipe/blob/0a412e2abbc1d4cd50439ec89070a69a794af7ba/app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java#L265-L281).

TobiGr: @gzsombor As a user of the extractor, what solution do you prefer?
The YouTube logo, an empty string or a constant?

@mauriciocolli @wb9688 I'd be okay with returning nothing, too

gzsombor: What about returning a AvatarIcon class, something like this:
```
class AvatarIcon {
   String url;
   IconType type;
   enum IconType { YOUTUBE_LOGO, OTHER_LOGO, .... }
}

```

Either the url or the type would be filled - if the URL the frontend could use that, if not, it should revert to some custom magic.

XiangRongLin:
> 
> 
> > I noticed that there are quite a few duplicates if you continue scrolling down in a mix, especially the first video of the mix is repeated often.
>
>
> Impossible to fix it currently, only option is to either implement it or just disable the next page feature for now.

@mauriciocolli 
I also managed to implement it [like this](https://github.com/XiangRongLin/NewPipeExtractor/commit/4ac47da62c9424650f6af383d4e7f1a13912ef37), but the only problem i have left is that `PlaylistInfo.getMoreItems()` returns a new instance of the extractor. This way i can't have the cookie be a member variable and reference it, when the user scrolls down and requests more items.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
TobiGr: @gzsombor That's basically done in https://github.com/TeamNewPipe/NewPipeExtractor/pull/268. So we should wait for this PR to be merged. Let's go with the empty string for now and change this later in or after #268


@XiangRongLin Can you change this to 
```suggestion
        return "";
```

Stypox: Ok, I was about to ask ;-)

Stypox:
Oh no, I messed up everything... @wb9688 is there a way to solve this?

XiangRongLin:
@Stypox Was something actually changed through the force push or can i just rebase my stuff on top this?

mauriciocolli:
@XiangRongLin

The solutions that you proposed could work, but using static variables is not really ideal for the intended use case of those parameters. They can be cleared at any moment, (i.e. application killed in the background because phone ran out of memory), so they are not reliable enough. We should avoid introducing a dependence on the state of the application (not necessarily Android).

What I suggested is that we should save all the necessary information for later, just like it's done with the next page url right now, so we can retrieve it at any time, even across process death.

I already started working on it to make this functionality possible, a rough prototype is in order right now, I think I'll have at least a draft pull request ready soon™.

---

_Offtopic:_


>@Stypox: Why am I considered as 2nd committer of all of these commits?

Rather @XiangRongLin is the author and you the committer, because, well, you did just that while rebasing it.

Also @XiangRongLin, forgot to ask you about this:

> Since i can't reopen PR (#253) because i force pushed

I noticed this happening a few more times (e.g. TeamNewPipe/NewPipe#3194) and I keep wondering if I missed a warning or something in them.

Isn't Github supposed to just show the differences between it and the target branch on the target repository? It doesn't matter in which commit you opened the pull request or when.

What exactly kept you from just force pushing the source branch again (in this case, in your fork)? It doesn't seem to be locked or anything, so you should be able to do just that.

Just curious to know what happened.

XiangRongLin:
@mauriciocolli 
> Also @XiangRongLin, forgot to ask you about this:
> 
> > Since i can't reopen PR (#253) because i force pushed
> 
> 
> Isn't Github supposed to just show the differences between it and the target branch on the target repository? It doesn't matter in which commit you opened the pull request or when.
> 
> What exactly kept you from just force pushing the source branch again (in this case, in your fork)? It doesn't seem to be locked or anything, so you should be able to do just that.

I closed the PR on my own, since i needed to change it to the new user agent. After doing that and force pushing, the "reopen" button was greyed out. It sayed that since i force pushed the branch, I couldn't reopen the PR.

TobiGr:
> Was something actually changed through the force push or can i just rebase my stuff on top this?

You could cherry-pick the changes if needed. I am not sure if rebasing is possible.

> I closed the PR on my own, since i needed to change it to the new user agent. After doing that and force pushing, the "reopen" button was greyed out. It sayed that since i force pushed the branch, I couldn't reopen the PR.

Did you delete the branch in your GitHub repository? If one does that, it is not possible to reopen a PR.

Stypox:
@XiangRongLin I added the `Use ITEM_COUNT_UNKNOWN everywhere instead of -1` commit.
@opusforlife2 found a problem, but I am unsure if/how it needs to be solved:
> While testing, I found that Newpipe doesn't show private videos in the playlist, but counts them in the tally. For example: https://www.youtube.com/playlist?list=PL3CB4CB7A114C1070

@wb9688 what are we doing elsewhere with private videos? Are we just ignoring them? It might be ok to leave this as it is, since private videos are not playable anyway.

wb9688:
@Stypox: We're not ignoring them, we're returning `true` for them in `isAd()`, which was also the case before yt_new.

Stypox:
Oh ok. Then that's not a problem of this PR, as it is consistent with the rest of the extractor. @opusforlife2

wb9688:
@Stypox: That link is not even a mix. Plus this is handled in YoutubeStreamInfoItemExtractor, which hasn't changed in this PR I think.

opusforlife2:
Oh. OH. Sorry, I used ambiguous wording here.

> While testing, I found that Newpipe doesn't show private videos in the playlist, but counts them in the tally. For example: https://www.youtube.com/playlist?list=PL3CB4CB7A114C1070

By Newpipe here, I meant the release version. Not the debug apk you provided, @Stypox.

XiangRongLin:
> 
> @mauriciocolli 
> Did you delete the branch in your GitHub repository? If one does that, it is not possible to reopen a PR.

No. I rebased and sqashed the commits locally, so that the commits with the old user agent won't appear anymore and force pushed these changes onto the same branch

TobiGr:
Heads up @XiangRongLin. @wb9688 and I created https://github.com/TeamNewPipe/NewPipeExtractor/commit/647e7cd45054921712d6c6e592f39f70a1289e3b to reject mixes for now. Parts of that should be reverted by you in this PR.

XiangRongLin:
> @wb9688 What I suggested is that we should save all the necessary information for later, just like it's done with the next page url right now, so we can retrieve it at any time, even across process death.
>
>I already started working on it to make this functionality possible, a rough prototype is in order right now, I think I'll have at least a draft pull request ready soon™.

I'll then just stop this PR here for now until that is implemented and then rebase this PR on top of that.

> @TobiGr  Heads up @XiangRongLin. @wb9688 and I created 647e7cd to reject mixes for now. Parts of that should be reverted by you in this PR.

I'll do that when I rebase the part above.

Stypox:
I rebased, squashed and added the newly-added subchannel functions. Now there are 11 commits instead of 26 ;-)

XiangRongLin:
I'll finish this at the start of next month since i have exams now.

XiangRongLin:
I adjusted my extractor to the `Page` class and my tests for that change and the change to getNextPageUrl.
Also a lot of misc checkstyle stuff.
The new APK: 
[app-debug.zip](https://github.com/TeamNewPipe/NewPipeExtractor/files/5010013/app-debug.zip)

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
```diff
@@ -195,6 +202,26 @@ public static Calendar parseDateFrom(String textualUploadDate) throws ParsingExc
         return uploadDate;
     }
 
+    /**
+     * Checks if the given playlist id is a youtube mix (auto-generated playlist)
+     * Ids from a youtube mix start with "RD"
+     * @param playlistId
+     * @return Whether given id belongs to a youtube mix
```
TobiGr, line 209:
```suggestion
     * Checks if the given playlist id is a YouTube Mix (auto-generated playlist)
     * Ids from a YouTube Mix start with "RD"
     * @param playlistId
     * @return Whether given id belongs to a YouTube Mix
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
```diff
@@ -195,6 +202,26 @@ public static Calendar parseDateFrom(String textualUploadDate) throws ParsingExc
...
+     */
+    public static boolean isYoutubeMixId(String playlistId) {
+        return playlistId.startsWith("RD") && !isYoutubeMusicMixId(playlistId);
+    }
+
+    /**
+     * Checks if the given playlist id is a youtube music mix (auto-generated playlist)
+     * Ids from a youtube music mix start with "RD"
+     * @param playlistId
+     * @return Whether given id belongs to a youtube music mix
```
TobiGr, line 219:
```suggestion
     * Checks if the given playlist id is a YouTube Music Mix (auto-generated playlist)
     * Ids from a YouTube Music Mix start with "RD"
     * @param playlistId
     * @return Whether given id belongs to a YouTube Music Mix
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubePlaylistLinkHandlerFactory.java
```diff
@@ -67,4 +74,31 @@ public boolean onAcceptUrl(final String url) {
         }
         return true;
     }
+
+    /**
+     * If it is a mix (auto-generated playlist) url, return a Linkhandler where the url is like.
+     * youtube.com/watch?v=videoId&list=playlistId
```
TobiGr, line 80:
```suggestion
     * If it is a mix (auto-generated playlist) URL, return a {@link LinkHandler} where the URL is like
     * <code>https://youtube.com/watch?v=videoId&list=playlistId</code>
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubePlaylistLinkHandlerFactory.java
```diff
@@ -67,4 +74,31 @@ public boolean onAcceptUrl(final String url) {
...
+    public ListLinkHandler fromUrl(final String url) throws ParsingException {
+        try {
+            URL urlObj = Utils.stringToURL(url);
+            String listID = Utils.getQueryValue(urlObj, "list");
+            if (listID != null && YoutubeParsingHelper.isYoutubeMixId(listID)) {
+                String videoID = Utils.getQueryValue(urlObj, "v");
+                if (videoID == null) {
+                    videoID = listID.substring(2);
+                }
+                String newUrl = "https://www.youtube.com/watch?v=" + videoID + "&list=" + listID;
```
TobiGr, line 93:
```suggestion
                final String newUrl = "https://www.youtube.com/watch?v=" + videoID + "&list=" + listID;
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubePlaylistLinkHandlerFactory.java
```diff
@@ -67,4 +74,31 @@ public boolean onAcceptUrl(final String url) {
...
+    /**
+     * If it is a mix (auto-generated playlist) url, return a Linkhandler where the url is like.
+     * youtube.com/watch?v=videoId&list=playlistId
+     * <p>Otherwise use super</p>
+     */
+    @Override
+    public ListLinkHandler fromUrl(final String url) throws ParsingException {
+        try {
+            URL urlObj = Utils.stringToURL(url);
+            String listID = Utils.getQueryValue(urlObj, "list");
```
TobiGr, line 87:
```suggestion
            final String listID = Utils.getQueryValue(urlObj, "list");
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,228 @@
...
+
+import static org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper.extractCookieValue;
+import static org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper.getJsonResponse;
+import static org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper.getResponse;
+import static org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper.getUrlFromNavigationEndpoint;
+import static org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper.toJsonArray;
+
+/**
+ * A YoutubePlaylistExtractor for a mix (auto-generated playlist). It handles urls in the format of
+ * "youtube.com/watch?v=videoId&list=playlistId"
```
TobiGr, line 34:
```suggestion
 * A {@link YoutubePlaylistExtractor} for a mix (auto-generated playlist). It handles URLs in the format of
 * <code>youtube.com/watch?v=videoId&list=playlistId</code>
```

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeServiceTest.java
```diff
@@ -54,4 +58,30 @@ public void testGetKioskAvailableKiosks() throws Exception {
...
+    @Test
+    public void getPlayListExtractorIsNormalPlaylist() throws Exception {
+        PlaylistExtractor extractor = service.getPlaylistExtractor(
+            "https://www.youtube.com/watch?v=JhqtYOnNrTs&list=PL-EkZZikQIQVqk9rBWzEo5b-2GeozElS");
+        assertTrue(extractor instanceof YoutubePlaylistExtractor);
+    }
+
+    @Test
+    public void getPlaylistExtractorIsMix() throws Exception {
+        String videoId = "_AzeUSL9lZc";
```
TobiGr, line 72:
```suggestion
        final String videoId = "_AzeUSL9lZc";
```

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeServiceTest.java
```diff
@@ -54,4 +58,30 @@ public void testGetKioskAvailableKiosks() throws Exception {
     public void testGetDefaultKiosk() throws Exception {
         assertEquals(kioskList.getDefaultKioskExtractor(null).getId(), "Trending");
     }
+
+
+    @Test
+    public void getPlayListExtractorIsNormalPlaylist() throws Exception {
+        PlaylistExtractor extractor = service.getPlaylistExtractor(
```
TobiGr, line 65:
```suggestion
        final PlaylistExtractor extractor = service.getPlaylistExtractor(
```

XiangRongLin:
@TobiGr I applied your suggestions. When the PR gets merged that commit can be squashed into the previous one.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -30,8 +30,8 @@
 import static org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper.toJsonArray;
 
 /**
- * A YoutubePlaylistExtractor for a mix (auto-generated playlist). It handles urls in the format of
- * "youtube.com/watch?v=videoId&list=playlistId"
+ * A {@link YoutubePlaylistExtractor} for a mix (auto-generated playlist). It handles URLs in the format of
+ * <code>youtube.com/watch?v=videoId&list=playlistId"</code>
```
TobiGr, line 34:
```suggestion
 * <code>youtube.com/watch?v=videoId&list=playlistId</code>
```
Looks like I did not remove that in my suggestion.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
```diff
@@ -504,22 +535,65 @@ public static String getValidJsonResponseBody(final Response response)
...
+        if (!isNullOrEmpty(page.getCookies())) {
+            headers.put("Cookie", Collections.singletonList(join(";", "=", page.getCookies())));
+        }
+        headers.put("X-YouTube-Client-Name", Collections.singletonList("1"));
+        headers.put("X-YouTube-Client-Version", Collections.singletonList(getClientVersion()));
+
+        final Response response = getDownloader().get(page.getUrl(), headers, localization);
+        getValidJsonResponseBody(response);
+
+        return toJsonArray(response.responseBody());
```
Stypox, line 593: Like above
```suggestion
        return toJsonArray(getValidJsonResponseBody(response));
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,228 @@
...
+                return getThumbnailUrlFromPlaylistId(playlistId);
+            } catch (ParsingException e) {
+                //fallback to thumbnail of current video. Always the case for channel mix
+                return getThumbnailUrlFromVideoId(
+                    initialData.getObject("currentVideoEndpoint").getObject("watchEndpoint")
+                        .getString("videoId"));
+            }
+        } catch (Exception e) {
+            throw new ParsingException("Could not get playlist thumbnail", e);
+        }
```
Stypox, line 89: Do this, so that the original error is reported if even the fallback fails
```suggestion
        try {
            return getThumbnailUrlFromPlaylistId(playlistData.getString("playlistId"));
        } catch (final Exception e) {
            try {
                //fallback to thumbnail of current video. Always the case for channel mix
                return getThumbnailUrlFromVideoId(
                    initialData.getObject("currentVideoEndpoint").getObject("watchEndpoint")
                        .getString("videoId"));
            } catch (final Exception ignored) {
            }
            throw new ParsingException("Could not get playlist thumbnail", e);
        }
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,228 @@
...
+
+        final JsonArray ajaxJson = getJsonResponse(page, getExtractorLocalization());
+        final JsonObject playlistJson =
+                ajaxJson.getObject(3).getObject("response").getObject("contents")
+                        .getObject("twoColumnWatchNextResults").getObject("playlist")
+                        .getObject("playlist");
+        final JsonArray allStreams = playlistJson.getArray("contents");
+        // Sublist because youtube returns up to 24 previous streams in the mix
+        // +1 because the stream of "currentIndex" was already extracted in previous request
+        List<Object> newStreams =
```
Stypox, line 162: `final`

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,228 @@
...
+                allStreams.subList(playlistJson.getInt("currentIndex") + 1, allStreams.size());
+
+        StreamInfoItemsCollector collector = new StreamInfoItemsCollector(getServiceId());
```
Stypox, line 165: `final`

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,228 @@
...
+            @Nonnull final StreamInfoItemsCollector collector,
+            @Nullable final List<Object> streams) {
+
+        if (streams == null) {
+            return;
+        }
+
+        final TimeAgoParser timeAgoParser = getTimeAgoParser();
+
+        for (Object stream : streams) {
```
Stypox, line 181: `final`

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,228 @@
...
+            if (stream instanceof JsonObject) {
+                JsonObject streamInfo = ((JsonObject) stream)
```
Stypox, line 183: `final`

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
wb9688: Also use `{@code something}` instead of `<code>something</code>`

wb9688:
I'll do a review soon. I want to look at this before it gets merged.

XiangRongLin:
Does the checkstyle.xml from #368 (or from the main repo) not catch all the missing `final`s. I just set up the checkstyle plugin with that rule, but it didn't show me all of that.

wb9688:
@XiangRongLin: It only require all arguments of a function to be final

opusforlife2:
Will this fix https://github.com/TeamNewPipe/NewPipe/issues/3431?

XiangRongLin:
> 
> 
> Will this fix [TeamNewPipe/NewPipe#3431](https://github.com/TeamNewPipe/NewPipe/issues/3431)?

@opusforlife2 I don't think so. I haven't changed anything related to youtube music.
The new extractor only works for normal youtube, not music ones.

opusforlife2:
Ah. I forgot that music URLs are RDAMVM.

extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
```diff
@@ -235,4 +236,15 @@ public static String join(final CharSequence delimiter, final Iterable<? extends
         }
         return stringBuilder.toString();
     }
+
+    public static String join(final String delimiter, final String mapJoin,
```
Stypox, line 240: Maybe you could add a test for this to [UtilsTest.testJoin()](https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/test/java/org/schabi/newpipe/extractor/utils/UtilsTest.java). But you can leave this to another PR, since the code looks correct ;-)

Stypox (approved):
When you solve a review comment, be sure to mark it as resolved, so that I/others know the review was looked at and fixed.
@wb9688 please take a look if you have time, as this is basically finished.
Thank you @XiangRongLin for the maintained effort :-D

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
```diff
@@ -330,7 +358,7 @@ public static boolean areHardcodedYoutubeMusicKeysValid() throws IOException, Re
                 .end()
                 .value("query", "test")
                 .value("params", "Eg-KAQwIARAAGAAgACgAMABqChAEEAUQAxAKEAk%3D")
-            .end().done().getBytes("UTF-8");
+            .end().done().getBytes(StandardCharsets.UTF_8);
```
wb9688, line 361: Please don't do this, it will break NewPipe Legacy unfortunately

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,228 @@
...
+    @Nonnull
+    @Override
+    public InfoItemsPage<StreamInfoItem> getInitialPage() throws ExtractionException {
+        final StreamInfoItemsCollector collector = new StreamInfoItemsCollector(getServiceId());
+        collectStreamsFrom(collector, playlistData.getArray("contents"));
+        return new InfoItemsPage<>(collector,
+                new Page(getNextPageUrl(), Collections.singletonMap(COOKIE_NAME, cookieValue)));
+    }
+
+    private String getNextPageUrl() throws ExtractionException {
```
wb9688, line 130: This function is unnecessary imho

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubePlaylistLinkHandlerFactory.java
```diff
@@ -67,4 +74,33 @@ public boolean onAcceptUrl(final String url) {
...
+     */
+    @Override
+    public ListLinkHandler fromUrl(final String url) throws ParsingException {
+        try {
+            final URL urlObj = Utils.stringToURL(url);
+            final String listID = Utils.getQueryValue(urlObj, "list");
+            if (listID != null && YoutubeParsingHelper.isYoutubeMixId(listID)) {
+                String videoID = Utils.getQueryValue(urlObj, "v");
+                if (videoID == null) {
+                    videoID = listID.substring(2);
```
wb9688, line 92: This doesn't work properly for the Mix types having a prefix longer than 2 letters.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -0,0 +1,228 @@
...
+
+        return getUrlFromNavigationEndpoint(
+                lastStream.getObject("playlistPanelVideoRenderer").getObject("navigationEndpoint"))
+                + "&pbj=1";
+    }
+
+    @Override
+    public InfoItemsPage<StreamInfoItem> getPage(final Page page)
+            throws ExtractionException, IOException {
+        if (page == null || page.getUrl().isEmpty()) {
```
wb9688, line 149: It should require the cookie.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeMixPlaylistExtractorTest.java
```diff
@@ -0,0 +1,341 @@
...
+        @Test
+        public void getInitialPage() throws Exception {
+            final InfoItemsPage<StreamInfoItem> streams = extractor.getInitialPage();
+            assertFalse(streams.getItems().isEmpty());
+            assertTrue(streams.hasNextPage());
+        }
+
+        @Test
+        public void getPage() throws Exception {
+            final InfoItemsPage<StreamInfoItem> streams = extractor.getPage(
```
wb9688, line 85: When requiring the cookie, this won't work anymore.

wb9688 (changes requested):
I found these issues when reviewing the code. I haven't tested it in-app yet.

XiangRongLin:
@wb9688 I applied your suggestions.

I also noticed another edge case with your note that my extraction of the videoId does not work for prefixes longer than 2.
For channel mixes the playlist id does not contain the video id, it is the channel id instead.
So now the `PlaylistLinkHandlerFactory` rejects urls like these

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
XiangRongLin: wow, the cleanup feature on commit from Android Studio changed that to StandardCharsets when i changed it to the String.

TobiGr (approved):
Code looks good to me.
Thank you for the patience!
Can you fix the conflicts, please?

XiangRongLin:
@TobiGr It seems like that this is already so old, that YouTube has already changed their json format. So I'll have to look into it again.

On another note, NewPipe want to be very privacy friendly and (I think) avoid any way to allow tracking. Does this feature fit in there? Because it uses cookies to allow for continuations.

TheAssassin:
> On another note, NewPipe want to be very privacy friendly and (I think) avoid any way to allow tracking. Does this feature fit in there? Because it uses cookies to allow for continuations.

That's quite a good point. You'd have to make sure to limit tracking _really_ well (ideally store different cookies for different mixes, etc.). It allows for tracking across very long periods of times, too. Sometimes a mix may run for hours.

We should discuss that in the team, I guess.

XiangRongLin:
> You'd have to make sure to limit tracking really well (ideally store different cookies for different mixes, etc.). 

If i remember correctly, YouTube provides the cookie and i just send it back on continuations request. For a mix to not provide duplicate videos, it is needed for as long as the mix is running

Stypox:
> If i remember correctly, YouTube provides the cookie and i just send it back on continuations request. For a mix to not provide duplicate videos, it is needed for as long as the mix is running

In my opinion this is ok, as it only tells YouTube that a random user seems to be really listening the mix, which YouTube would already know by looking at the video details request.

Stypox:
> If i remember correctly, YouTube provides the cookie and i just send it back on continuations request. For a mix to not provide duplicate videos, it is needed for as long as the mix is running

In my opinion this is ok, as it only tells YouTube that a random user seems to be really listening the mix, which YouTube would already know by looking at the video details request.
@XiangRongLin sorry for not merging this for a long long time. If you fix youtube mixes again I promise I'm going to review and finally merge in a short time (and ping me multiple times if I don't ;-) )

XiangRongLin:
@Stypox Problem was smaller than i expected, see the last commit of the bunch

As for the rebase, only conflicting file was `YoutubeParsingHelper, where i think i got the correct stuff merged together.

XiangRongLin:
apk at the top is already updated

Stypox:
I tested it with a couple mixes, and also with continuations, and it works as expected. Thank you! Sorry for taking so long ;-)

PR #232, opened 2020-01-03
Bandcamp support

fynngodau:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I did test the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).
- [x] Icons need to be added to NewPipe for `Featured` and `Radio` and for Bandcamp itself. TeamNewPipe/NewPipe#3741

---
Fixes #138
Fixes TeamNewPipe/NewPipe#1284

<del>I had to add the `org.json` library because the website contains some JSON that has unquoted keys.</del>

- [x] Search
- [x] Multiple search pages
- [x] Search suggestions
- [x] Streams
 <del>Descriptions are suboptimal because their newlines get eaten and then lyrics are really difficult to read</del> Not anymore for some reason.
- [ ] Channels
 Actually implemented, but depends on #227
- [x] Playlists work
 <del>Playlists with a lot of streams take ages to load, because for every playlist info item, a separate request needs to be made to find out the URL of its cover. I'll see whether I can add a workaround just for playlists with more than a number (e.g. 10) of streams</del> Workaround ba967f1a1520f7fadded3b6bfd67c4da0f188c2b only loads cover art of each track individually for playlists <= 10 streams
- Kiosks
   - [x] `Featured` kiosk
   Currently only loads nine items currently can't load further "pages"; can contain preorder albums whose cover is not always displayed correctly and which shows errors but works once opened
   - [x] Bandcamp weekly (`Radio`) kiosk
    Clicking uploader's (primary host's) name causes the crash reporter to open as it is a link to a fan page
- [ ] Fan pages
 Are extra work because they look different and don't really provide a lot of value I think
- [ ] Comments
 (`fanthanks` / `writings`)
- [x] Unit tests
 But I'm not sure of their quality

PR #285, opened 2020-03-12
Fix Crash Related To New Alerts Check With Youtube-Music Links

PeterHindes:

Fixes TeamNewPipe/NewPipe#3218
Needs review from the [creator of the modified snippet](https://github.com/TeamNewPipe/NewPipeExtractor/commit/e65333c3cea86b6fa8197044b8969b21df47291e)

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java
@@ -495,8 +495,7 @@ public class YoutubeParsingHelper {
                 if (alertRendererHolder.has("alertRenderer")){
                     final JsonObject alertRenderer = alertRendererHolder.getObject("alertRenderer");
 
-                    String alertText = "Failed To Generate Error Text";
-                    String alertType = "Failed To Generate Error Type";
+                    String alertText, alertType;
                     try {
                         alertText = alertRenderer.getObject("text").getString("simpleText");
                         alertType = alertRenderer.getString("type");
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java
@@ -488,29 +488,20 @@ public class YoutubeParsingHelper {
         /* Should ignore this alert type
         Get high-quality audio, gapless playback and personalised music recommendations.
          */
-        if (initialData.has("alerts")) {
-            final JsonArray alerts = initialData.getArray("alerts");
-            if (!alerts.isEmpty() && alerts.has(0)) {
-                final JsonObject alertRendererHolder = alerts.getObject(0);
-                if (alertRendererHolder.has("alertRenderer")){
-                    final JsonObject alertRenderer = alertRendererHolder.getObject("alertRenderer");
+        
+        try {
+            final JsonObject alertRenderer = initialData.getArray("alerts").getObject(0).getObject("alertRenderer");
+            final String alertText = alertRenderer.getObject("text").getString("simpleText");
+            final String alertType = alertRenderer.getString("type");
 
-                    String alertText, alertType;
-                    try {
-                        alertText = alertRenderer.getObject("text").getString("simpleText");
-                        alertType = alertRenderer.getString("type");
-                    } catch (Exception ignored) {
-                        alertText = null;
-                        alertType = null;
-                    }
-                    if (alertType.equalsIgnoreCase("ERROR")) {
-                        if (alertText == null || alertText.isEmpty())
-                            throw new ContentNotAvailableException("Got unknown alert error");
-                        else
-                            throw new ContentNotAvailableException("Got alert error: \"" + alertText + "\"");
-                    }
-                }
+            if (alertType.equalsIgnoreCase("ERROR")) {
+                if (alertText == null || alertText.isEmpty())
+                    throw new ContentNotAvailableException("Got unknown alert error");
+                else
+                    throw new ContentNotAvailableException("Got alert error: \"" + alertText + "\"");
             }
-        }
+
+        } catch (Exception ignored) {}
+
     }
 }
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java
@@ -497,21 +497,19 @@ public class YoutubeParsingHelper {
          */
 
         try {
-            //System.out.println("Getting alert renderer");
             final JsonObject alertRenderer = initialData.getArray("alerts").getObject(0).getObject("alertRenderer"); // Exits here if No Alerts
-            //System.out.println("Getting alert type");
-            if (alertRenderer.getString("type").equalsIgnoreCase("ERROR")) { // Often exits here With Youtube-Music // Exits here if no errors
+            if (alertRenderer.getString("type").equalsIgnoreCase("ERROR")) { // Often exits here With Youtube-Music // Exits here if alert has no errors
                 try {
                     throw new ContentNotAvailableException("Got alert error: \"" + alertRenderer.getObject("text").getString("simpleText") + "\""); // Finishes here if errors have text
                 } catch (Exception ignored) {
                     throw new ContentNotAvailableException("Got unknown alert error"); // Finishes here if errors have no text
                 }
             } else {
-                //System.out.println("Non Error Alert Caught, and ignored.");
+                // This condition will only be reached if we encounter an alert that has a type, but is not an error
             }
 
         } catch (Exception ignored) {
-            //System.out.println("Failed to parse alert.");
+            // We will frequently reach this condition when alerts are found with no type (eg. youtube music ad text)
         }
 
     }
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java
@@ -499,7 +499,10 @@ public class YoutubeParsingHelper {
                     try {
                         alertText = alertRenderer.getObject("text").getString("simpleText");
                         alertType = alertRenderer.getString("type");
-                    } catch (Exception ignored) {}
+                    } catch (Exception ignored) {
+                        alertText = null;
+                        alertType = null;
+                    }
                     if (alertType.equalsIgnoreCase("ERROR")) {
                         if (alertText == null || alertText.isEmpty())
                             throw new ContentNotAvailableException("Got unknown alert error");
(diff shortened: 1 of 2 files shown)


Stypox:
You can tell GitHub to automatically close the issue from the other repo when this pr is merged by adding "Fixes TeamNewPipe/NewPipe#3218". I edited your comment ;-)

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java
```diff
@@ -393,13 +393,26 @@ public static JsonArray getJsonResponse(String url, Localization localization) t
...
+         */
+        if (initialData.has("alerts")) {
+            final JsonArray alerts = initialData.getArray("alerts");
+            if (!alerts.isEmpty() && alerts.has(0)) {
+                final JsonObject alertRendererHolder = alerts.getObject(0);
+                if (alertRendererHolder.has("alertRenderer")){
+                    final JsonObject alertRenderer = alertRendererHolder.getObject("alertRenderer");
+
+                    String alertText = "Failed To Generate Error Text";
+                    String alertType = "Failed To Generate Error Type";
```
Stypox: There is no need to set these strings to some value, if that value is unused or overwritten.
```suggestion
                    String alertText, alertType;
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java
```diff
@@ -393,13 +393,26 @@ public static JsonArray getJsonResponse(String url, Localization localization) t
...
+                    try {
+                        alertText = alertRenderer.getObject("text").getString("simpleText");
+                        alertType = alertRenderer.getString("type");
+                    } catch (Exception ignored) {}
+                    if (alertType.equalsIgnoreCase("ERROR")) {
```
Stypox:
```suggestion
                    if (alertType != null && alertType.equalsIgnoreCase("ERROR")) {
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java
```diff
@@ -393,13 +393,26 @@ public static JsonArray getJsonResponse(String url, Localization localization) t
...
+                        throw new ContentNotAvailableException("Got error: \"" + alertText + "\"");
```
Stypox: Maybe you should also check here if `alertText` is null. Also note I slightly changed the thrown message (to provide more information on the source).
```suggestion
                        if (alertText == null || alertText.isEmpty())
                            throw new ContentNotAvailableException("Got unknown alert error");
                        else
                            throw new ContentNotAvailableException("Got alert error: \"" + alertText + "\"");
```

PeterHindes:
Hey real quick, is this `    implementation 'com.github.PeterHindes:NewPipeExtractor:yt-music-alert-fix:c852d18'` how you include a branch that is not main from github in gradle?

TobiGr:
You don't need the branch name. The commit hash is enough

PeterHindes:
I ask because I am getting this error `ERROR: Failed to resolve: com.github.PeterHindes:NewPipeExtractor:c852d18`
with it configured normally

PeterHindes:
It seems to work if I use a snapshot from before accepting the changes on github. Weird git issues.

PeterHindes:
`Hey @Stypox is this error important

wb9688:
Yes, an error is an error, meaning it won't even compile.

PeterHindes:
Turns out that was both of my problems. Good I caught that before merge.

PeterHindes:
Hey out of curiosity what does the alert code normally do? What conditions would it be triggered by?

PeterHindes:
Needs another code review. I refactored it to use less memory.

PeterHindes:
Sorry about the extra commits. I couldn't find a way to build NewPipe with the local code instead of a GitHub repo.

PeterHindes:
Sure thing. Just glad to be free of that error while playing youtube music.
Any idea why youtube music throws an alert randomly with advertising text?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java
```diff
@@ -393,14 +393,32 @@ public static JsonArray getJsonResponse(String url, Localization localization) t
...
+                try {
+                    throw new ContentNotAvailableException("Got alert error: \"" + alertRenderer.getObject("text").getString("simpleText") + "\""); // Finishes here if errors have text
+                } catch (Exception ignored) {
+                    throw new ContentNotAvailableException("Got unknown alert error"); // Finishes here if errors have no text
+                }
+            } else {
+                // This condition will only be reached if we encounter an alert that has a type, but is not an error
             }
+
+        } catch (Exception ignored) {
```
TobiGr: This will ignore all exceptions including a ContentNotAvailableException thrown above

PeterHindes: What should happen? Should we throw an error?

PeterHindes: Give an example of when that type of error would be thrown.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java
```diff
@@ -393,14 +393,32 @@ public static JsonArray getJsonResponse(String url, Localization localization) t
...
+            // We will frequently reach this condition when alerts are found with no type (eg. youtube music ad text)
```
TobiGr: I didn't test this, but it should work. Please test it before committing.
```suggestion
        } catch (Exception e) {
            if (e instanceof ContentNotAvailableException) {
                throw e;
            }
            // Ignore all other exceptions
            // We will frequently reach this condition when alerts are found with no type (eg. YouTube Music ad text)
```

TobiGr: This is cleaner than my first suggestion.
```suggestion
        } catch (ContentNotAvailableException e) {
            throw e;
        } catch (Exception ignored) {
            // We will frequently reach this condition when alerts are found with no type (e.g. YouTube Music ad text)
```

PeterHindes:
Ok, Ill try that fix in a bit.

PeterHindes:
Those changes seem to work. Thanks @TobiGr

PeterHindes:
Can you quickly explain what `ContentNotAvailableException` means? Under what condition would we get that.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java
```diff
@@ -398,14 +398,34 @@ public static JsonArray getJsonResponse(String url, Localization localization) t
      * @throws ContentNotAvailableException if an alert is detected
      */
     public static void defaultAlertsCheck(JsonObject initialData) throws ContentNotAvailableException {
```
mauriciocolli, line 400: I suggest to use the newly introduced default values, that way, we can avoid all those nasty try-catch (and null pointer exceptions).

###### Maybe we should switch to a new library that does this by default, or even fork and modify the one we are using.

```java
final JsonObject alertRenderer = initialData
        .getArray("alerts", EMPTY_ARRAY)
        .getObject(0, EMPTY_OBJECT)
        .getObject("alertRenderer", EMPTY_OBJECT);

final String alertType = alertRenderer.getString("type", EMPTY_STRING);
if (alertType.equalsIgnoreCase("ERROR")) {
    final String alertText = alertRenderer
            .getObject("text", EMPTY_OBJECT)
            .getString("simpleText", EMPTY_STRING);

    if (alertText.isEmpty()) {
        throw new ContentNotAvailableException("Got an unknown alert error");
    } else {
        throw new ContentNotAvailableException("Got alert error: \"" + alertText + "\"");
    }
}
```

mauriciocolli:
@PeterHindes 

> I couldn't find a way to build NewPipe with the local code instead of a GitHub repo.

You can find one way to do it in the README, [check it out](https://github.com/TeamNewPipe/NewPipeExtractor#testing-changes).

Another way would be to use JitPack, [check out their docs](https://jitpack.io/docs/#building-with-jitpack). A disadvantage is that you would have to push every little change to your repository, wasting network and time waiting for it to build in the process. Personally, I think it's much more sane to use the local repository approach for testing changes.

> what does the alert code normally do? What conditions would it be triggered by?

In case of a error, the reason will be there. Try going to a [non-existent playlist](https://www.youtube.com/playlist?list=PL11111111111111111111111111111111).

> Any idea why youtube music throws an alert randomly with advertising text?

I suppose they wanted to generalize that alert message element, because it is used in errors as well.

> Can you quickly explain what ContentNotAvailableException means? Under what condition would we get that.

It is a general exception representing exactly what the name implies: the content was not available, the reasons can vary, like: deleted, not found, unavailable, etc.

As a side note, it is planned to have more detailed exceptions of this type in the future, representing the reasons I mentioned above so we can tell the exact reason to the user.

wb9688:
>Maybe we should switch to a new library that does this by default, or even fork and modify the one we are using.

@mauriciocolli: We're already going to fork nanojson, so why not just add that as well?

>You can find one way to do it in the README, check it out.

A better way imho is to add something like the following to `settings.gradle`:

```groovy
includeBuild('../NewPipeExtractor') {
    dependencySubstitution {
        substitute module('com.github.TeamNewPipe:NewPipeExtractor') with project(':extractor')
    }
}
```

mauriciocolli:
@wb9688 

> We're already going to fork nanojson, so why not just add that as well?

Oh, I didn't know, would be a good addition.

> A better way imho is to add something like the following to settings.gradle:

I remember seeing that in a PR, is that [from here](https://github.com/TeamNewPipe/NewPipe/pull/2535#issue-308341438)? I think I've used it before as well.

It is indeed more practical than what is being suggested right now, would be nice to update that section in the README.

wb9688:
>Oh, I didn't know, would be a good addition.

@mauriciocolli: Yup, see #232 for the reason why.

>I remember seeing that in a PR, is that from here? I think I've used it before as well.
>
>It is indeed more practical than what is being suggested right now, would be nice to update that section in the README.

Yup, that's where I got it from. We should indeed add that to the `README.md`.

PeterHindes:
So what should I do with this pull request going forward.

wb9688:
@PeterHindes: Use the default values, rebase to make one commit of it, and force push it.

PeterHindes:
just a quick question. Couldn't we avoid most of these issues by stripping the `music.` part of `music.youtube.com` from incoming urls opened with the app.

PeterHindes:
I'm having trouble building.

PeterHindes:
@wb9688  The 
`includeBuild('../NewPipeExtractor') {
    dependencySubstitution {
        substitute module('com.github.TeamNewPipe:NewPipeExtractor') with project(':extractor')
    }
}`

method results in errors

wb9688:
Closing in favor of #323

PR #287, opened 2020-03-14
[YouTube] Fix channel with redirects directly in the response

### LINKED ISSUE #283: ParsingException: Could not get channel id
There are some channels where I get this exception.

There are regular channels where id has changed and page is redirecting to new id:
https://www.youtube.com/channel/UCZo_3DUateypfhV_fytbwWA
https://www.youtube.com/channel/UCITk7Ky4iE5_xISw9IaHqpQ

Same behavior, but different ids redirecting to same Youtube's "Movies & Shows" channel:
https://www.youtube.com/channel/UCuJcl0Ju-gPDoksRjK1ya-w
https://www.youtube.com/channel/UChBfWrfBXL9wS6tQtgjt_OQ
https://www.youtube.com/channel/UCok7UTQQEP1Rsctxiv3gwSQ

Some channels where id hasn't changed but still getting same exception:
https://www.youtube.com/channel/UC0KU8F9jJqSLS11LRXvFWmg
https://www.youtube.com/channel/UCHcMtE3EiKmqrJZBZyUBuRQ

mauriciocolli:
Some redirects were embed directly into the response as instructions for the page, instead of the usual http redirects.

- Fix channel extraction when redirects are in the response.
- Fix id extraction for some channels.
    - Some channels had no reliable way to get the redirected id in the response.

There's still some problems though, channels generated by YouTube won't have video tabs and in some cases, even the `subscriberCountText` has completely unrelated text in it  (e.g. [Movies & Shows](https://www.youtube.com/channel/UClgRkhTL3_hImCAmdLfDE4g)).

@wb9688 Maybe we have to extract the essentials and report that it has no videos (technically the truth), or even ignore those generated channels completely? I guess the essentials are better than nothing (like the old extractor did).

---

Fixes #283.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
@@ -46,6 +46,18 @@ public class YoutubeChannelExtractor extends ChannelExtractor {
     private JsonObject initialData;
     private JsonObject videoTab;
 
+    /**
+     * Some channels have response redirects and the only way to reliably get the id is by saving it.
+     *<p>
+     * "Movies & Shows":
+     * <pre>
+     * UCuJcl0Ju-gPDoksRjK1ya-w ┐
+     * UChBfWrfBXL9wS6tQtgjt_OQ ├ UClgRkhTL3_hImCAmdLfDE4g
+     * UCok7UTQQEP1Rsctxiv3gwSQ ┘
+     * </pre>
+     */
+    private String redirectedChannelId;
+
     public YoutubeChannelExtractor(StreamingService service, ListLinkHandler linkHandler) {
         super(service, linkHandler);
     }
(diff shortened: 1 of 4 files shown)


wb9688:
@mauriciocolli: What will happen in my tabs PR is that it will extract the name and stuff, and it'll always display both the Videos and the Playlists tab, both with just 'No items' in it in that case. The displayed tabs will be changed in a future PR. That Movies & Shows channel is completely worthless for us though.

B0pol:
> @mauriciocolli: What will happen in my tabs PR is that it will extract the name and stuff, and it'll always display both the Videos and the Playlists tab, both with just 'No items' in it in that case. The displayed tabs will be changed in a future PR. That Movies & Shows channel is completely worthless for us though.

I agree with Films and TV (which looks like not even a channel but a shop), but that's not true for all YouTube-generated channels. I think Music, News, Live, Sport, Video Games would interest lots of people (just look at the subcount of these channels, it's very high).

wb9688:
@B0pol: I agree, but as I said, that'll be for a later PR.

Stypox:
> I guess the essentials are better than nothing (like the old extractor did).

I agree

PR #288, opened 2020-03-14
[YouTube] Fix videos without next stream throwing error and add test for unlisted videos

Stypox:
- [ x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

(*from discussion on IRC*) In my opinion getNextStream should be removed from the extractor and the next video should be just prepended, when available, to the suggested videos. The app would then consider the first stream of the suggested videos as the next one. This imo would solve many issues, such as those complaining some videos have no next stream, and remove unneded complexity from the extractor.

But for now I went with the `return null` option, which is consistent with what other services do.
I also added a test for unlisted videos (apart from the "next video" function there should be no differences, but having one more test never hurt anyone ;-) ) 

Fixes TeamNewPipe/NewPipe#3226
Closes TeamNewPipe/NewPipe#2658

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
@@ -523,10 +523,18 @@ public class YoutubeStreamExtractor extends StreamExtractor {
         if (getAgeLimit() != NO_AGE_LIMIT) return null;
 
         try {
-            final JsonObject videoInfo = initialData.getObject("contents").getObject("twoColumnWatchNextResults")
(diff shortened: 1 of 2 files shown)


wb9688:
>(from discussion on IRC) In my opinion getNextStream should be removed from the extractor and the next video should be just prepended, when available, to the suggested videos. The app would then consider the first stream of the suggested videos as the next one. This imo would solve many issues, such as those complaining some videos have no next stream, and remove unneded complexity from the extractor.

That's what I've been thinking as well, plus e.g. on SoundCloud autoplay doesn't work as it doesn't have a next stream, just related streams.

TobiGr:
I agree with that, too.

PR #291, opened 2020-03-17
Add support for YouTube Music search

wb9688:

I have to do a bit of cleaning up, make `getYoutubeMusicKeys()` more reliable and add tests. The problem with adding tests is that `DownloaderTestImpl` somehow behaves weirdly, which causes it to not work. The code works fine with OkHttp on Android in NewPipe.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicSearchExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicSearchExtractor.java
@@ -9,8 +9,6 @@ import com.grack.nanojson.JsonWriter;
 import org.schabi.newpipe.extractor.InfoItem;
 import org.schabi.newpipe.extractor.StreamingService;
 import org.schabi.newpipe.extractor.downloader.Downloader;
-import org.schabi.newpipe.extractor.downloader.Response;
-import org.schabi.newpipe.extractor.exceptions.ContentNotAvailableException;
 import org.schabi.newpipe.extractor.exceptions.ExtractionException;
 import org.schabi.newpipe.extractor.exceptions.ParsingException;
 import org.schabi.newpipe.extractor.exceptions.ReCaptchaException;
@@ -31,6 +29,7 @@ import java.util.Map;
 import javax.annotation.Nonnull;
 
 import static org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeParsingHelper.fixThumbnailUrl;
+import static org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeParsingHelper.getValidResponseBody;
 import static org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeParsingHelper.getTextFromObject;
 import static org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeParsingHelper.getUrlFromNavigationEndpoint;
 import static org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeSearchQueryHandlerFactory.MUSIC_ALBUMS;
@@ -112,23 +111,7 @@ public class YoutubeMusicSearchExtractor extends SearchExtractor {
         headers.put("Referer", Collections.singletonList("music.youtube.com"));
         headers.put("Content-Type", Collections.singletonList("application/json"));
 
-        final Response response = getDownloader().post(url, headers, json);
-
-        if (response.responseCode() == 404) {
-            throw new ContentNotAvailableException("Not found" +
-                    " (\"" + response.responseCode() + " " + response.responseMessage() + "\")");
-        }
-
-        final String responseBody = response.responseBody();
-        if (responseBody.length() < 50) { // ensure to have a valid response
-            throw new ParsingException("JSON response is too short");
-        }
-
-        final String responseContentType = response.getHeader("Content-Type");
-        if (responseContentType != null && responseContentType.toLowerCase().contains("text/html")) {
-            throw new ParsingException("Got HTML document, expected JSON response" +
-                    " (latest url was: \"" + response.latestUrl() + "\")");
-        }
+        final String responseBody = getValidResponseBody(getDownloader().post(url, headers, json));
 
         try {
             initialData = JsonParser.object().from(responseBody);
@@ -232,23 +215,7 @@ public class YoutubeMusicSearchExtractor extends SearchExtractor {
         headers.put("Referer", Collections.singletonList("music.youtube.com"));
         headers.put("Content-Type", Collections.singletonList("application/json"));
 
-        final Response response = getDownloader().post(pageUrl, headers, json);
-
-        if (response.responseCode() == 404) {
-            throw new ContentNotAvailableException("Not found" +
-                    " (\"" + response.responseCode() + " " + response.responseMessage() + "\")");
-        }
-
-        final String responseBody = response.responseBody();
-        if (responseBody.length() < 50) { // ensure to have a valid response
-            throw new ParsingException("JSON response is too short");
-        }
-
-        final String responseContentType = response.getHeader("Content-Type");
-        if (responseContentType != null && responseContentType.toLowerCase().contains("text/html")) {
-            throw new ParsingException("Got HTML document, expected JSON response" +
-                    " (latest url was: \"" + response.latestUrl() + "\")");
-        }
+        final String responseBody = getValidResponseBody(getDownloader().post(pageUrl, headers, json));
 
         final JsonObject ajaxJson;
         try {
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicSearchExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicSearchExtractor.java
@@ -319,31 +319,32 @@ public class YoutubeMusicSearchExtractor extends SearchExtractor {
 
                         @Override
                         public String getUploaderUrl() throws ParsingException {
-                            String url = null;
-
                             if (searchType.equals(MUSIC_VIDEOS)) {
                                 JsonArray items = info.getObject("menu").getObject("menuRenderer").getArray("items");
                                 for (Object item : items) {
                                     final JsonObject menuNavigationItemRenderer = ((JsonObject) item).getObject("menuNavigationItemRenderer");
                                     if (menuNavigationItemRenderer != null && menuNavigationItemRenderer.getObject("icon").getString("iconType").equals("ARTIST")) {
-                                        url = getUrlFromNavigationEndpoint(menuNavigationItemRenderer.getObject("navigationEndpoint"));
-                                        break;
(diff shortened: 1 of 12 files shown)


B0pol:
Already talked about in irc, but channel are unusuable from youtube music search : no streams, no subscribers count, there are only banner and profile picture

wb9688:
This PR is 100% finished now.

@TobiGr: Could this be included in v0.19.0? It's not really a big change.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSearchExtractor.java
```diff
@@ -49,52 +71,169 @@ public YoutubeSearchExtractor(StreamingService service, SearchQueryHandler linkH
...
+                    .end()
+                    .value("query", getSearchString())
+                    .value("params", params)
+                .end().done().getBytes("UTF-8");
+            // @formatter:on
+
+            Map<String, List<String>> headers = new HashMap<>();
+            headers.put("X-YouTube-Client-Name", Collections.singletonList(youtubeMusicKeys[1]));
+            headers.put("X-YouTube-Client-Version", Collections.singletonList(youtubeMusicKeys[2]));
+            headers.put("Origin", Collections.singletonList("https://music.youtube.com"));
```
mauriciocolli: As discussed, the default http client from java [has some restrictions about request headers](https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/sun/net/www/protocol/http/HttpURLConnection.java#l194).

I think using another client for tests is the right choice (like you did earlier, OkHttp is a good one).

Should we do it in its own pull request? The front end is already using it, don't think too many problems would be a result of this, but then we can test thoroughly what kind of effects it will have.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSearchExtractor.java
```diff
@@ -133,15 +352,198 @@ private void collectStreamsFrom(InfoItemsSearchCollector collector, JsonArray vi
...
+                            String viewCount = getTextFromObject(info.getArray("flexColumns").getObject(2)
+                                    .getObject("musicResponsiveListItemFlexColumnRenderer").getObject("text"));
+                            if (viewCount != null && !viewCount.isEmpty()) return Utils.mixedNumberWordToLong(viewCount);
+                            throw new ParsingException("Could not get view count");
+                        }
+
+                        @Override
+                        public String getThumbnailUrl() throws ParsingException {
+                            try {
+                                // TODO: Don't simply get the first item, but look at all thumbnails and their resolution
```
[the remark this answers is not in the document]
wb9688: I already discussed this with @B0pol, but forgot about it. I'll fix it in a minute.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSearchExtractor.java
wb9688: This should indeed be a different PR. I think OkHttp should be used, so we have the exact same behavior in the tests as in NewPipe.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/search/YoutubeSearchExtractorMusicTest.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import static java.util.Arrays.asList;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.schabi.newpipe.extractor.ServiceList.YouTube;
+
+public class YoutubeSearchExtractorMusicTest extends YoutubeSearchExtractorBaseTest {
```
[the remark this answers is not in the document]
wb9688: True, however we appear to not use the `DefaultTests` at all for the SearchExtractor.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSearchExtractor.java
```diff
@@ -133,15 +353,217 @@ private void collectStreamsFrom(InfoItemsSearchCollector collector, JsonArray vi
...
+                        }
+
+                        @Override
+                        public String getUploaderUrl() throws ParsingException {
+                            if (searchType.equals(MUSIC_VIDEOS)) return null;
+                            String url = getUrlFromNavigationEndpoint(info.getArray("flexColumns")
+                                    .getObject(1).getObject("musicResponsiveListItemFlexColumnRenderer")
+                                    .getObject("text").getArray("runs").getObject(0).getObject("navigationEndpoint"));
+                            if (url != null && !url.isEmpty()) return url;
+                            throw new ParsingException("Could not get uploader url");
```
mauriciocolli: Null pointer exceptions when the uploader url is not available, a null should be returned in those cases.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSearchExtractor.java
```diff
@@ -133,15 +353,217 @@ private void collectStreamsFrom(InfoItemsSearchCollector collector, JsonArray vi
...
+                        public String getUploaderName() throws ParsingException {
+                            String name = getTextFromObject(info.getArray("flexColumns").getObject(1)
+                                    .getObject("musicResponsiveListItemFlexColumnRenderer").getObject("text"));
+                            if (name != null && !name.isEmpty()) return name;
+                            throw new ParsingException("Could not get uploader name");
+                        }
+
+                        @Override
+                        public String getUploaderUrl() throws ParsingException {
+                            if (searchType.equals(MUSIC_VIDEOS)) return null;
```
mauriciocolli: It is available in some video items which have the "Go to artist" button in the item menu, though it is usually for the "topic" channel instead, maybe we should ignore them anyway?

We don't even use this uploader url for anything right now.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSearchExtractor.java
wb9688: Could you tell me where you've seen that happen? I've never seen that happen, and I don't think @opusforlife2 has seen that happen either.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSearchExtractor.java
wb9688: I'll add support for them, though topic channels will only be useful once I made my follow-up tabs PR.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSearchExtractor.java
opusforlife2: @mauriciocolli Can you provide a link?

wb9688: I found one. Thanks!

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/search/YoutubeMusicSearchExtractorTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeSearchQueryHandlerFactory;
+
+import java.net.URLEncoder;
+
+import javax.annotation.Nullable;
+
+import static java.util.Collections.singletonList;
+import static org.schabi.newpipe.extractor.ServiceList.YouTube;
+
+public class YoutubeMusicSearchExtractorTest {
```
mauriciocolli, line 19: This is not testing all available filters for YouTube Music, it should test all of them.

wb9688:
>In a second look, I think it makes sense to create a separate class specifically for the music version, as you already did in the tests.

@mauriciocolli: Like we're doing with the mixes? I agree and I'll fix that today.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicSearchExtractor.java
```diff
@@ -0,0 +1,527 @@
...
+        final String responseBody = response.responseBody();
+        if (responseBody.length() < 50) { // ensure to have a valid response
+            throw new ParsingException("JSON response is too short");
+        }
+
+        final String responseContentType = response.getHeader("Content-Type");
+        if (responseContentType != null && responseContentType.toLowerCase().contains("text/html")) {
+            throw new ParsingException("Got HTML document, expected JSON response" +
+                    " (latest url was: \"" + response.latestUrl() + "\")");
+        }
```
mauriciocolli, line 131: The only thing now are these duplicated checks from `YoutubeParsingHelper`, maybe we could extract them to yet another general response check in there?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicSearchExtractor.java
```diff
@@ -0,0 +1,527 @@
...
```
mauriciocolli, line 251: And here.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicSearchExtractor.java
wb9688: Yeah, I already thought it wasn't that clean.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java
```diff
@@ -351,12 +428,8 @@ public static String fixThumbnailUrl(String thumbnailUrl) {
         return thumbnailUrl;
     }
 
-    public static JsonArray getJsonResponse(String url, Localization localization) throws IOException, ExtractionException {
-        Map<String, List<String>> headers = new HashMap<>();
-        headers.put("X-YouTube-Client-Name", Collections.singletonList("1"));
-        headers.put("X-YouTube-Client-Version", Collections.singletonList(getClientVersion()));
-        final Response response = getDownloader().get(url, headers, localization);
-
+    public static String getValidResponseBody(final Response response)
```
mauriciocolli, line 431: Maybe `getValidJsonResponseBody` would be better to represent what is actually done. But meh, up to you to decide.

TobiGr: I agree

PR #292, opened 2020-03-17
Fix YouTube videos with no views

wb9688:

Fixes TeamNewPipe/NewPipe#3238

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
@@ -279,6 +279,8 @@ public class YoutubeStreamExtractor extends StreamExtractor {
             if (views == null) throw new ParsingException("Could not get view count");
         }
 
+        if (views.toLowerCase().contains("no views")) return 0;
+
         return Long.parseLong(Utils.removeNonDigitCharacters(views));
     }
 


Stypox:
@wb9688 what happens in other localizations? I would put a check on the result of `Utils.removeNonDigitCharacters(views)`: if it is an empty string I think it would be safe to return 0, as "No views" and hopefully all of its translations do not contain digits.

wb9688:
@Stypox: There are multiple places where we use localized strings, e.g. https://github.com/wb9688/NewPipeExtractor/blob/cb57123be199dd1633776e65020fae487692a1f5/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java#L243. @B0pol said he gave up on supporting localized YouTube sites though.

Stypox:
I know localized YouTube is not correctly supported, but since there is the opportunity not to introduce another dependency on English, let's not do it ;-)

gzsombor:
I've run into the same issue - independently - and fixed this way: https://github.com/TeamNewPipe/NewPipeExtractor/pull/294

wb9688:
Updated PR to use same approach as in YoutubeStreamInfoItemExtractor.

Stypox:
@TobiGr oh, I was about to do it, I already added the change to the changelog ;-)

TobiGr:
I added it, too. Maybe I have overridden it :laughing:

PR #293, opened 2020-03-17
[SoundCloud] Fix by migrating to `api-v2` and other improvements

### LINKED ISSUE #290: api.soundcloud.com can't be used anymore
On IRC we came to the conclusion that we can't use api.soundcloud.com anymore with SoundCloud's client ID, so we'll have to rewrite those parts of NewPipeExtractor to use api-v2.soundcloud.com.

Stypox:
- [ x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [ x] I did test the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).

Even though now `api` started to work again with the hardcoded id, it seems to be unstable, as yesterday it returned a "403 Forbidden". `api-v2` works better than `api`, anyway, since I was able to unignore some previously ignored tests. This PR migrates every Extractor to `api-v2`, along with some other fixes. 
Fixes #290

These are all the improvements:
- Use `api-v2` in PlaylistExtractor: rewrote methods to calculate next page url and to get items from it. `api-v2` is different from `api` since the initial playlist page contains (usually) the full info of the first 3 streams and only the id of the others. Then the single tracks can be requested in batch using `/tracks?ids=id1,id2,...`, so that's the next url we get.
- Improve thumbnail url extraction in PlaylistExtractor: prevent NullPointerExceptions and remove duplicate code.
- Use api-v2 in SubscriptionExtractor: changing the url was enough. Also added some tests for channel urls with an `api-v2` url.
- Improve playlists tests: assert playlists do not have a banner, unignore every ignored test (now they all succeed)

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudPlaylistExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudPlaylistExtractor.java
@@ -169,14 +169,14 @@ public class SoundcloudPlaylistExtractor extends PlaylistExtractor {
                 + "&ids=").length();
         final int lengthOfEveryStream = 11;
 
-        String currentPageUrl;
+        String currentPageUrl, nextUrl;
         int lengthMaxStreams = lengthFirstPartOfUrl + lengthOfEveryStream * streamsPerRequestedPage;
         if (pageUrl.length() <= lengthMaxStreams) {
             currentPageUrl = pageUrl; // fetch every remaining video, there are less than the max
-            nextPageUrl = ""; // afterwards the list is complete
+            nextUrl = ""; // afterwards the list is complete
         } else {
             currentPageUrl = pageUrl.substring(0, lengthMaxStreams);
-            nextPageUrl = pageUrl.substring(0, lengthFirstPartOfUrl) + pageUrl.substring(lengthMaxStreams);
+            nextUrl = pageUrl.substring(0, lengthFirstPartOfUrl) + pageUrl.substring(lengthMaxStreams);
         }
 
         StreamInfoItemsCollector collector = new StreamInfoItemsCollector(getServiceId());
@@ -193,6 +193,6 @@ public class SoundcloudPlaylistExtractor extends PlaylistExtractor {
             throw new ParsingException("Could not parse json response", e);
         }
 
-        return new InfoItemsPage<>(collector, nextPageUrl);
+        return new InfoItemsPage<>(collector, nextUrl);
     }
 }
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudPlaylistExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudPlaylistExtractor.java
@@ -30,8 +30,6 @@ public class SoundcloudPlaylistExtractor extends PlaylistExtractor {
     private JsonObject playlist;
 
     private StreamInfoItemsCollector streamInfoItemsCollector;
-    private List<Integer> nextTrackIds;
-    private int nextTrackIdsIndex;
     private String nextPageUrl;
 
     public SoundcloudPlaylistExtractor(StreamingService service, ListLinkHandler linkHandler) {
@@ -121,15 +119,16 @@ public class SoundcloudPlaylistExtractor extends PlaylistExtractor {
     @Override
     public InfoItemsPage<StreamInfoItem> getInitialPage() throws IOException, ExtractionException {
         if (streamInfoItemsCollector == null) {
-            computeInitialTracksAndNextIds();
+            computeInitialTracksAndNextPageUrl();
         }
-        return new InfoItemsPage<>(streamInfoItemsCollector, getNextPageUrl());
+        return new InfoItemsPage<>(streamInfoItemsCollector, nextPageUrl);
     }
 
-    private void computeInitialTracksAndNextIds() {
+    private void computeInitialTracksAndNextPageUrl() throws IOException, ExtractionException {
         streamInfoItemsCollector = new StreamInfoItemsCollector(getServiceId());
-        nextTrackIds = new ArrayList<>();
-        nextTrackIdsIndex = 0;
+        StringBuilder nextPageUrlBuilder = new StringBuilder("https://api-v2.soundcloud.com/tracks?client_id=");
+        nextPageUrlBuilder.append(SoundcloudParsingHelper.clientId());
+        nextPageUrlBuilder.append("&ids=");
 
         JsonArray tracks = playlist.getArray("tracks");
         for (Object o : tracks) {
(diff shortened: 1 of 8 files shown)


Stypox:
I FOUND AN ISSUE, DO NOT MERGE

Stypox:
Ok, this should be ready for merge. I tested thoroughly inside NewPipe and all is working correctly. SoundCloud tests are all succeeding :-D

extractor/src/test/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudPlaylistExtractorTest.java
```diff
@@ -229,6 +234,15 @@ public static void setUp() throws Exception {
...
             defaultTestGetPageInNewExtractor(extractor, newExtractor);
+            String page1 = newExtractor.getNextPageUrl();
+            defaultTestMoreItems(newExtractor); // there has to be another page
+            String page2 = newExtractor.getNextPageUrl();
+            defaultTestMoreItems(newExtractor); // and another one
+            String page3 = newExtractor.getNextPageUrl();
+
+            assertNotEquals("Same pages", page1, page2);
+            assertNotEquals("Same pages", page2, page3);
+            assertNotEquals("Same pages", page3, page1);
```
wb9688: What are you even doing here? getNextPageUrl should always return the same URL, as it's the next page relative to the initial page, see https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/main/java/org/schabi/newpipe/extractor/ListExtractor.java#L36.

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudPlaylistExtractor.java
```diff
@@ -64,21 +74,20 @@ public String getThumbnailUrl() {
             // if it also fails, return null
             try {
                 final InfoItemsPage<StreamInfoItem> infoItems = getInitialPage();
-                if (infoItems.getItems().isEmpty()) return null;
 
                 for (StreamInfoItem item : infoItems.getItems()) {
-                    final String thumbnailUrl = item.getThumbnailUrl();
-                    if (thumbnailUrl == null || thumbnailUrl.isEmpty()) continue;
-
-                    String thumbnailUrlBetterResolution = thumbnailUrl.replace("large.jpg", "crop.jpg");
```
wb9688: Why did you remove the code for better resolution? Does api-v2 return that by default?

extractor/src/test/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudPlaylistExtractorTest.java
```diff
@@ -70,7 +73,9 @@ public void testRelatedItems() throws Exception {
         }
 
         @Test
-        public void testMoreRelatedItems() {
+        public void testMoreRelatedItems() throws Exception {
+            defaultTestMoreItems(extractor);
```
wb9688: Why did you change this? This test was to check that this playlist indeed doesn't have a next page of items.

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudStreamExtractor.java
```diff
@@ -55,14 +57,14 @@ public String getName() {
 
     @Nonnull
     @Override
-    public String getTextualUploadDate() {
-        return track.getString("created_at");
+    public String getTextualUploadDate() throws ParsingException {
+        return track.getString("created_at").replace("T"," ").replace("Z", "");
```
wb9688: Why is this change needed? Also, I don't think those additional imports are needed here.

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudPlaylistExtractor.java
Stypox: I didn't remove it, I just moved it below

extractor/src/test/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudPlaylistExtractorTest.java
Stypox: Now the first page of items does not contain the 10 items, but only 3 of them

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudPlaylistExtractor.java
wb9688: Sorry for not noticing this, didn't I tell you I was tired?

extractor/src/test/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudPlaylistExtractorTest.java
Stypox: So why did the old extractor do the same thing as I do? Every time getPage was called it overwrote nextPageUrl with the url of the page coming after the one provided as the argument of getPage, thus having the same behaviour as now.

extractor/src/test/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudPlaylistExtractorTest.java
wb9688: Then the lines below should be deleted, however it would be better to change it to a playlist with less than 3 streams.

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudStreamExtractor.java
Stypox: For some reason SoundCloud api-v2 returns upload date in format "2016-07-31 18:18:07" instead of "2016/07/31T18:18:07Z+0000"

Stypox: For some reason SoundCloud api-v2 returns upload date in format "2016-07-31 18:18:07" instead of "2016/07/31T18:18:07Z+0000", so even though `getTextualUploadDate` should return the scraped date as a fallback, I converted it to a human readable one since it takes no effort and works reliably.

Stypox: Will remove the imports though

extractor/src/test/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudPlaylistExtractorTest.java
wb9688: I don't know, that would be a bug. YouTube doesn't do it: https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubePlaylistExtractor.java#L81, and YouTube is the 'golden' service implementation.

Stypox: Ok, will change this ;-)

extractor/src/test/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudPlaylistExtractorTest.java
Stypox: Ok, will do

Stypox:
Done

Stypox:
There is still an issue with search suggestions I missed earlier

Stypox:
I tested everything again everything works. Search suggestions, too.

Stypox:
@TobiGr @wb9688 this is ready. I tested thoroughly and everything works as expected, and an order of magnitude faster than Youtube ;-).
The only issue is that [some tracks](https://soundcloud.com/aviciiofficial/heaven-david-guetta-morten-2) do not have `mp3` streams and thus they cause a crash, but as discussed in IRC this is not fixable for now, and is also not related to these changes.

wb9688:
@TobiGr: Once this is merged, will you rebase dev on master?

PR #294, opened 2020-03-17
Catch a NFE in YoutubeStreamExtractor.getViewCount - for new videos, …

gzsombor:
…sometimes the viewCount is empty


wb9688:
This isn't the correct way to fix it as it will still throw an exception for no views instead of returning 0.

gzsombor:
Ok, fair point!

Issue #255, opened 2020-02-11
[Feature Request] Allow additional regexes to be passed to YouTubeStreamExtractor


CWftw:
See: https://github.com/0xced/XCDYouTubeKit/issues/463 for an example of how a similar project on iOS implemented this

Allowing NewPipeExtractor to accept additional regexes as a parameter would give NewPipe the ability to pull updated regexes from a server, file etc. to fix extraction errors without requiring an update - leading to reduced downtime.

PR #297, opened 2020-03-25
Pass additional regexes to youtube extractor

nik-vanderveen:

Adds support for passing additional regexes (from file, server, etc.) through youtube's extractor in order to fix extraction errors without an update. Closes TeamNewPipe/NewPipeExtractor#255

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeService.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeService.java
@@ -37,6 +37,8 @@ import org.schabi.newpipe.extractor.stream.StreamExtractor;
 import org.schabi.newpipe.extractor.subscription.SubscriptionExtractor;
 import org.schabi.newpipe.extractor.suggestion.SuggestionExtractor;
 
+import javax.annotation.Nonnull;
+import java.util.ArrayList;
 import java.util.List;
 
 import javax.annotation.Nonnull;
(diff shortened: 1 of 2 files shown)


TobiGr:
Thank you. Is there a reason why @nik-vanderveen creates a PR from @CWftw's repo?

CWftw:
> Thank you. Is there a reason why @nik-vanderveen creates a PR from @CWftw's repo?

I am CWftw and I approve this message

B0pol:
It has both a benefits and downsides:
:ok_hand: people will be able to add the regex found somewhere in settings
:+1: less people would leave NewPipe due to it being broken
:-1: people will be able to stay on an old version, without the other bug fixes…

I'm totally for this possibility.

opusforlife2:
Have you guys discussed how this should be presented to the user?

My instinct as a user is to basically have a tappable option in settings to download additional regexes which you could host on the Newpipe website. All it would do is present a toast whether the download succeeded or failed, and people should see their videos opening again.

Where the downloaded file goes is inconsequential to the end user, though it should probably be part of the app's internal storage so that even if a user doesn't grant storage permission to Newpipe, the functionality doesn't break.

litetex:
I'm closing this for now:
* no progress since over a year
* merge conflicts
* PR is in the current form no longer applicable. There have been massive reworks in the Extractor in the meantime (e.g. #604) and the suggested fixes with the regex e.g. won't solve #683

Feel free to reopen when there is progress again

PR #298, opened 2020-03-28
[peertube] add feed url in channels

B0pol:
- ~~I agree to ASAP create a PULL request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) for making in compatible when I changed the api.~~ no need

This will add the rss icon on peertube channels.
I haven't made feedextractor in this pr, because it's not faster (the original page is already small).


TobiGr:
Please change the tests ([here](https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/test/java/org/schabi/newpipe/extractor/services/peertube/PeertubeChannelExtractorTest.java#L101) and [here](https://github.com/TeamNewPipe/NewPipeExtractor/blob/647e7cd45054921712d6c6e592f39f70a1289e3b/extractor/src/test/java/org/schabi/newpipe/extractor/services/peertube/PeertubeChannelExtractorTest.java#L198)) accordingly,

Issue #250, opened 2020-01-24
[Soundcloud] Could not get any stream error on some tracks


B0pol:
Hi
Steps to reproduce:
Go in Soundcloud in NewPipe.
Search for « Dua Lipa »
Go on her channel page
Try to open « Future Nostalgia », or « Don't start now (Kungs Remix) », or « Don't Start Now », or a lot of other video on this page

## Exception
* __User Action:__ requested stream
* __Request:__ https://soundcloud.com/dualipa/future-nostalgia
* __Content Language:__ FR
* __Service:__ SoundCloud
* __Version:__ 0.18.2
* __OS:__ Linux Android 8.1.0 - 27

<details><summary><b>Crash log</b></summary><p>

```
org.schabi.newpipe.extractor.stream.StreamInfo$StreamExtractException: Could not get any stream. See error variable to get further details.
	at org.schabi.newpipe.extractor.stream.StreamInfo.extractStreams(StreamInfo.java:188)
	at org.schabi.newpipe.extractor.stream.StreamInfo.getInfo(StreamInfo.java:68)
	at org.schabi.newpipe.extractor.stream.StreamInfo.getInfo(StreamInfo.java:60)
	at org.schabi.newpipe.util.ExtractorHelper.lambda$getStreamInfo$3(ExtractorHelper.java:115)
	at org.schabi.newpipe.util.-$$Lambda$ExtractorHelper$5fJcha6Sq5APJBLdG6osaJby-mc.call(Unknown Source:4)
	at io.reactivex.internal.operators.single.SingleFromCallable.subscribeActual(SingleFromCallable.java:44)
	at io.reactivex.Single.subscribe(Single.java:3438)
	at io.reactivex.internal.operators.single.SingleDoOnSuccess.subscribeActual(SingleDoOnSuccess.java:35)
	at io.reactivex.Single.subscribe(Single.java:3438)
	at io.reactivex.internal.operators.maybe.MaybeFromSingle.subscribeActual(MaybeFromSingle.java:41)
	at io.reactivex.Maybe.subscribe(Maybe.java:4154)
	at io.reactivex.internal.operators.maybe.MaybeConcatArray$ConcatMaybeObserver.drain(MaybeConcatArray.java:153)
	at io.reactivex.internal.operators.maybe.MaybeConcatArray$ConcatMaybeObserver.request(MaybeConcatArray.java:78)
[...]

PR #300, opened 2020-03-30
throw ContentNotSupportedException when content is known to be unsuppo…

B0pol:
…rted

Follows the discussion we had in #250
It will greatly improve SoundCloud usability, because instead of randomly crashing, it will show the error with an appropriated UI, not a crash report, then the average user will understand why, and he can go back to play another track.
Before, the user was confused, and it went back to main kiosk each time, so it was really annoying.
Screenshots from UI:
 

PR in Front-end: https://github.com/TeamNewPipe/NewPipe/pull/3300

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudStreamExtractor.java
@@ -10,6 +10,7 @@ import org.schabi.newpipe.extractor.NewPipe;
 import org.schabi.newpipe.extractor.StreamingService;
 import org.schabi.newpipe.extractor.downloader.Downloader;
 import org.schabi.newpipe.extractor.exceptions.ContentNotAvailableException;
+import org.schabi.newpipe.extractor.exceptions.ContentNotSupportedException;
 import org.schabi.newpipe.extractor.exceptions.ExtractionException;
 import org.schabi.newpipe.extractor.exceptions.ParsingException;
 import org.schabi.newpipe.extractor.linkhandler.LinkHandler;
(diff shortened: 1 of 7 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamInfo.java
```diff
@@ -131,6 +132,8 @@ private static StreamInfo extractStreams(StreamInfo streamInfo, StreamExtractor
         /* Load and extract audio */
         try {
             streamInfo.setAudioStreams(extractor.getAudioStreams());
+        } catch (ContentNotSupportedException e) {
```
wb9688, line 135: Hmm… is this really the proper way to do this?

B0pol: I found this [stackoverflow answer](https://stackoverflow.com/a/20355868/12680950) and then did that, but if you know a cleaner way if there is, I'd be pleased to change it.

wb9688:
Oh, and please also do this for channels without Videos tab (e.g. Topic channels). You'll need to handle that in the front-end as well. Btw you forgot a `\n` in one translation there iirc, and I'd personally prefer 2 newlines, since I think that'd look better.

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamInfo.java
TobiGr: I think this is the cleanest way to do this. :+1:
Or at least the best way I know :)

theScrabi:
> (Off-topic: @theScrabi are you going to work more on NewPipe again?)

I'll try to get into it again (at leas a little). Although I feel a little alien right now as I don't exactly know what is going on right now or how the code looks right now.

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudStreamExtractor.java
```diff
@@ -196,6 +197,10 @@ public String getHlsUrl() throws ParsingException {
             throw new ExtractionException("Could not get SoundCloud's track audio url", e);
         }
 
+        if (audioStreams.isEmpty()) {
+            throw new ContentNotSupportedException("HLS audio streams / opus streams are not yet supported");
```
wb9688, line 201: Opus is with a capital O, however I'd suggest to not mention that at all, but just HLS, since the tracks that have Opus HLS streams, also have MP3 HLS streams.

PR #307, opened 2020-04-09
[PeerTube] playlist support

B0pol:

I've added playlist support and removed code duplication with `getNextPageUrl()` across peertube classes.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubePlaylistExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubePlaylistExtractor.java
@@ -65,6 +65,24 @@ public class PeertubePlaylistExtractor extends PlaylistExtractor {
 
     @Nonnull
     @Override
+    public String getSubChannelName() throws ParsingException {
+        return playlistInfo.getObject("videoChannel").getString("displayName");
+    }
+
+    @Nonnull
+    @Override
+    public String getSubChannelUrl() throws ParsingException {
+        return playlistInfo.getObject("videoChannel").getString("url");
+    }
+
+    @Nonnull
+    @Override
+    public String getSubChannelAvatarUrl() throws ParsingException {
+        return getBaseUrl() + playlistInfo.getObject("videoChannel").getObject("avatar").getString("path");
+    }
+
+    @Nonnull
+    @Override
     public InfoItemsPage<StreamInfoItem> getInitialPage() throws IOException, ExtractionException {
         return getPage(initialPageUrl);
     }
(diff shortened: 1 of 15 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubePlaylistExtractor.java
```diff
@@ -1,85 +1,117 @@
...
-        return 0;
+        return playlistInfo.getNumber("videosLength").longValue();
     }
 
+    @Nonnull
     @Override
     public InfoItemsPage<StreamInfoItem> getInitialPage() throws IOException, ExtractionException {
-        // TODO Auto-generated method stub
-        return null;
+        super.fetchPage();
```
wb9688, line 69: When `getInitialPage()` is called, `fetchPage()` should always already have been called, so no need to call it here. You also shouldn't fetch the initial page in `onFetchPage()` yet, but only in `getInitialPage()` and `getNextPageUrl()` if it hasn't been fetched yet.

I see that something similar is done in some other services though, but that should also be changed, and imho we should just remove `getNextPageUrl()`.

extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/linkHandler/PeertubeStreamLinkHandlerFactory.java
```diff
@@ -37,6 +37,12 @@ public String getId(String url) throws ParsingException, IllegalArgumentExceptio
 
     @Override
     public boolean onAcceptUrl(final String url) throws FoundAdException {
-        return url.contains("/videos/");
+        if (url.contains("playlist")) return false;
```
yausername, line 40: maybe use `/playlist/` ? that way it would not filter `/videos/watch/696969playlist696969`

B0pol: It was crashing without it, but it was because it was at the beginning and was badly implemented. This check is not even neeeded.

wb9688:
Could you add subchannel stuff for playlists as well?

extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubePlaylistExtractor.java
```diff
@@ -1,85 +1,135 @@
...
-        return null;
+        Response response = getDownloader().get(pageUrl);
+        try {
+            playlistVideos = JsonParser.object().from(response.responseBody());
+        } catch (JsonParserException jpe) {
+            throw new ExtractionException("Could not parse json", jpe);
+        }
+        PeertubeParsingHelper.validate(playlistVideos);
+
+        this.total = JsonUtils.getNumber(playlistVideos, "total").longValue();
```
wb9688, line 105: This shouldn't happen like this, however it's not worth fixing it until my next page PR (where you could just make it a normal variable in the function), especially as it's done like this in the other PeerTube ListExtractors as well.

PR #306, opened 2020-04-09
Extract metadata for youtube, soundcloud & mediaccc

B0pol:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).
- [x] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API.

Extraction for metadata which were added ~~not so~~ long ago (now), but only for PeerTube

fixes https://github.com/TeamNewPipe/NewPipeExtractor/issues/393

PR #309, opened 2020-04-09
Improve stream extractor tests & various fixes

Stypox:

This PR makes all stream extractor tests consistent by having them all extend the same default base class. I took inspiration from @mauriciocolli's work in #296. ~~Note that I didn't add tests for metadata properties, that will fit in another PR after both this pr and #306 are merged.~~

List of test improvements:
- Add base classes for stream extractor tests with the same structure as in #296, and extend them in each and every StreamExtractorTest. Some methods are not abstract and they provide a default value; this does not imply risks (i.e. forgetting to change a default) since if a value is different from the default one the tests will fail (not silently)
- Thumbnail urls and uploader avatar urls are only `assertSecureUrl`, but not compared, since they could change I think. In MediaCCC, where they were being tested before, I left those tests there untouched.
- Peertube has many description-only tests; I didn't create a whole test class for every one of them, but left them untouched, as a full test is not really needed imo.
- Rename `*StreamExtractorDefaultTest` into `*StreamExtractorTest` for services when there is only one file containing stream tests (that's obviously the default one ;-) )
- Remove assert checking whether the upload date of list items is in the past. Actually, this does not have to hold true: youtube premieres turn up in search results even though they are in the future.
- Add `assertAtLeast` method to utils
- Optimize imports

List of fixes (taken from commit messages):
- [YouTube] Fix `NullPointerException` on obtaining error message when there is none: return null instead, as per javadoc
- [YouTube] Return 0 when there is no timestamp in a video url, not `-2` (that internally indicates that the regex failed to match), as per javadoc
- [YouTube] Fix frame extraction for livestreams: another json field is used in that case. Also use saved playerResponse instead of parsing the same json once again.
- [SoundCloud] Return empty stream list instead of null in getVideoStreams() (best-practice). Also replace every instance of `return new ArrayList<>();` with `return Collections.emptyList();`
- [PeerTube] Return empty audio stream list instead of null in `getAudioStreams()`
- [PeerTube] Fix link handler inconsistency, that used to provide API links to the user instead of the human-readable ones. Also improve stream and channel link handler tests.
- [PeerTube] Parse timestamp from url (previously unimplemented)
- [MediaCCC] Fix link handler inconsistency, that used to provide API links to the user instead of the human-readable ones. Use regex instead of substrings.
- [MediaCCC] Add tests for stream and conference link handlers.
- [MediaCCC] Implement tag metadata extraction (unrelated to the other fixes, will probably be overwritten by #306)
- [MediaCCC] Return null instead of empty items collector at `getRelatedStreams()`, as per javadoc
- [MediaCCC] Remove useless overridden method `getOriginalUrl()` that does the same as the base one
- [MediaCCC] Return empty video-only stream list instead of null in `getVideoOnlyStreams()`


extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/linkHandler/MediaCCCStreamLinkHandlerFactory.java
```diff
@@ -9,10 +8,12 @@
...
+    public static final String VIDEO_API_ENDPOINT = "https://api.media.ccc.de/public/events/";
+    private static final String VIDEO_PATH = "https://media.ccc.de/v/";
 
     @Override
     public String getId(String urlString) throws ParsingException {
-        if (urlString.startsWith("https://api.media.ccc.de/public/events/") &&
+        if (urlString.startsWith(VIDEO_API_ENDPOINT) &&
                 !urlString.contains("?q=")) {
             return urlString.substring(39); //remove api…/public/events part
         }
```
wb9688, line 19: Why are you using a regex for conferences while using `substring()` for API URLs here and using `URL#getPath()` + `substring()` for non-API URLs here?

extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/linkHandler/PeertubeChannelLinkHandlerFactory.java
```diff
@@ -31,7 +32,7 @@ public String getUrl(String id, List<String> contentFilters, String searchFilter
     @Override
```
wb9688, line 32: Please also handle the API URLs in `getId()` here

extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/linkHandler/PeertubeStreamLinkHandlerFactory.java
```diff
@@ -27,7 +28,7 @@ public String getUrl(String id) {
 
     @Override
     public String getUrl(String id, String baseUrl) {
-        return baseUrl + VIDEO_ENDPOINT + id;
+        return baseUrl + VIDEO_PATH + id;
     }
 
     @Override
```
wb9688, line 34: Please also handle the API URLs in `getId()` here

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -258,7 +258,14 @@ public long getLength() throws ParsingException {
      */
     @Override
     public long getTimeStamp() throws ParsingException {
-        return getTimestampSeconds("((#|&|\\?)t=\\d{0,3}h?\\d{0,3}m?\\d{1,3}s?)");
+        long timestamp = getTimestampSeconds("((#|&|\\?)t=\\d{0,3}h?\\d{0,3}m?\\d{1,3}s?)");
```
wb9688, line 261: Could be `final`

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -579,7 +586,7 @@ public String getErrorMessage() {
         try {
             return getTextFromObject(initialAjaxJson.getObject(2).getObject("playerResponse").getObject("playabilityStatus")
                     .getObject("errorScreen").getObject("playerErrorMessageRenderer").getObject("reason"));
-        } catch (ParsingException e) {
+        } catch (Exception e) {
```
wb9688, line 589: Why was this changed?

extractor/src/test/java/org/schabi/newpipe/extractor/services/DefaultStreamExtractorTest.java
```diff
@@ -0,0 +1,289 @@
...
+    @Test
+    @Override
+    public void testThumbnailUrl() throws Exception {
+        assertIsSecureUrl(extractor().getThumbnailUrl());
+    }
+
+    @Test
+    @Override
+    public void testDescription() throws Exception {
+        Description description = extractor().getDescription();
```
wb9688, line 94: Could be `final`, just like a lot of variables in this class

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
Stypox: So that it also catches NullPointerExceptions, which are generated when there is no error message. In that case `null` should be returned

extractor/src/test/java/org/schabi/newpipe/extractor/services/DefaultStreamExtractorTest.java
Stypox: Ok, will fix this in all modified files

extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/linkHandler/MediaCCCStreamLinkHandlerFactory.java
Stypox: I left everything as it was before, I only replaced the hardcoded url. Will fix this

extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/linkHandler/PeertubeStreamLinkHandlerFactory.java
Stypox: Oh right, for some reason I thought they were already handled in some strange way

Stypox: Those are already handled by the regex: `/videos/(watch/)?([^/?&#]*)`
Both "framatube.org/api/v1/videos/id" and "framatube.org/videos/id" match it and the second group is obtained correctly

extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/linkHandler/PeertubeChannelLinkHandlerFactory.java
Stypox: Fixed while rebasing with the latest changes about accounts and channels

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
B0pol: Then why aren't you catching `(ParsingException | NullPointerException e)`?

Stypox:
Ok, I fixed everything that was suggested plus an oversight when testing video streams, check commit messages for more info. @B0pol @wb9688

wb9688:
@Stypox: It's rebase time! Yes, _again_. Anyway, I'll try to take an actual look at it tomorrow.

Stypox:
@wb9688 rebased and squashed some commits (in particular, I squashed all commits related to tests into only one, while living fixes into their separate commits). I learned how to rebase interactively :-D

Stypox:
I rebased

Stypox:
Rebased

extractor/src/test/java/org/schabi/newpipe/extractor/services/DefaultStreamExtractorTest.java
```diff
@@ -0,0 +1,297 @@
...
+        videoStreams.addAll(videoOnlyStreams);
+
+        if (expectedHasVideoStreams()) {
+            assertFalse(videoStreams.isEmpty());
+
+            for (VideoStream stream : videoStreams) {
+                assertIsSecureUrl(stream.getUrl());
+                assertFalse(stream.getResolution().isEmpty());
+
+                int formatId = stream.getFormatId();
```
wb9688, line 210: `final`

extractor/src/test/java/org/schabi/newpipe/extractor/services/DefaultStreamExtractorTest.java
```diff
@@ -0,0 +1,297 @@
...
+        final List<AudioStream> audioStreams = extractor().getAudioStreams();
+        assertNotNull(audioStreams);
+
+        if (expectedHasAudioStreams()) {
+            assertFalse(audioStreams.isEmpty());
+
+            for (AudioStream stream : audioStreams) {
+                assertIsSecureUrl(stream.getUrl());
+
+                int formatId = stream.getFormatId();
```
wb9688, line 231: `final`

extractor/src/test/java/org/schabi/newpipe/extractor/services/DefaultStreamExtractorTest.java
```diff
@@ -0,0 +1,297 @@
...
+            }
+        } else {
+            assertTrue(audioStreams.isEmpty());
+        }
+    }
+
+    @Test
+    @Override
+    public void testSubtitles() throws Exception {
+        List<SubtitlesStream> subtitles = extractor().getSubtitlesDefault();
```
wb9688, line 243: `final`

extractor/src/test/java/org/schabi/newpipe/extractor/services/DefaultStreamExtractorTest.java
```diff
@@ -0,0 +1,297 @@
...
+        assertNotNull(subtitles);
+
+        if (expectedHasSubtitles()) {
+            assertFalse(subtitles.isEmpty());
+
+            for (SubtitlesStream stream : subtitles) {
+                assertIsSecureUrl(stream.getUrl());
+
+                int formatId = stream.getFormatId();
```
wb9688, line 252: `final`

extractor/src/test/java/org/schabi/newpipe/extractor/services/DefaultStreamExtractorTest.java
```diff
@@ -0,0 +1,297 @@
...
+                assertTrue("format id does not fit an audio stream: " + formatId,
+                        0x1000 <= formatId && formatId < 0x10000);
+            }
+        } else {
+            assertTrue(subtitles.isEmpty());
+
+            MediaFormat[] formats = {MediaFormat.VTT, MediaFormat.TTML, MediaFormat.TRANSCRIPT1,
```
wb9688, line 259: `final`

extractor/src/test/java/org/schabi/newpipe/extractor/services/DefaultStreamExtractorTest.java
```diff
@@ -0,0 +1,297 @@
...
+                    MediaFormat.TRANSCRIPT2, MediaFormat.TRANSCRIPT3, MediaFormat.SRT};
+            for (MediaFormat format : formats) {
+                subtitles = extractor().getSubtitles(format);
```
wb9688, line 262: Make this a variable that's declared here, because this is imho unclear

Stypox:
@wb9688 fixed the failing test (for some reason SoundCloud started serving a date as GMT, causing a 2h difference between the expected parsed value and the actual one) and the inconsistencies you have found. Implemented metadata tests (currently only used in PeerTube, waiting for #306) and sub-channel tests. Rebased. ;-)

Stypox:
Oh, apparently tarvis is provided with dates in a different format. What the heck?

Stypox:
I thought the problem lied in PeerTube and SoundCloud date parsers, but actually this problem had to do with Youtube: the upload date wasn't being parsed as a GMT date, but using the device timezone, thus leading to inconsistencies. Tests were also wrong, because they tested dates using the devide timezone (I did try to prevent this, but I put the correct statement on the wrong line ;-/) Now everything seems to work, both on my end and on tarvis. It's incredible how 1085 tests succeed ;-D

wb9688:
Could you rebase?

Stypox:
Done. I'll soon also push a test for the DashMpd url that you recently fixed

Stypox:
Done

Stypox:
Merged with dev and fixed PewDiePie test

opusforlife2:
> fixed PewDiePie test

He needs all the help he can get.

Stypox:
I rebased. There are many failing tests, but all seem to be unrelated from the changes here, since they are caused by "YouTube did not provide player config even after three attempts", PeerTube accounts/channels not providing related items, SoundCloud channels & playlists with some outdated data, youtube channels with no Videos tab, YouTube comments extractor (all not touched by this PR).

TobiGr (approved):
Looks good to me!
We could make a bunch of vars and arguments final, but that can wait

Stypox:
@TobiGr done ;-)

extractor/src/test/java/org/schabi/newpipe/extractor/services/DefaultStreamExtractorTest.java
```diff
@@ -0,0 +1,382 @@
...
+
+        if (expectedHasAudioStreams()) {
+            assertFalse(audioStreams.isEmpty());
+
+            for (final AudioStream stream : audioStreams) {
+                assertIsSecureUrl(stream.getUrl());
+
+                final int formatId = stream.getFormatId();
+                assertTrue("format id does not fit an audio stream: " + formatId,
+                        0x100 <= formatId && formatId < 0x1000);
```
B0pol, line 274: Could you explain why 0x100 <= formatId < 0x1000 and also above (video streams) or below (subtitles) and add comments for it?

B0pol (changes requested):

extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCStreamExtractor.java
```diff
@@ -216,7 +219,7 @@ public StreamType getStreamType() {
 
     @Override
     public StreamInfoItemsCollector getRelatedStreams() {
-        return new StreamInfoItemsCollector(getServiceId());
+        return null;
```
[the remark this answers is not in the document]
Stypox: This was done as per the documentation in the base `getRelatedStreams()`, and also to prevent creating new objects at random. I tested in NewPipe and it works as expected (there are `null` checks), but the code which takes care of it being `null` actually lies [in the extractor](https://github.com/TeamNewPipe/NewPipeExtractor/blob/030465b5d4d8c738a42cfc52bd1d78f54b0f6bd4/extractor/src/main/java/org/schabi/newpipe/extractor/utils/ExtractorHelper.java#L33), where the collector is turned into a list. I now added @Nullable to all of the functions, as it was missing.

extractor/src/test/java/org/schabi/newpipe/extractor/services/DefaultStreamExtractorTest.java
Stypox: Ok

extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCStreamExtractor.java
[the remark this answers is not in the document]
Stypox: Oh, and also, YouTube already returns `null` for age restricted videos.

Stypox:
@B0pol done everything

Stypox:
Testing apk: [app-debug.zip](https://github.com/TeamNewPipe/NewPipeExtractor/files/5434271/app-debug.zip)

B0pol (approved):

PR #312, opened 2020-04-13
Do not report subscriber count, if it's not presented

### LINKED ISSUE #305: getSubscriberCount returns 0 incorrectly for YouTube
It seems that after [this](https://github.com/TeamNewPipe/NewPipeExtractor/commit/0798bdd5cd03bb44c0593a28b9c23df472e762ef#diff-1c13023aae5fe2aea7edc4e2a7658e15) getSubscriberCount returns 0 even when there is no known subscriber count, for example [here](https://www.youtube.com/channel/UCC76FBg2d-5NWFD8aK4ql5A/). I'm sure, this channel has more than 0 subscriber, just hidden, but as it has a 'Subscribe' button, which misleads the code a bit.
 
Found in https://github.com/ram-on/SkyTube/issues/728

gzsombor:

Fix #305

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
@@ -195,12 +195,7 @@ public class YoutubeChannelExtractor extends ChannelExtractor {
                 throw new ParsingException("Could not get subscriber count", e);
             }
         } else {
-            // If there's no subscribe button, the channel has the subscriber count disabled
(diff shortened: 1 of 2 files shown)


B0pol:
I don't think this should be merged. You should handle 0 in your front-end. If you want, update javadoc for getSubscriberCount: for youtube, 0 can mean either 0 or hidden.
[Full explanation here](https://github.com/TeamNewPipe/NewPipeExtractor/issues/305#issuecomment-613081317) (and please keep the discussion in only one place, then #305).

B0pol:
fixed in https://github.com/TeamNewPipe/NewPipeExtractor/commit/a39a2cca8261c7a6fe0286a503c3e16b8e22875d

PR #313, opened 2020-04-13
Display parent channel details

Royosef:
Add parent channel info to ChannelExtractor, StreamExtractor: name, url & avatar url

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamExtractor.java
@@ -179,6 +179,39 @@ public abstract class StreamExtractor extends Extractor {
     public abstract String getUploaderAvatarUrl() throws ParsingException;
 
     /**
+     * The Url to the page of the parent chanel of the stream. This must not be a homepage,
+     * but the page offered by the service the extractor handles. This url will be handled by the
+     * <a href="https://teamnewpipe.github.io/documentation/03_Implement_a_service/#channel">ChannelExtractor</a>,
+     * so be sure to implement that one before you return a value here, otherwise NewPipe will crash if one selects
+     * this url.
+     *
+     * @return the url to the page of the parent chanel of the stream or an empty String
+     * @throws ParsingException
+     */
+    @Nonnull
+    public abstract String getParentChannelUrl() throws ParsingException;
+
+    /**
+     * The name of the parent chanel of the stream.
+     * If the name is not available you can simply return an empty string.
+     *
+     * @return the name of the parent chanel of the stream or an empty String
+     * @throws ParsingException
+     */
+    @Nonnull
+    public abstract String getParentChannelName() throws ParsingException;
+
+    /**
+     * The url to the image file/profile picture/avatar of the parent chanel of the stream.
+     * If the url is not available you can return an empty String.
+     *
+     * @return The url of the image file of the parent chanel or an empty String
+     * @throws ParsingException
+     */
+    @Nonnull
+    public abstract String getParentChannelAvatarUrl() throws ParsingException;
+
+    /**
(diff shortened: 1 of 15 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamExtractor.java
```diff
@@ -178,6 +178,39 @@ public StreamExtractor(StreamingService service, LinkHandler linkHandler) {
     @Nonnull
     public abstract String getUploaderAvatarUrl() throws ParsingException;
 
+    /**
+     * The Url to the page of the parent chanel of the stream. This must not be a homepage,
+     * but the page offered by the service the extractor handles. This url will be handled by the
+     * <a href="https://teamnewpipe.github.io/documentation/03_Implement_a_service/#channel">ChannelExtractor</a>,
```
wb9688, line 184: Use `{@link ChannelExtractor}` instead

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamExtractor.java
```diff
@@ -178,6 +178,39 @@ public StreamExtractor(StreamingService service, LinkHandler linkHandler) {
...
+     * so be sure to implement that one before you return a value here, otherwise NewPipe will crash if one selects
+     * this url.
+     *
+     * @return the url to the page of the parent chanel of the stream or an empty String
+     * @throws ParsingException
+     */
+    @Nonnull
+    public abstract String getParentChannelUrl() throws ParsingException;
```
wb9688, line 192: Should we add a `getParentChannelUrl` or a `getSubchannelUrl` here? Same thing for name and avatar URL obviously,

Imho we should add `getSubchannelUrl`, as currently we return the parent channel's URL, so the values returned by `getChannelUrl` stays the same. I want someone else's opinion here first though.

Edit: it looks like you're returning the subchannel's URL here, so that's obviously not correct anyway.

Royosef: I'm not really sure myself what will be the right naming. In Peertube it's "account page" and "channel page", maybe we should do the same?
image

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamExtractor.java
Royosef: hmm, truth is I just copied the correspond comment from `getUploaderUrl`. How can I use `{@link ChannelExtractor}`?

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamExtractor.java
wb9688: @Royosef: No, just change it to `getSubchannelUrl()`. We aren't going to call it accounts and channels as that's unclear and only PeerTube calls it that.

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamExtractor.java
wb9688: @Royosef: Just change the whole `<a href="https://teamnewpipe.github.io/documentation/03_Implement_a_service/#channel">ChannelExtractor</a>` to `{@link ChannelExtractor}`.

extractor/build.gradle
```diff
@@ -1,3 +1,6 @@
+compileJava.options.encoding = "UTF-8"
+compileTestJava.options.encoding = "UTF-8"
+
```
wb9688, line 3: This is unrelated to the PR.

wb9688:
I rebased and removed the unrelated `extractor/build.gradle` change.

PR #317, opened 2020-04-16
Use our fork of nanojson

wb9688:

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicSearchExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicSearchExtractor.java
@@ -5,7 +5,6 @@ import com.grack.nanojson.JsonObject;
 import com.grack.nanojson.JsonParser;
 import com.grack.nanojson.JsonParserException;
 import com.grack.nanojson.JsonWriter;
-
 import org.schabi.newpipe.extractor.InfoItem;
 import org.schabi.newpipe.extractor.StreamingService;
 import org.schabi.newpipe.extractor.downloader.Downloader;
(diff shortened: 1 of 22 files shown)


extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/stream/YoutubeStreamExtractorDefaultTest.java
```diff
@@ -22,6 +22,7 @@
 import java.text.SimpleDateFormat;
 import java.util.Calendar;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
```
TobiGr, line 25: unused

wb9688: Yup, I used that for getting JSON that nanojson fails to parse without spamming YouTube's servers.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicSearchExtractor.java
```diff
@@ -301,16 +294,7 @@ public String getUploaderUrl() throws ParsingException {
...
-                                    return null;
-                                }
-
-                                final String url = getUrlFromNavigationEndpoint(navigationEndpoint);
-
-                                if (url != null && !url.isEmpty()) {
-                                    return url;
-                                }
-                                throw new ParsingException("Could not get uploader url");
+                                return getUrlFromNavigationEndpoint(navigationEndpoint);
```
TobiGr, line 297: `getUrlFromNavigationEndpoint(navigationEndpoint)` returns null when navigationEndpoint is an empty Object or YouTube changed something in their JSON. In this case, we could not get the uploader URL and should throw a ParsingException. So please keep the null check for the URL.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
```diff
@@ -98,7 +93,7 @@ public String getUrl() throws ParsingException {
     @Override
     public String getName() throws ParsingException {
         String name = getTextFromObject(videoInfo.getObject("title"));
-        if (name != null && !name.isEmpty()) return name;
+        if (name == null || !name.isEmpty()) return name;
```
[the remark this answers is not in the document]
wb9688: This change is indeed an error, and `name` could indeed not be `null`. There are probably other usages of `getTextFromObject()` where that check could be removed as well. I'll do that tomorrow.

PR #330, opened 2020-05-11
add vscode / eclipse files to .gitignore

B0pol:

This would save me lots of `git clean -d -f` when I just wanna quickly check a file in vscode.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@ gradle-app.setting
 *.project
 *.settings
 **/bin
+**.vscode
+*.code-workspace
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,9 @@ gradle-app.setting
 
 # # Work around https://youtrack.jetbrains.com/issue/IDEA-116898
 # gradle/wrapper/gradle-wrapper.properties
+
+# vscode / eclipse files
+*.classpath
+*.project
+*.settings
+**/bin


Stypox:
Shouldn't you also add `*.vscode/`?

B0pol:
> Shouldn't you also add `*.vscode/`?

I've never encountered this file and it seems they doesn't exist by looking on internet.

wb9688:
See:

- https://github.com/github/gitignore/blob/master/Java.gitignore
- https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore
- https://github.com/github/gitignore/blob/master/Global/Eclipse.gitignore
- https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore

Obviously we shouldn't add the things that are irrelevant to us.

B0pol:
> Obviously we shouldn't add the things that are irrelevant to us.

Yes, that's why I won't add `.vscode`, it's irrelevant. It's not doing that when I use vscode. If it does, someone can open another PR.

Stypox:
> > Shouldn't you also add `*.vscode/`?
> 
> I've never encountered this file and it seems they doesn't exist by looking on internet.

This is the folder containing all vscode configuration for a project, how can you have never seen it? See https://code.visualstudio.com/docs/getstarted/settings

B0pol:
I don't now, i just press `code newpipe` or `code newpipeextractor`, or open folver via GUI. I'll try with vanilla vscode.

B0pol:
I've tried with vanilla vscode.
1. I've never used workspaces, but just used `open folder`. That may be a reason why.
2. I've tried to change some settings, and it was asked to create a .code-workspace file, but it didn't created any `.vscode` folder

B0pol:
Can this be merged?

wb9688:
Quoting what I said on IRC:

>No, I didn't approve the NPE one for a reason:  **/bin should be bin/ and  **.vscode should be .vscode/

>In .gitignore, it'll match any directory called "bin" with "bin/". "/bin/" would make it relative to the .gitignore file. So "\*\*/bin" is pointless. "**.vscode" seems to do the same thing as "*.vscode", but the folder is always just called ".vscode", so you should just do ".vscode/". When it has a "/" at the end, it'll only match directories with that name instead of all inodes with that name, but it mainly makes it clearer to anyone reading it whether that's supposed to be a file or a folder.

PR #332, opened 2020-05-13
Support YouTube's learning playlists

wb9688:

Fixes the issue mentioned by @AwesomelyHumble in TeamNewPipe/NewPipe#3569.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubePlaylistExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubePlaylistExtractor.java
@@ -32,6 +32,7 @@ import static org.schabi.newpipe.extractor.utils.Utils.isNullOrEmpty;
 
 @SuppressWarnings("WeakerAccess")
 public class YoutubePlaylistExtractor extends PlaylistExtractor {
+    private JsonArray initialAjaxJson;
     private JsonObject initialData;
     private JsonObject playlistInfo;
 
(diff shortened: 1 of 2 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubePlaylistExtractor.java
```diff
@@ -189,18 +208,18 @@ public String getSubChannelAvatarUrl() throws ParsingException {
             throw new ExtractionException(new IllegalArgumentException("Page url is empty or null"));
         }
 
-        StreamInfoItemsCollector collector = new StreamInfoItemsCollector(getServiceId());
+        final StreamInfoItemsCollector collector = new StreamInfoItemsCollector(getServiceId());
         final JsonArray ajaxJson = getJsonResponse(pageUrl, getExtractorLocalization());
```
Stypox, line 212: I think this name should be changed, `ajaxJson` is already used as a class field.

wb9688: You're right, this is why we need Checkstyle here as well…

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubePlaylistExtractor.java
```diff
@@ -27,6 +32,7 @@
 
 @SuppressWarnings("WeakerAccess")
 public class YoutubePlaylistExtractor extends PlaylistExtractor {
+    private JsonArray ajaxJson;
```
Stypox, line 35: .

PR #337, opened 2020-05-30
Remove getNextStream() from StreamExtractor

wb9688:

This is the NPE part of the fix for TeamNewPipe/NewPipe#3684, I'll do the NP part later today.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCStreamExtractor.java
@@ -15,7 +15,6 @@ import org.schabi.newpipe.extractor.localization.DateWrapper;
 import org.schabi.newpipe.extractor.stream.AudioStream;
 import org.schabi.newpipe.extractor.stream.Description;
 import org.schabi.newpipe.extractor.stream.StreamExtractor;
-import org.schabi.newpipe.extractor.stream.StreamInfoItem;
 import org.schabi.newpipe.extractor.stream.StreamInfoItemsCollector;
 import org.schabi.newpipe.extractor.stream.StreamType;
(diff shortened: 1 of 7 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCStreamExtractor.java
```diff
@@ -215,11 +214,6 @@ public StreamType getStreamType() {
         return StreamType.VIDEO_STREAM;
     }
 
-    @Override
-    public StreamInfoItem getNextStream() {
-        return null;
-    }
-
     @Override
     public StreamInfoItemsCollector getRelatedStreams() {
```
TobiGr, line 218: Now that `getNextStream()` is removed, we should add a JDoc here to say that the first related stream might be what was previously handled as "next stream".

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -528,10 +526,7 @@ public StreamInfoItem getNextStream() throws ExtractionException {
                     .getArray("contents").getObject(0).getObject("compactVideoRenderer");
 
             final TimeAgoParser timeAgoParser = getTimeAgoParser();
-            StreamInfoItemsCollector collector = new StreamInfoItemsCollector(getServiceId());
-
-            collector.commit(new YoutubeStreamInfoItemExtractor(videoInfo, timeAgoParser));
-            return collector.getItems().get(0);
+            return new YoutubeStreamInfoItemExtractor(videoInfo, timeAgoParser);
```
Stypox, line 529:
```suggestion
            return new YoutubeStreamInfoItemExtractor(videoInfo, getTimeAgoParser());
```
and remove the line above

PR #338, opened 2020-05-30
Add tests for parseDurationString()

wb9688:

I added some tests for `parseDurationString()` and found out that apparently the days behavior was already broken, however YouTube doesn't seem to use days anymore.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
@@ -142,9 +142,10 @@ public class YoutubeParsingHelper {
             default:
                 throw new ParsingException("Error duration string with unknown format: " + input);
         }
-        return ((((Long.parseLong(Utils.removeNonDigitCharacters(days)) * 24)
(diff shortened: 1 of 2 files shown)


TobiGr:
@Redirion you approved this PR although you asked to include the changes from #345. I think you suggested changes are good. Is there a reason which I missed to not adopt them?

wb9688:
@TobiGr: My personal reason would be that the function that calls it returns a long anyway, plus the speed difference is pretty much negligible I think.

Redirion:
@TobiGr
I didn't notice the pr earlier (or more specific that it included a fix). It is good to merge.

@wb9688 think of 32bit arm entry level phones. Long arithmetics should be avoided if not needed. 100 negligible optimizations could make a feelable difference. :)



My PR did also change the return type.

wb9688:
@Redirion: I meant e.g. https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java#L102 and all the other functions calling it.

Redirion:
Long arithmetics on a 32bit cpu require more ressources and operations. Storing and reading it as long should not be an issue.

PR #349, opened 2020-06-11
Use OkHttp for tests like in NewPipe

wb9688:

### DIFF
--- a/extractor/build.gradle
+++ b/extractor/build.gradle
@@ -8,4 +8,5 @@ dependencies {
     implementation 'org.nibor.autolink:autolink:0.10.0'
 
     testImplementation 'junit:junit:4.13'
+    testImplementation "com.squareup.okhttp3:okhttp:3.12.11"
 }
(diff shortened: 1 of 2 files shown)


extractor/build.gradle
```diff
@@ -8,4 +8,5 @@ dependencies {
     implementation 'org.nibor.autolink:autolink:0.10.0'
 
     testImplementation 'junit:junit:4.13'
+    testImplementation "com.squareup.okhttp3:okhttp:3.12.11"
```
Stypox, line 11: Can't you use the latest version? Since this only needs to be run in tests, it does not need to be API19-compatible

wb9688: Yes, however I think it's a good idea to use the same version in NewPipe and NewPipeExtractor's tests. OkHttp 3.12 will also just still be supported until December 31, 2021.

Stypox: Ok, I see

Stypox:
@TobiGr I wasn't able to merge it due to the failing Travis CI, but I already added the PR to the release notes. So just merge ;-)

PR #359, opened 2020-06-24
[SoundCloud] Fix concurrency issue when getting the client id

mauriciocolli:
The client id method didn't work properly when the hardcoded key was no longer valid, an easy example would be when playing a playlist that pre-loaded more than 1 item.

Simply making the method `synchronized` should solve this problem.


wb9688:
Thanks a lot for fixing this!

Why does this not happen with YouTube or YouTube Music though?

PS: I didn't take a look at the relevant code yet, but I definitely will tomorrow hopefully.

PR #361, opened 2020-06-26
expose extractor via api scope

yausername:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,7 +15,7 @@ allprojects {
 }
 
 dependencies {
-    implementation project(':extractor')
+    api project(':extractor')
     implementation project(':timeago-parser')
 }
 


wb9688:
I think this change is good, however why are we currently able to call stuff in  `:extractor` from e.g. NewPipe then?

yausername:
> I think this change is good, however why are we currently able to call stuff in `:extractor` from e.g. NewPipe then?

That's because jitpack packages a super module which exposes the root module and all submodules.
You can see the difference here (notice the `scope` tag in dependencies)
[the actual root module](https://jitpack.io/com/github/TeamNewPipe/NewPipeExtractor/NewPipeExtractor/v0.19.5/NewPipeExtractor-v0.19.5.pom)
[the super module packaged by jitpack](https://jitpack.io/com/github/TeamNewPipe/NewPipeExtractor/v0.19.5/NewPipeExtractor-v0.19.5.pom)

So in short, it works if you use jitpack.

PR #352, opened 2020-06-11
Invidious support

B0pol:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [ ] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).
- [ ] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API.

TODO
- [x] implement InvidiousChannelExtractor
- [x] implement InvidiousChannelInfoItemExtractor
- [x] implement InvidiousCommentsExtractor
- [x] implement InvidiousCommentsInfoItemExtractor
- [x] implement InvidiousFeedExtractor
- [x] implement InvidiousStreamInfoItemExtractor
- [x] implement InvidiousPlaylistExtractor
- [x] implement InvidiousPlaylistInfoItemExtractor
- [x] implement InvidiousSearchExtractor
- [x] implement InvidiousSuggestionExtractor
- [x] implement InvidiousTrendingExtractor
- [ ] implement InvidiousSubscriptionExtractor
- [ ] add tests in NewPipeTest.java for Invidious (it will partially `fix` #347)
- [ ] handle when instance is semi-blocked: i.e. when video is available but metadata are missing, e.g. 
image
- [ ] **add tests**
- [ ] test in NewPipe
- [ ] next page logic when there is no `continuation` field but a `page` url parameter.
- [ ] mix extractor?

what can't be done:

- MusicSearchExtractor

maybe:
- search filters? the api provides 
```
sort_by: "relevance", "rating", "upload_date", "view_count"
date: "hour", "today", "week", "month", "year"
duration: "short", "long"
type: "video", "playlist", "channel", "all", (default: video)
features: "hd", "subtitles", "creative_commons", "3d", "live", "purchased", "4k", "360", "location", "hdr" (comma separated: e.g. "&features=hd,subtitles,3d,live")
region: ISO 3166 country code (default: "US")
```
[...]

PR #365, opened 2020-07-02
Update inv instances

B0pol:

https://github.com/TeamNewPipe/NewPipe/issues/3654 part 1 of 2


B0pol:
> Looks good.
> Should we set up tests to detect if any of these instances is down / removed from the "official" instances list?

No, in the extractor the list will be dropped with #352: we will dynamically know if it's an invidious instance by querying the API.
Though the list in manifest must be hardcoded, and I agree we could detect with tests.

TobiGr:
Ah yes, I forgot about #352.

PR #358, opened 2020-06-21
Add support for different types of stream delivery by a service

mauriciocolli:
I was working in an extractor for Twitch, and this was needed as it only offer HLS streams.

This also fixes some issues that some YouTube videos were having, as they only had DASH streams. We should be able to fix SoundCloud as well now, I'll leave it for later.

- Add a field describing how a stream is delivered by a service
- YouTube
  - Improve DASH manifest parsing
  - Ignore streams using the "OTF" delivery format

---

Closes #273

PR #367, opened 2020-07-15
Properly support other delivery methods

wb9688:
- [x] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it 
compatible with the changed API.

### Changes

- [x] Add `DeliveryFormat` enum
- [x] Rename `url` in `Stream` to `content` and add `isUrl` to indicate whether `content` contains the URL or the file itself (which is useful when we have to split DASH/HLS manifests)
- [x] Add `id` in `Stream` to identify a specific stream (there could be multiple `Stream`s with the same ID if the same stream is offered through multiple delivery methods; in YouTube's case the ID would be the itag)
- [x] Support Opus and MP3 over HLS for SoundCloud
- [x] Parse YouTube's DASH MPD

When all changes are finished, I'm going to work on implementing them in NewPipe's downloader. When I'm done with that and the unified player has been merged, I'll look at implementing it in the player as well.

I'll leave not hardcoding itags to another PR.


dbenjaminmiller:
Is anyone able to review this?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -636,6 +648,10 @@ public void onFetchPage(@Nonnull Downloader downloader) throws IOException, Extr
         if (subtitlesInfos.isEmpty()) {
             subtitlesInfos.addAll(getAvailableSubtitlesInfo());
         }
+
+        if (!this.getDashMpdUrl().isEmpty()) {
+            dashResult = DashMpdParser.getStreams(this.getDashMpdUrl());
```
Stypox, line 653: Do not calculate `getDashMpdUrl()` twice

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -909,27 +926,35 @@ private static String getVideoInfoUrl(final String id, final String sts) {
...
-                    ItagItem itagItem = ItagItem.getItag(itag);
+                    final ItagItem itagItem = ItagItem.getItag(itag);
                     if (itagItem.itagType == itagTypeWanted) {
+                        // Ignore streams that are delivered using YouTube's OTF format,
+                        // as those only work with DASH and not with progressive HTTP.
+                        // Those should be extracted when parsing the DASH MPD
+                        if (formatData.getString("type", EMPTY_STRING)
+                                .equalsIgnoreCase("FORMAT_STREAM_TYPE_OTF")) {
+                            continue;
+                        }
```
Stypox, line 951: Note: from the research I did a while ago it it theoretically possible to generate the manifest content based on this URL, without having to fetch the manifest with another time-consuming request

wb9688: I know, plus we don't need the whole `ItagItem` class, but I'll leave that to a future PR.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
wb9688: Yeah, I'll fix that later (today?) and add Mauricio as co-author of 8a9876067df764a81ed2430f39df267bd27bd288 as that part was mostly from #358.

dbenjaminmiller: Have you been able to make the fix yet @wb9688 ?

wb9688:
@Stypox: Could you review this again? This PR shouldn't be merged yet though, as I haven't properly tested e.g. whether the YouTube DASH extraction is actually correct (the output *looks* correct though). Btw I just started working on the NewPipe part of this.

extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubeStreamExtractor.java
```diff
@@ -360,8 +359,10 @@ private void loadSubtitles() {
                         final String languageCode = JsonUtils.getString(caption, "language.id");
                         final String ext = url.substring(url.lastIndexOf(".") + 1);
                         final MediaFormat fmt = MediaFormat.getFromSuffix(ext);
-                        if (fmt != null && languageCode != null)
-                            subtitles.add(new SubtitlesStream(fmt, languageCode, url, false));
+                        if (fmt != null && languageCode != null) {
+                            final String id = languageCode + "." + fmt.suffix;
```
Stypox, line 363: Make a constructor of SubtitlesStream without id, so that it can be automatically set to this. I see this is being used also in YoutubeStreamExtractor

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
```diff
@@ -174,52 +174,56 @@ public String getHlsUrl() throws ParsingException {
...
+            String url = t.getString("url");
+
+            if (!isNullOrEmpty(url)) {
+                final MediaFormat mediaFormat;
+                if (t.getString("preset").contains("mp3")) {
+                    mediaFormat = MediaFormat.MP3;
+                } else if (t.getString("preset").contains("opus")) {
+                    mediaFormat = MediaFormat.OPUS;
+                } else {
+                    break;
```
Stypox, line 201: Why `break` and not `continue`? Also below.

Stypox (changes requested):
LGTM, thank you :-D
I only pointed out two small things, after they are solved and you have finished with your tests this can be merged imo

PR #370, opened 2020-07-18
Fix age restricted YouTube videos

wb9688:


wb9688:
This does fix the "content not available", however you can't see the channel avatar or the likes/dislikes. Channel avatar could be fixed by parsing `embedded_player_response`, however I gave up on doing that today. That would also make the `get_video_info` request unneeded (though it'd be a good idea to keep that around as fallback method).

PR #379, opened 2020-08-02
Allow adding services

wb9688:

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/ServiceList.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/ServiceList.java
@@ -5,6 +5,8 @@ import org.schabi.newpipe.extractor.services.peertube.PeertubeService;
 import org.schabi.newpipe.extractor.services.soundcloud.SoundcloudService;
 import org.schabi.newpipe.extractor.services.youtube.YoutubeService;
 
+import java.lang.reflect.InvocationTargetException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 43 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/ServiceList.java
```diff
@@ -44,20 +46,33 @@ private ServiceList() {
...
+    }
+
+    public static void addService(final Class<StreamingService> service) throws NoSuchMethodException, IllegalAccessException, InvocationTargetException, InstantiationException {
+        services.add(service.getConstructor(new Class[]{int.class}).newInstance(nextService));
+        nextService++;
+    }
+
+    public static void replaceService(final Class<StreamingService> service, final int serviceId) throws NoSuchMethodException, IllegalAccessException, InvocationTargetException, InstantiationException {
+        services.set(serviceId, service.getConstructor(new Class[]{int.class}).newInstance(serviceId));
+        nextService++;
```
TobiGr, line 76: that's a security nightmare

TobiGr:
If we want to allow loading additional services, we need a proper API management. e.g. the extractor needs an internal API version so that only compatible services are loaded.

extractor/src/main/java/org/schabi/newpipe/extractor/ServiceList.java
wb9688: Reflection in itself isn't a security nightmare. The NewPipe-side with `PathClassLoader` is perhaps not so great security wise (but other ways of having extensions can't be more secure, except when we add signing), but that's why I've clearly added a message in NewPipe that you should only install extensions that you trust, though perhaps I should emphasize that more.

wb9688:
@TobiGr: I agree that we need an internal NewPipeExtractor version. For now I just added a simple version check in NewPipe, which is stupid. I also need to change NewPipe's DB scheme to store the service name instead of ID.

PR #394, opened 2020-09-03
Fix issue #4213:  trim the last char of the URL if it is a closing parenthesis

kirisakow:
Fix [#4213](https://github.com/TeamNewPipe/NewPipe/issues/4213) issue
trim the last char of the URL if it is a closing parenthesis

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeStreamLinkHandlerFactory.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeStreamLinkHandlerFactory.java
@@ -12,6 +12,8 @@ import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import static org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper.BASE_YOUTUBE_INTENT_URL;
 
@@ -37,6 +39,7 @@ import static org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper
 
 public class YoutubeStreamLinkHandlerFactory extends LinkHandlerFactory {
 
+    private static final Pattern YOUTUBE_VIDEO_ID_REGEX_PATTERN = Pattern.compile("([a-zA-Z0-9_-]{11})");
     private static final YoutubeStreamLinkHandlerFactory instance = new YoutubeStreamLinkHandlerFactory();
 
     private YoutubeStreamLinkHandlerFactory() {
@@ -46,13 +49,19 @@ public class YoutubeStreamLinkHandlerFactory extends LinkHandlerFactory {
         return instance;
     }
 
-    private static boolean isId(@Nullable String id) {
-        return id != null && id.matches("[a-zA-Z0-9_-]{11}");
+    @Nullable
+    private static String extractId(@Nullable final String id) {
+        if (id != null) {
+            final Matcher m = YOUTUBE_VIDEO_ID_REGEX_PATTERN.matcher(id);
+            return m.find() ? m.group(1) : null;
+        }
+        return null;
     }
 
-    private static String assertIsId(@Nullable String id) throws ParsingException {
-        if (isId(id)) {
-            return id;
+    private static String assertIsId(@Nullable final String id) throws ParsingException {
+        final String extractedId = extractId(id);
+        if (extractedId != null) {
+            return extractedId;
         } else {
             throw new ParsingException("The given string is not a Youtube-Video-ID");
         }
@@ -81,9 +90,9 @@ public class YoutubeStreamLinkHandlerFactory extends LinkHandlerFactory {
             if (scheme != null && (scheme.equals("vnd.youtube") || scheme.equals("vnd.youtube.launch"))) {
                 String schemeSpecificPart = uri.getSchemeSpecificPart();
                 if (schemeSpecificPart.startsWith("//")) {
-                    final String possiblyId = schemeSpecificPart.substring(2);
-                    if (isId(possiblyId)) {
-                        return possiblyId;
+                    final String extractedId = extractId(schemeSpecificPart.substring(2));
+                    if (extractedId != null) {
+                        return extractedId;
                     }
 
                     urlString = "https:" + schemeSpecificPart;
--- a/extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeStreamLinkHandlerFactoryTest.java
+++ b/extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeStreamLinkHandlerFactoryTest.java
@@ -81,6 +81,14 @@ public class YoutubeStreamLinkHandlerFactoryTest {
         assertEquals("jZViOEv90dI", linkHandler.fromUrl("vnd.youtube:jZViOEv90dI").getId());
         assertEquals("n8X9_MgEdCg", linkHandler.fromUrl("vnd.youtube://n8X9_MgEdCg").getId());
         assertEquals("O0EDx9WAelc", linkHandler.fromUrl("https://music.youtube.com/watch?v=O0EDx9WAelc").getId());
+        assertEquals("-cdveCh1kQk", linkHandler.fromUrl("https://m.youtube.com/watch?v=-cdveCh1kQk)").getId());
+        assertEquals("-cdveCh1kQk", linkHandler.fromUrl("https://www.youtube.com/watch?v=-cdveCh1kQk-").getId());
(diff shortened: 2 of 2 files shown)


wb9688:
YouTube trims any additional characters from `v`. This implementation is incorrect.

kirisakow:
Dear @wb9688, perhaps I was not clear enough. What I am trying to fix here is NewPipe's being unable to handle YouTube links which, inadvertedly, include a closing parenthesis at the end. I invite you to open this very page in your favorite web browser for Android, and try to open with NewPipe each one of the two links below: the first shall work, whereas the second shall not. And that's what I am trying to fix here by trimming the last character from the video's ID if that character is a closing parenthesis.

1. [https://www.youtube.com/watch?v=mBuoZY7kAlg](https://www.youtube.com/watch?v=mBuoZY7kAlg)
2. [https://www.youtube.com/watch?v=mBuoZY7kAlg)](https://www.youtube.com/watch?v=mBuoZY7kAlg%29)

Thank you for your attention. Please help me implement that fix.

wb9688:
Try opening e.g. https://www.youtube.com/watch?v=mBuoZY7kAlgwhatever. Also note how with e.g. https://www.youtube.com/watch?list=PLEqQd15F3NFGFUYJhz8KvGNz-QDhJVynT&v=OGS7c0-CmRswhatever it will strip the other parameters.

kirisakow:
Dear friend, please watch me reproducing the issue with my Android device in Firefox for Android web browser:

https://youtu.be/bZP0jtp9WrQ

wb9688:
Again, YouTube strips any additional characters (and if there are additional characters, it will strip other parameters), while NewPipe considers the URL simply invalid, however this PR naively strips the last character if it's ), which is also incorrect as it should exactly follow YouTube's behavior.

kirisakow:
All right, as you wish

opusforlife2:
@kirisakow He's talking about implementing the generic Youtube-like behaviour so it handles all cases of improper URLs, not just the parenthesis one. Did you close the PR because you don't wish to take that on?

kirisakow:
I closed because l wasn't told what needs to be done — only that my solution was incorrect.

If you are open to tell me what you think needs to be done — I'll do it.

Stypox:
Let me try to provide a clearer explanation, sorry for the confusion ;-) 
Your implementation is not optimal since:
- it is applied to all services, but some services could have a closing parenthesis as  part of the id, so in that case the link would be made invalid. Therefore it should only be applied to YouTube (see YoutubeStreamLinkHandlerFactory, YoutubePlaylistLinkHandlerFactory and the similar ones)
- it does not behave the same way as the official youtube client. What wb9688 suggested was to not only remove parenthesis, but also all spare characters at the end of the url, while you are at it. That's what the official yt app and official browser version do: every character after the id is just ignored. To make this compliant you could trim the id to only the first N characters (where N is the length of the id, =11 for yt videos)

Thank you for your contribution and sorry for the confusion ;-)

Stypox:
Oh, forgot to add that you should keep this in mind while implementing: when the official browser version gets a video [didn't check other features] url with an id parameter `v=` longer than 11 chars, all of the following url parameters are ignored

kirisakow:
Another solution applied to `YoutubeStreamLinkHandlerFactory` class alone (for now). Before I move on, please tell me if you guys are happy with it.

kirisakow:
I applied the suggested solution to the `YoutubeStreamLinkHandlerFactory` class alone (for now). Before I move on, please tell me if you guys are happy with it. @Stypox @wb9688 @opusforlife2 @gzsombor

opusforlife2:
I'm not a dev so I can't comment on that aspect, but thank you for continuing to work on this. 🥂

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeStreamLinkHandlerFactory.java
```diff
@@ -46,13 +48,25 @@ public static YoutubeStreamLinkHandlerFactory getInstance() {
         return instance;
     }
 
+    private static String extractId(@Nullable String id) {
+        if (id != null) {
+            Matcher m = Pattern.compile("([a-zA-Z0-9_-]{11})").matcher(id);
+            return m.find() ? m.group(1) : "";
```
Stypox, line 54: Return `null` instead of an empty string, if you found nothing (also remember to add `@Nullable` to the function)

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeStreamLinkHandlerFactory.java
```diff
@@ -46,13 +48,25 @@ public static YoutubeStreamLinkHandlerFactory getInstance() {
...
+        }
+        return "";
+    }
+
     private static boolean isId(@Nullable String id) {
```
Stypox, line 59: We might consider removing `isId()` altogether and replacing
```java
if (isId()) {
    return extractId()
}
```
with 
```java
final String extractedId = extractId(id);
if (extractedId != null) {
    return extractedId;
}
```
which does one regex comparison less, and is even clearer imo

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeStreamLinkHandlerFactory.java
```diff
@@ -46,13 +48,25 @@ public static YoutubeStreamLinkHandlerFactory getInstance() {
         return instance;
     }
 
+    private static String extractId(@Nullable String id) {
+        if (id != null) {
+            Matcher m = Pattern.compile("([a-zA-Z0-9_-]{11})").matcher(id);
```
Stypox, line 53: Move `Pattern.compile("([a-zA-Z0-9_-]{11})")` to a `public static final` class field, to prevent compiling it every time.

Stypox (changes requested):
This looks better, thank you ;-)
I only pointed out a few style issues, but the overall approach is good. Add `final` to all variables as you can, including function arguments.

kirisakow:
@Stypox Done

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeStreamLinkHandlerFactory.java
```diff
@@ -82,8 +91,9 @@ public String getId(String urlString) throws ParsingException, IllegalArgumentEx
                 String schemeSpecificPart = uri.getSchemeSpecificPart();
                 if (schemeSpecificPart.startsWith("//")) {
                     final String possiblyId = schemeSpecificPart.substring(2);
-                    if (isId(possiblyId)) {
-                        return possiblyId;
+                    final String extractedId = extractId(possiblyId);
```
Stypox, line 94: Merge this and the above line into
```java
final String possiblyId = extractId(schemeSpecificPart.substring(2));
```

Stypox (changes requested):
Please add one or two tests for urls with this longer-than-usual id (in YoutubeStreamLinkHandlerFactoryTest should do)
Thank you!

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeStreamLinkHandlerFactory.java
```diff
@@ -46,13 +49,19 @@ public static YoutubeStreamLinkHandlerFactory getInstance() {
         return instance;
     }
 
-    private static boolean isId(@Nullable String id) {
-        return id != null && id.matches("[a-zA-Z0-9_-]{11}");
+    @Nullable
+    private static String extractId(@Nullable final String id) {
+        if (id != null) {
+            Matcher m = YOUTUBE_VIDEO_ID_REGEX_PATTERN.matcher(id);
```
TobiGr, line 55:
```suggestion
            final Matcher m = YOUTUBE_VIDEO_ID_REGEX_PATTERN.matcher(id);
```

TobiGr (commented):
Thanks

TobiGr (approved):

wb9688 (changes requested):
Issues I pointed out still aren't fixed…

Stypox:
Even though you this does not handle urls exactly like youtube, like `@wb9688` said, I prefer this behaviour. The reason that urls sometimes have additional characters at the end is not something done by youtube, but is something done by accident when writing e.g. messages. So if NewPipe can understand that a `&list=` url comes from a playlist instead of falling back to only using `&v=` like YouTube does it's only a plus. Therefore I'm merging this.

Stypox (approved):
I rebased on dev and run the tests again, everything works fine. Thank you!
~~@ TobiGr I can't merge due to other failing tests, could you do that?~~ nevermind, I can

PR #400, opened 2020-09-29
Do not list YouTube's OTF streams

wb9688:

Ignore streams that are delivered using YouTube's OTF format, as those only work with DASH and not with progressive HTTP.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
@@ -957,6 +957,13 @@ public class YoutubeStreamExtractor extends StreamExtractor {
                 try {
                     ItagItem itagItem = ItagItem.getItag(itag);
                     if (itagItem.itagType == itagTypeWanted) {
+                        // Ignore streams that are delivered using YouTube's OTF format,
+                        // as those only work with DASH and not with progressive HTTP.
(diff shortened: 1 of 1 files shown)


Stypox:
Does this remove the streams causing infinite loading?

wb9688:
@Stypox: Yes, that's what this PR is for.

Stypox:
@TobiGr I'd merge this in 0.20.0 to bring better stability, it is such a small change

Stypox (commented):
Do you think we should add a test for this? Maybe we can do without, if you tested well, since this code is going to be removed anyway in the future

TobiGr:
Stuff like this should be done in the frontend and not the extractor. We can merge this now, but IMO we should extract all itags and provide all stream formats through the extractor. The frontend needs to sort out unsupported formats.

wb9688:
@TobiGr: No. Currently NPE is supposed to only return progressive HTTP streams (and sometimes a torrent as well). These streams aren't progressive HTTP, so NPE shouldn't return them. There is also not really a proper way how NP could figure that out. With my DASH/HLS support PR, there will be a way to specify what kind of streaming method should be used, and then a frontend should skip e.g. DASH streams if they don't support that. However that's not the case yet. After I finish that PR and that gets merged, I'll make a PR extracting all itags based on the metadata that YouTube provides, which could include e.g. 360° videos, which should then also be filtered out by frontends not supporting that.

TobiGr (approved):
Looks good to me.
I do not have time to write tests for this PR, @wb9688 hasn't either. Let's merge this and wait until @wb9688 finishes #367

PR #410, opened 2020-10-11
Fix parsing Soundcloud tracks that contain the term 'sets'

Scrxtchy:

In the current builds of newpipe, if a track contained the term _sets_ in their url, regardless of position, it would be treated as a playlist, causing a ParsingException.
On the assumption that all sets on the soundcloud URL scheme that playlist sets are always within slashes, this should provide an adequate solution

Before the change, the following Exception would be displayed
## Exception
* __User Action:__ requested stream
* __Request:__ https://soundcloud.com/armadamusic/armin-van-buuren-a-state-of-trance-650-warm-up-sets-out-now
* __Content Country:__ AU
* __Content Language:__ en-AU
* __App Language:__ en_AU
* __Service:__ SoundCloud
* __Version:__ 0.20.0

```
org.schabi.newpipe.extractor.exceptions.ParsingException: failed to find pattern ""uri":\s*"https:\/\/api\.soundcloud\.com\/playlists\/((\d)*?)"
	at org.schabi.newpipe.extractor.services.soundcloud.linkHandler.SoundcloudStreamLinkHandlerFactory.getId(SoundcloudStreamLinkHandlerFactory.java:37)
	at org.schabi.newpipe.extractor.linkhandler.LinkHandlerFactory.fromUrl(LinkHandlerFactory.java:57)
	at org.schabi.newpipe.extractor.linkhandler.LinkHandlerFactory.fromUrl(LinkHandlerFactory.java:48)
	at org.schabi.newpipe.extractor.StreamingService.getStreamExtractor(StreamingService.java:261)
	at org.schabi.newpipe.extractor.stream.StreamInfo.getInfo(StreamInfo.java:64)
	at org.schabi.newpipe.util.ExtractorHelper.lambda$getStreamInfo$3(ExtractorHelper.java:116)
	at org.schabi.newpipe.util.-$$Lambda$ExtractorHelper$5fJcha6Sq5APJBLdG6osaJby-mc.call(Unknown Source:4)
	at io.reactivex.Single.subscribe(Single.java:3666)
	at io.reactivex.Single.subscribe(Single.java:3666)
	at io.reactivex.Maybe.subscribe(Maybe.java:4290)
	at io.reactivex.Flowable.subscribe(Flowable.java:14935)
	at io.reactivex.Maybe.subscribe(Maybe.java:4290)
	at io.reactivex.Single.subscribe(Single.java:3666)
Caused by: org.schabi.newpipe.extractor.utils.Parser$RegexException: failed to find pattern ""uri":\s*"https:\/\/api\.soundcloud\.com\/playlists\/((\d)*?)"
	at org.schabi.newpipe.extractor.utils.Parser.matchGroup(Parser.java:72)
	at org.schabi.newpipe.extractor.utils.Parser.matchGroup(Parser.java:61)
	at org.schabi.newpipe.extractor.utils.Parser.matchGroup1(Parser.java:52)
	at org.schabi.newpipe.extractor.services.soundcloud.SoundcloudParsingHelper.resolveIdWithEmbedPlayer(SoundcloudParsingHelper.java:157)
	at org.schabi.newpipe.extractor.services.soundcloud.linkHandler.SoundcloudStreamLinkHandlerFactory.getId(SoundcloudStreamLinkHandlerFactory.java:35)
	... 30 more

```
</details>
<hr>

and now the given URL is parsed correctly
image

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
@@ -25,6 +25,8 @@ import org.schabi.newpipe.extractor.utils.Utils;
 
 import javax.annotation.Nonnull;
 import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.net.URLEncoder;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
@@ -148,12 +150,21 @@ public class SoundcloudParsingHelper {
      *
      * @return the resolved id
      */
-    public static String resolveIdWithEmbedPlayer(String url) throws IOException, ReCaptchaException, ParsingException {
+    public static String resolveIdWithEmbedPlayer(String urlString) throws IOException, ReCaptchaException, ParsingException {
+        // Remove the tailing slash from URLs due to issues with the SoundCloud API
+        if (urlString.charAt(urlString.length() -1) == '/') urlString = urlString.substring(0, urlString.length()-1);
+
+        URL url;
+        try {
+            url = Utils.stringToURL(urlString);
+        } catch (MalformedURLException e){
+            throw new IllegalArgumentException("The given URL is not valid");
+        }
 
         String response = NewPipe.getDownloader().get("https://w.soundcloud.com/player/?url="
-                + URLEncoder.encode(url, "UTF-8"), SoundCloud.getLocalization()).responseBody();
+                + URLEncoder.encode(url.toString(), "UTF-8"), SoundCloud.getLocalization()).responseBody();
         // handle playlists / sets different and get playlist id via uir field in JSON
-        if (url.contains("sets") && !url.endsWith("sets") && !url.endsWith("sets/"))
+        if (url.getPath().contains("/sets/") && !url.getPath().endsWith("/sets"))
             return Parser.matchGroup1("\"uri\":\\s*\"https:\\/\\/api\\.soundcloud\\.com\\/playlists\\/((\\d)*?)\"", response);
         return Parser.matchGroup1(",\"id\":(([^}\\n])*?),", response);
     }
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/linkHandler/SoundcloudStreamLinkHandlerFactory.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/linkHandler/SoundcloudStreamLinkHandlerFactory.java
@@ -30,8 +30,6 @@ public class SoundcloudStreamLinkHandlerFactory extends LinkHandlerFactory {
     @Override
     public String getId(String url) throws ParsingException {
         Utils.checkUrl(URL_PATTERN, url);
-        // Remove the tailing slash from URLs due to issues with the SoundCloud API
-        if (url.charAt(url.length() -1) == '/') url = url.substring(0, url.length()-1);
 
         try {
             return SoundcloudParsingHelper.resolveIdWithEmbedPlayer(url);
--- a/extractor/src/test/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudStreamLinkHandlerFactoryTest.java
+++ b/extractor/src/test/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudStreamLinkHandlerFactoryTest.java
@@ -53,6 +53,7 @@ public class SoundcloudStreamLinkHandlerFactoryTest {
         assertEquals("309689103", linkHandler.fromUrl("https://soundcloud.com/liluzivert/15-ysl").getId());
         assertEquals("309689082", linkHandler.fromUrl("https://www.soundcloud.com/liluzivert/15-luv-scars-ko").getId());
         assertEquals("309689035", linkHandler.fromUrl("http://soundcloud.com/liluzivert/15-boring-shit").getId());
+        assertEquals("259273264", linkHandler.fromUrl("https://soundcloud.com/liluzivert/ps-qs-produced-by-don-cannon/").getId());
         assertEquals("294488599", linkHandler.fromUrl("http://www.soundcloud.com/liluzivert/secure-the-bag-produced-by-glohan-beats").getId());
         assertEquals("294488438", linkHandler.fromUrl("HtTpS://sOuNdClOuD.cOm/LiLuZiVeRt/In-O4-pRoDuCeD-bY-dP-bEaTz").getId());
         assertEquals("294488147", linkHandler.fromUrl("https://soundcloud.com/liluzivert/fresh-produced-by-zaytoven#t=69").getId());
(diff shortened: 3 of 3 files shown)


Stypox (approved):

TobiGr (approved):

wb9688 (commented):
This is OK-ish, to ideally we would parse it through `URL`.

Edit: I think the `!url.endsWith("sets")` isn't needed anymore then though.

Scrxtchy:
> ideally we would parse it through `URL`

I could look into altering the solution to use this method when I wake up in the morning. I'm unfamiliar with the larger scale of the codebase, so I just kept it to it's current behaviour. Should not be a hassle to change

Scrxtchy:
Interestingly enough, users are forbidden from naming their tracks _sets_
image

Just debugging a new revision now

Scrxtchy:
Feels a bit strange jumping between two issues, but if we were under the task that the tailing slash were to be removed, a simple check for the existence of the term `"/sets/"` within the urlPath would be an accurate check for a playlist URL. I am unsure how that issue is going to be steered going forward.

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
```diff
@@ -153,7 +153,7 @@ public static String resolveIdWithEmbedPlayer(String url) throws IOException, Re
         String response = NewPipe.getDownloader().get("https://w.soundcloud.com/player/?url="
                 + URLEncoder.encode(url, "UTF-8"), SoundCloud.getLocalization()).responseBody();
         // handle playlists / sets different and get playlist id via uir field in JSON
-        if (url.contains("sets") && !url.endsWith("sets") && !url.endsWith("sets/"))
+        if (url.contains("/sets/") && !url.endsWith("sets") && !url.endsWith("sets/"))
```
wb9688, line 156: Actually I think I was wrong above and that this should be changed to `url.contains("/sets/") && !url.endsWith("/sets") && !url.endsWith("/sets/")` so that it only doesn't get soundcloud.com/artist/sets URLs, while e.g. soundcloud.com/artist/track-ending-with-sets still works.

wb9688 (changes requested):

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
Scrxtchy: > so that it only doesn't get soundcloud.com/artist/sets

Yeah, this seems to be working, but it will still error out on a widget-related issue, similar to that of https://github.com/TeamNewPipe/NewPipeExtractor/issues/412#issuecomment-708876455 in where it lacks data. Otherwise it seems to be detecting it as a channel type request just fine

Scrxtchy:
I have made changes based on https://github.com/TeamNewPipe/NewPipeExtractor/pull/410#pullrequestreview-508288281 to parse the URL as a Java URL and return a malformed URL exception as per actions used elsewhere in the application

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
Scrxtchy: With changes in f803558d1d3c87f9cefcbec2819615721d71d05a, as URLs need to have their tailing slash removed, it is now impossible for a URL to end in sets/. However I am having a decent struggle determining if ending with `sets` should be checked if `/sets/` is our primary check. This has genuinely caused quite a headache. I have left it in there for safety reasons, but if you think that this would be adequate to be removed, then it shall not be a hassle

Scrxtchy:
Much like changes made in #414, these also need to apply to elements related to this PR which I caught when returning back to this issue. With some extra time I would hopefully be able to test this on a greater scale within the extractor, but this is already getting a bit exhausting.

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/linkHandler/SoundcloudChannelLinkHandlerFactory.java
```diff
@@ -21,6 +21,8 @@ public static SoundcloudChannelLinkHandlerFactory getInstance() {
     @Override
     public String getId(String url) throws ParsingException {
         Utils.checkUrl(URL_PATTERN, url);
+        // Remove the tailing slash from URLs due to issues with the SoundCloud API
+        if (url.charAt(url.length() -1) == '/') url = url.substring(0, url.length()-1);
```
Stypox, line 25: Why not moving this inside `resolveIdWithEmbedPlayer`?

Stypox (changes requested):
Thank you! Is this ready?

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/linkHandler/SoundcloudChannelLinkHandlerFactory.java
Scrxtchy: I think I was just mimicking what I did in 6a70cb9d50132c1cf1bf59478f029b199d65fefc
but I think I can move everything to that instead

I don't really remember my branches that well anymore, but I think I got it

Stypox (commented):
Could you add some tests for this in soundcloud link handler tests?

Scrxtchy:
Testing of adding a slash to the end of the url
https://github.com/TeamNewPipe/NewPipeExtractor/blob/56a6b7c42c23e1e3eddd2ee0640488df7ec7e19b/extractor/src/test/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudStreamLinkHandlerFactoryTest.java#L56

Tests for sets in the track url, this also benefits from not relying on some other artist as a source
https://github.com/TeamNewPipe/NewPipeExtractor/blob/56a6b7c42c23e1e3eddd2ee0640488df7ec7e19b/extractor/src/test/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudStreamLinkHandlerFactoryTest.java#L64

Scrxtchy:
Wait, that's not going to test the sets URL correctly at all, I remembered why that URL parses, but it only does so since the parameters are dropped, I think I'm going to need another artist source for that one

Scrxtchy:
Would the sample used in the OP be acceptable for the repo? I couldn't find anything suitable on users such as `trapcity` and `nocopyrightsounds`, these would have been preferable as to not add another account to be dependant on

B0pol:
> Would the sample used in the OP be acceptable for the repo? I couldn't find anything suitable on users such as `trapcity` and `nocopyrightsounds`, these would have been preferable as to not add another account to be dependant on

https://soundcloud.com/kechuspider-sets-1
Choose any track here

Scrxtchy:
Thanks, 947ce3ee10d3a68947cfd7f3a3df6196f8414174 should be final

Stypox (approved):
Thank you! Looks good

PR #413, opened 2020-10-16
Fix SoundCloud test artist account name

Scrxtchy:

The account used to run tests on no longer has their account name as specified in the testing suite.
image

There are a few other tests that fail on my system, but I believe this to be a problem with the testing utilities, as my system locale is set to Japanese
image

Ideally, these should be run on controlled accounts run by maintainers. This is something I am not putting my name forward for. In the meantime this is a simple solution


TobiGr (approved):
thanks

PR #417, opened 2020-10-18
Use the Java 8 Date/Time API.

Isira-Seneviratne:


wb9688:
D8/R8 supports desugaring up to Java 12. (However AGP only supports Java 8 due to `bootclasspath`, so it's limited to Java 8 for Android code i.e. NewPipe, while you could use Java 12 in non-Android code i.e. NewPipeExtractor.)

Isira-Seneviratne:
> D8/R8 supports desugaring up to Java 12. (However AGP only supports Java 8 due to `bootclasspath`, so it's limited to Java 8 for Android code i.e. NewPipe, while you could use Java 12 in non-Android code i.e. NewPipeExtractor.)

So should I set the JDK version to 12 instead of 8?

Isira-Seneviratne:
The test failures appear to be due to outdated links being used.

Stypox (approved):
Looks good to me, I haven't yet tested though

TobiGr (approved):

TobiGr:
Ah. Well. That was not a good review :laughing: We missed, that the PR does not even build and breaks some tests...

extractor/src/main/java/org/schabi/newpipe/extractor/localization/DateWrapper.java
```diff
@@ -3,30 +3,61 @@
...
+        this(calendar, false);
+    }
+
+    /**
+     * @deprecated Use {@link #DateWrapper(OffsetDateTime, boolean)} instead.
+     */
+    @Deprecated
+    public DateWrapper(@NonNull Calendar calendar, boolean isApproximation) {
+        offsetDateTime = OffsetDateTime.ofInstant(calendar.toInstant(), ZoneOffset.UTC);
+        this.isApproximation = isApproximation;
```
B0pol, line 32: you should use 
```java
this(OffsetDateTime.ofInstant(calendar.toInstant(), ZoneOffset.UTC), isApproximation);
```

Isira-Seneviratne: Shall I commit this in a separate PR?

B0pol: That would be ideal, thanks

PR #419, opened 2020-10-19
Use Objects methods.

Isira-Seneviratne:

**Note:** The `Objects` method calls are desugared when the `minSdkVersion` is below 19 and Java 8 support is enabled, so the changes are compatible with NewPipe Legacy:

Screenshot 2020-10-19 183047


wb9688:
>Note: The Objects method calls are desugared when the minSdkVersion is below 19, so the changes are compatible with NewPipe Legacy.

Only with core library desugaring, right?

Isira-Seneviratne:
> > Note: The Objects method calls are desugared when the minSdkVersion is below 19, so the changes are compatible with NewPipe Legacy.
> 
> Only with core library desugaring, right?

No, I found that it just needs Java 8 support enabled.

Stypox (approved):
Thank you :-D

PR #444, opened 2020-10-26
Remove any reference to decrypt and improve error message

Stypox:

This definitely removes any reference to "decrypt": it was still used to build the custom JS to send to the interpreter, therefore it can be changed if changed everywhere. On my machine tests pass.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
@@ -637,7 +637,7 @@ public class YoutubeStreamExtractor extends StreamExtractor {
     private static final String FORMATS = "formats";
     private static final String ADAPTIVE_FORMATS = "adaptiveFormats";
     private static final String HTTPS = "https:";
-    private static final String DEOBFUSCATION_FUNC_NAME = "decrypt";
+    private static final String DEOBFUSCATION_FUNC_NAME = "deobfuscate";
 
     private final static String[] REGEXES = {
             "(?:\\b|[^a-zA-Z0-9$])([a-zA-Z0-9$]{2})\\s*=\\s*function\\(\\s*a\\s*\\)\\s*\\{\\s*a\\s*=\\s*a\\.split\\(\\s*\"\"\\s*\\)",
@@ -794,7 +794,7 @@ public class YoutubeStreamExtractor extends StreamExtractor {
 
         } catch (IOException e) {
             throw new ParsingException(
-                    "Could load deobfuscation code form restricted video for the Youtube service.", e);
+                    "Could not load deobfuscation code from YouTube video embed", e);
         }
     }
 
@@ -839,8 +839,8 @@ public class YoutubeStreamExtractor extends StreamExtractor {
         final Object result;
         try {
             final ScriptableObject scope = context.initSafeStandardObjects();
-            context.evaluateString(scope, deobfuscationCode, "decryptionCode", 1, null);
-            final Function deobfuscateFunc = (Function) scope.get("decrypt", scope);
+            context.evaluateString(scope, deobfuscationCode, "deobfuscationCode", 1, null);
+            final Function deobfuscateFunc = (Function) scope.get(DEOBFUSCATION_FUNC_NAME, scope);
             result = deobfuscateFunc.call(context, scope, scope, new Object[]{obfuscatedSig});
         } catch (Exception e) {
             throw new DeobfuscateException("Could not get deobfuscate signature", e);


Stypox:
@TheAssassin @B0pol is this ok?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -632,7 +632,7 @@ public String getErrorMessage() {
     private static final String FORMATS = "formats";
     private static final String ADAPTIVE_FORMATS = "adaptiveFormats";
     private static final String HTTPS = "https:";
-    private static final String DEOBFUSCATION_FUNC_NAME = "decrypt";
+    private static final String DEOBFUSCATION_FUNC_NAME = "deobfuscate";
```
TheAssassin, line 635: This doesn't seem okay to me. That name is string matched against the page contents. You shouldn't change it.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -770,7 +770,7 @@ private EmbeddedInfo getEmbeddedInfo() throws ParsingException, ReCaptchaExcepti
 
         } catch (IOException e) {
             throw new ParsingException(
-                    "Could load deobfuscation code form restricted video for the Youtube service.", e);
+                    "Could not load deobfuscation code for YouTube restricted video", e);
```
TheAssassin, line 773: Heh, good catch, that "not". I'd remove the "restricted", though:

```suggestion
                    "Could not load deobfuscation code for YouTube video", e);
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -815,8 +815,8 @@ private String deobfuscateSignature(String obfuscatedSig, String deobfuscationCo
         final Object result;
         try {
             final ScriptableObject scope = context.initSafeStandardObjects();
-            context.evaluateString(scope, deobfuscationCode, "decryptionCode", 1, null);
-            final Function deobfuscateFunc = (Function) scope.get("decrypt", scope);
+            context.evaluateString(scope, deobfuscationCode, "deobfuscationCode", 1, null);
```
TheAssassin, line 818: Same as above. Pretty sure you may not change this. However, you should consider extracting it into a global constant, too.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -815,8 +815,8 @@ private String deobfuscateSignature(String obfuscatedSig, String deobfuscationCo
...
+            final Function deobfuscateFunc = (Function) scope.get(DEOBFUSCATION_FUNC_NAME, scope);
```
TheAssassin, line 819: Good catch.

TheAssassin (changes requested):

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
Stypox: No, that name is not string-matched anywhere. It is only used as the name of the JS function that is manually built to then call the other functions inside the player's JS.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
Stypox: Oh, yes, this is not only used for restricted videos anymore

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
Stypox: `"deobfuscationCode"` is only the name of the JS file to be interpreted, which could be anything and for us has no effect whatsoever on code execution

Stypox:
@TheAssassin I modified the error message to "Could not load deobfuscation code from YouTube video embed", so that "embed" is mentioned. Also, I ran all stream extractor tests on my machine and they all work fine, meaning that changing "encryption" to "deobfuscation" (used for each and every video) has no effect as expected

TheAssassin (approved):
Looks good to me. Let's wait for @B0pol to have a look, too.

B0pol (approved):

PR #450, opened 2020-11-01
create FileUtils

B0pol:

Util class to write file to disk
Can be used to debug and test, for example writing a service's JSON response
(especially useful if the response provided by the service is not documented)


Stypox (changes requested):
Since this file should only be used by tests, I think you should put it under `src/test/java/...`. Other than that, if you tested the methods a little bit, I guess this can be merged without further thinking, as it is only a utility class for testing

B0pol:
I moved the file

Stypox (approved):

Issue #411, opened 2020-10-11
Comments don't show up but don't throw exception


B0pol:
Sometimes comments don't load, and they don't throw exception.

Here are two .har with random videos in trending page; 
[network-log.zip](https://github.com/TeamNewPipe/NewPipeExtractor/files/5360743/network-log.zip)

with the one working, response is complete: response: {responseContext: {,…}, contents: {,…}, continuationContents: {commentSectionContinuation: {,…}},…}

with the one not working, response has not `contents` but has `errors`: 

```JSON
"errors": {
    "error": [
        {
            "domain": "gdata.CoreErrorDomain",
            "code": "INVALID_VALUE",
            "location": "continuation",
            "externalErrorMessage": "Unknown error."
        }
    ]
}
```

PR #456, opened 2020-11-03
create YouTubeCommentsLinkHandlerFactoryTest

B0pol:
and remove invidious test from YouTubeCommentsExtractorTest, because it was just testing if the URL is accepted, then the extractor does the same thing, we don't need to test the same thing twice

no need for the last two, it's only about tests

BTW, the test fail but it's expected. See issue #411


opusforlife2:
Will the invidio.us links work correctly now that it's just a landing page? Maybe they should be changed to Snopyta links?

Stypox (approved):
Thank you!

Stypox:
@opusforlife2 the tests just check if they are accepted, which is ok

TobiGr (approved):

PR #458, opened 2020-11-05
Use Java 9 collection factory methods.

Isira-Seneviratne:

API level 30 added the immutable collection factory methods introduced in Java 9, and these are desugared to support API levels below 30:

Screenshot 2020-11-05 095605

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: java
 jdk:
-  - openjdk8
+  - openjdk9
 
 script:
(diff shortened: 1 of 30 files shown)


Stypox:
What happened?

Isira-Seneviratne:
> What happened?

There was an issue with the Android SDK on Travis when the JDK version was set to 9 (it seems to be because AGP only supports Java 8 due to `bootclasspath`).

Issue #457, opened 2020-11-04
Findings on how comments are fetched on the desktop version of YouTube


FireMasterK:
In basic pseudocode, it can be said like this:
```
String url = "https://www.youtube.com/comment_service_ajax?action_get_comments=1&pbj=1";
url += "&ctoken=" + continuation;
url += "&continuation=" + continuation;
url += "&itct=" + clickTrackingParams;
String body = "session_token=" + XSRF_TOKEN;

Unirest.post(url).header("Content-Type", "application/x-www-form-urlencoded").body(body).asJson().getBody();
```

Here's the reverse engineering I did to find out those:

![image](https://user-images.githubusercontent.com/20838718/98076840-11a63e00-1e95-11eb-9df4-9f7fc00f4814.png)
![image](https://user-images.githubusercontent.com/20838718/98076849-1539c500-1e95-11eb-8fab-1e839fc874fd.png)

This is the function b, used in the previous image, which also shows those query parameters are gotten, according to this code, it seems `itct` is probably an optional field.

![image](https://user-images.githubusercontent.com/20838718/98076854-1834b580-1e95-11eb-9f62-ae465251ceb6.png)

There's also a `session_token` form variable, which from my findings appears to be the `XSRF_TOKEN` in the page source.

FireMasterK:
@Stypox Here's the issue wanted 😁

Issue #455, opened 2020-11-02
comment JSON parsing exception 


eladkarako:
## Exception
* __User Action:__ requested comments
* __Request:__ https://m.youtube.com/watch?v=frO3wRYSHgs
* __Content Country:__ GB
* __Content Language:__ en-GB
* __App Language:__ en
* __Service:__ YouTube
* __Version:__ 0.20.2
* __OS:__ Linux Android 7.1.2 - 25
<details><summary><b>Crash log </b></summary><p>

```
org.schabi.newpipe.extractor.exceptions.ParsingException: Could not parse json data for comments
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeCommentsExtractor.getPage(YoutubeCommentsExtractor.java:97)
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeCommentsExtractor.getInitialPage(YoutubeCommentsExtractor.java:52)
	at org.schabi.newpipe.extractor.utils.ExtractorHelper.getItemsPageOrLogError(ExtractorHelper.java:19)
	at org.schabi.newpipe.extractor.comments.CommentsInfo.getInfo(CommentsInfo.java:40)
	at org.schabi.newpipe.extractor.comments.CommentsInfo.getInfo(CommentsInfo.java:25)
	at org.schabi.newpipe.util.ExtractorHelper.lambda$getCommentsInfo$7(ExtractorHelper.java:155)
	at org.schabi.newpipe.util.-$$Lambda$ExtractorHelper$Xyj84g1UeNZqX9LVYNvWZsdT-kI.call(lambda)
	at io.reactivex.internal.operators.single.SingleFromCallable.subscribeActual(SingleFromCallable.java:44)
	at io.reactivex.Single.subscribe(Single.java:3666)
	at io.reactivex.internal.operators.single.SingleDoOnSuccess.subscribeActual(SingleDoOnSuccess.java:35)
	at io.reactivex.Single.subscribe(Single.java:3666)
	at io.reactivex.internal.operators.maybe.MaybeFromSingle.subscribeActual(MaybeFromSingle.java:41)
	at io.reactivex.Maybe.subscribe(Maybe.java:4290)
	at io.reactivex.internal.operators.maybe.MaybeConcatArray$ConcatMaybeObserver.drain(MaybeConcatArray.java:153)
[...]

PR #477, opened 2020-12-09
Fix comments parsing

vkay94:

This PR should fix the annoying comments parsing error discussed in #455  in NewPipe by converting the `responseBody` into a "valid" format by escaping specific characters. When using the mobile user-agent the `ytInitialData` is formatted with chars as `\x`, which can't be parsed correctly.

Test NewPipe apk (just changed the implementation link to this PR's branch): [app-debug.zip](https://github.com/TeamNewPipe/NewPipeExtractor/files/5666415/app-debug.zip)

PS: Out of some reason I can't reply to the linked issue, so I'm responding to your question here, @TobiGr  : 

>is it possible to convert the unicode characters to utf-8 automatically? Who knows whether they decide to encode other special chars, too.
It there is no simple solution to this, I'd go with your proposed changes,

I've done some research on the headers and so. As said above the reason for this character mess-up is the mobile User-Agent which is required to get the necessary token - without or normal/desktop User-Agent it's formatted well but does contain a different token used for the desktop way, explained in detail in #457 (but the desktop version has also its problems because of POST and CORS protection).

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
@@ -160,8 +160,15 @@ public class YoutubeCommentsExtractor extends CommentsExtractor {
     }
 
     private String findValue(String doc, String start, String end) {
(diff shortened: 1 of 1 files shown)


TobiGr (approved):
thank you!

PR #479, opened 2020-12-11
Extract stream segments for YouTube

vkay94:

This PR extracts the stream segments data provided by YouTube and makes them accessible inside `StreamInfo`. Other services which don't support this data return an empty list when trying getting them. This also applies to streams in YouTube which aren't divided by the uploader.

<table>
    <tr>
        <td>Video chapters of a longer stream (desktop version)</td>
        <td></td>
    </tr>
</table>

The added data class `StreamSegment` contains the following data:
- `String title`:  Title of the chapter set by the uploader
- `int startTimeSeconds`: Starting point of the chapter in seconds
- `String url`: Direct link to the segment (`...?t=`)
- `String previewUrl`: Url to the image at the start position

#### Test against NewPipe
Demo video in NewPipe's player: https://streamable.com/e/3e27wa
Debug app: [app-debug-segments-20201212.zip](https://github.com/TeamNewPipe/NewPipeExtractor/files/5682843/app-debug-segments-20201212.zip)

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
@@ -35,6 +35,7 @@ import org.schabi.newpipe.extractor.stream.Stream;
 import org.schabi.newpipe.extractor.stream.StreamExtractor;
 import org.schabi.newpipe.extractor.stream.StreamInfoItemExtractor;
 import org.schabi.newpipe.extractor.stream.StreamInfoItemsCollector;
+import org.schabi.newpipe.extractor.stream.StreamSegment;
 import org.schabi.newpipe.extractor.stream.StreamType;
 import org.schabi.newpipe.extractor.stream.SubtitlesStream;
 import org.schabi.newpipe.extractor.stream.VideoStream;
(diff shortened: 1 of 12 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -1061,4 +1062,54 @@ public Locale getLanguageInfo() {
...
+                if (panels.getObject(i).getObject("engagementPanelSectionListRenderer")
+                        .getString("panelIdentifier").equals("engagement-panel-macro-markers")) {
+                    segmentArray = panels.getObject(i).getObject("engagementPanelSectionListRenderer")
+                            .getObject("content").getObject("macroMarkersListRenderer").getArray("contents");
+                    break;
+                }
+            }
+
+            if (segmentArray == null) {
+                throw new Exception("Couldn't find valid segments");
```
TobiGr, line 1085: Throwing Exceptions is expensive, because generating the tree of calls can take some time. Please invert the if statement as you are ignoring the thrown Exception anyway.

TobiGr (commented):
thank you!

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
vkay94: Alright, done.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -1061,4 +1062,54 @@ public Locale getLanguageInfo() {
...
+                    final StreamSegment segment = new StreamSegment(title, startTimeSeconds);
+                    if (!previewsArray.isEmpty()) {
+                        // Assume that the thumbnail with the highest resolution is at the last position
+                        segment.setPreviewUrl(previewsArray.getObject(previewsArray.size() - 1).getString("url"));
+                    }
+                    segment.setUrl(getUrl() + "?t=" + startTimeSeconds);
+                    segments.add(segment);
+                }
+            }
+        } catch (Exception ignored) {
```
TobiGr, line 1111: why do you ignore all exceptions?

vkay94: The reason I decided to ignore them is that `getLength` method which could throw a `ParsingException` and because the segments are optional data, not like the title or description.

But thinking about it I should've worked with `.has()`  and return the empty list when it's "obviouly" not present by non-existing JSON-key, otherwise doing the actual parsing with its exceptions throwing if there are errors retrieving the objects etc.

I'd reset the three commits and do it more clearer tomorrow, okay? ;)

opusforlife2:
Nice! Maybe there is a way to integrate this into the background player as well.

vkay94:
I've updated the commits by using `has()` and explicit `ParsingException`  throwing if the text and start time aren't valid. I did some testing with a sample video on my Account by trying various scenarios (unordered timestamps, empty titles, unavailable thumbnails at the beginning) and how the `ytInitialData` is formatted in those cases.

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamSegment.java
```diff
@@ -0,0 +1,69 @@
...
+        this.url = url;
+    }
+
+    @Nullable
+    public String getPreviewUrl() {
+        return previewUrl;
+    }
+
+    public void setPreviewUrl(@Nullable final String previewUrl) {
+        this.previewUrl = previewUrl.replaceAll("\\u0026", "&");
```
TobiGr, line 67: this looks YouTube specific and should therefore be moved from within the function to the function call.

TobiGr (commented):
Thanks. That looks better now!

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamSegment.java
vkay94: Actually, it can be removed completely. The encoding when it's printed in the console is different to the one when saving the html itself from source tab in Chrome/Firefox (I checked it via `System.out.println()`) :')
So I'll wrap it with `fixThumbnailUrl(url)` then like the other `getThumbnailUrl` calls.

TobiGr (approved):

Issue #481, opened 2020-12-12
Automated testing strategy in NewPipeExtractor


XiangRongLin:
Currently all CI builds on dev branch and PR are failing due to test failures. Then figuring out whether those are the usual suspects or caused by my own changes is a hassle and makes CI worthless.

Reason for that is the nature of unit tests relying on external services. In this case it seems Peertube is denying the connection. see https://travis-ci.org/github/TeamNewPipe/NewPipeExtractor/builds/749264136#L450
Since these tests are activly harmful, i would suggest ignoring them all for now.
There are also seem to be some SoundCloud test which are simply broken.

The long term solution is making the unit tests not rely on the external service. The expected websites would need to be saved locally as html and injected into the extractor in the tests. Additional benefits are that tests won't fail because the video/playlist/channel got deleted/changed.

A new problem that is introduced with this are missing tests for API changes of external services. For that a seperate test suite can be added, which only covers this (smoke test). Those should then NOT be run on every PR/Push but maybe nightly/weekly. Those test then also can have big buffers between executions to avoid getting blocked. Those jobs could also update the local html files.
Running without those short term is also fine since most API changes are breaking the extractor. Meaning the NewPipe repo will get bug reports depending on the feature (YT decrypt function, YT comments)

With this the unit tests are covering regressions and the smoke tests are covering whether the feature is actually working.

Thoughts?

---

In short:

1. - [x] ~~Ignore~~ Fix all peertube tests #511 #512
2. - [x] ~~Adjust architecture to allow injecting html into extractor~~ Missed that its already in place
3. - [x] Save websites needed for tests as html locally #482
   Functionality is implemented. See `YoutubeMixPlaylistExtractorTest` for usage
4. - [x] Adjust tests and inject html files into tests #513 #514 
5. - [x] ~~Implement smoke tests~~ Existing tests can be used with my approach. 
6. - [x] Add seperate CI job, which runs smoke tests nightly #516
7. - [x] Add option to mark tests as flaky. So they are not be run against real website in CI process.
  see comment https://github.com/TeamNewPipe/NewPipeExtractor/pull/516#issuecomment-761177779.
8. - [ ] Mark flaky tests

Stypox:
I already did some efforts a while ago at mocking extractor tests on the har-mock branch on my fork, maybe it can be useful: https://github.com/Stypox/NewPipeExtractor/tree/har-mock

XiangRongLin:
@Stypox 
[...]

PR #482, opened 2020-12-15
Mock mix pl tests

XiangRongLin:

My first draft to address the points abount mocking the responses in #481. Not yet merge ready, but before that i wanted to get some opinions on this type of solution.

1. `RecordingDownloader` relays request to the existing DownloaderTestImpl, but saves the request/response pair into a json file inside a folder specific to the test class. This should be done locally when the requests that are made by a test class change.
2. `MockDownloader` relies on the json files created by the `RecordingDownloader`. PR/Push CI jobs should then use this downloader.
3. To change the downloader type during development on the local machine change `DownloaderFactory.DEFAULT_DOWNLOADER`
4. To change the downloader depending on CI jobs, add a system property to gradle commands: `-Ddownloader=ABCD`, where ABCD is one of `DownloaderType`
5. The downloader specified through gradle takes precedence over the one specified inside `DownloaderFactory`. If none is giving it falls back to that one.

Some sitenotes:

- ~~`YoutubeParsingHelper` holds a global state about client version and client key. The first test class that gets executed with the `RecordingDownloader` makes a request and gets that request saved. Other test classes also need that request saved, but since the global state is then already initialized, that request won't be made. Currently I added that request as shared and add them to all `MockDownloader`s~~
- ~~apache-commons-io can be removed if needed. Was needed to clear the directories before `RecordingDownloader` fills them.~~
- Gson is used over nanojson, because nanojson couldn't serialize the responseBody
- I couldn't test if `gradle aggregatedJavadocs -Ddownloader=xyz` works because it won't work on windows due to file encoding. I only tried `gradle extractor:test -Ddownloader=xyz`
- ~~Ignore the first (move package) and last commit (generated jsons) for easier diffs~~
- What is that design pattern called, where you get a specific implementation based on an input. I don't think it's factory


XiangRongLin:
Running `gradle extractor:test -Ddownloader=RECORDING` does use the `RecordingDownloader` (tried with debugger) but does not create the files, which i don't understand.

XiangRongLin:
Sooo, are there any opinions about this approach? 

Because i don't want to put more time into it if the approach is already not good.

Stypox:
I missed this 20 days ago and I don't really have much time to dedicate to NewPipe currently, I may try to look at it the day after tomorrow.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeMixPlaylistExtractorTest.java
```diff
@@ -41,6 +43,7 @@
     private static final String VIDEO_ID = "_AzeUSL9lZc";
     private static final String VIDEO_TITLE =
             "Most Beautiful And Emotional  Piano: Anime Music Shigatsu wa Kimi no Uso OST IMO";
+    private static final String RESOURCE_PATH = "src/test/resources/org/schabi/newpipe/extractor/services/youtube/mix/";
```
Stypox, line 46: The base resource path should be put in some central class, like in `DownloaderFactory`. The root path should in that case be `"src/test/resources/org/schabi/newpipe/extractor/"`

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeMixPlaylistExtractorTest.java
```diff
@@ -50,7 +53,8 @@
 
         @BeforeClass
         public static void setUp() throws Exception {
-            NewPipe.init(DownloaderTestImpl.getInstance());
+            final File file = new File(RESOURCE_PATH + "mix");
+            NewPipe.init(new DownloaderFactory().getDownloader(file.getAbsolutePath()));
```
Stypox, line 57: Maybe you could also add a method to DownloaderFactory replacing these two lines, so that they don't have to be repeated every time. Something like `DownloaderFactory.initNewPipeWithDownloader(RESOURCE_PATH + "mix")`, and put the `.getAbsolutePath()` inside there, too.

extractor/src/test/java/org/schabi/newpipe/downloader/DownloaderFactory.java
```diff
@@ -0,0 +1,30 @@
...
+
+        switch (type) {
+            case REAL:
+                return DownloaderTestImpl.getInstance();
+            case MOCK:
+                return new MockDownloader(path);
+            case RECORDING:
+                return new RecordingDownloader(path);
+            default:
+                throw new UnsupportedOperationException("Unknown downloader type: " + type.toString());
```
Stypox, line 27: Can this even happen? Doesn't Java disallow this impossible cases, since there are only three possible instances of the enum?

extractor/build.gradle
```diff
@@ -9,4 +9,6 @@ dependencies {
 
     testImplementation 'junit:junit:4.13'
     testImplementation "com.squareup.okhttp3:okhttp:3.12.11"
+    testImplementation 'com.google.code.gson:gson:2.8.6'
```
Stypox, line 12: Are there limitations with `nanojson`?

extractor/src/test/java/org/schabi/newpipe/downloader/MockDownloader.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public MockDownloader(@Nonnull String path) throws IOException {
+        this.path = path;
+        this.mocks = new HashMap<>();
+        File folder = new File(path);
+        for (File file : folder.listFiles()) {
+            final FileReader reader = new FileReader(file);
+            final TestRequestResponse response = new GsonBuilder()
+                    .create()
+                    .fromJson(reader, TestRequestResponse.class);
```
Stypox, line 30: I am not sure, but maybe here you should catch errors, so that if we put some files of a different format in the same folder there are no issues (we may want to do that at some point, maybe)

Stypox (changes requested):
I like this approach, it allows for expansion and is simple to migrate to. Thank you for your dedication :-D

extractor/build.gradle
XiangRongLin: I don't quite remember what it was, but it did not work with nanojson

extractor/src/test/java/org/schabi/newpipe/downloader/MockDownloader.java
XiangRongLin: I resolved this by prefixing the files and only writing/reading from those.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeMixPlaylistExtractorTest.java
XiangRongLin: I found that creating the file object itself is unnecessary. without that line its already a single liner.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeMixPlaylistExtractorTest.java
XiangRongLin: done

XiangRongLin:
@Stypox i rebased on top of dev.
The new changes are seperate commits for easier reviewing. For the actual changes see commit history.
My first sitenote about the global state in `YoutubeParsingHelper` i just band-aid style fixed it with a reset method.

I will regenerate the mock files, when the pr is ready to be merged.

extractor/src/test/java/org/schabi/newpipe/downloader/DownloaderFactory.java
XiangRongLin: It can not, but the compiler doesn't know it.

First SO hit: https://stackoverflow.com/q/5013194

extractor/src/test/java/org/schabi/newpipe/downloader/RecordingDownloader.java
```diff
@@ -0,0 +1,58 @@
...
+        Path path = Paths.get(stringPath);
+        File folder = path.toFile();
+        if (folder.exists()) {
+            for (File file : folder.listFiles()) {
+                if (file.getName().startsWith(RecordingDownloader.FILE_NAME_PREFIX)) {
+                    file.delete();
+                }
+            }
+        }
+        Files.createDirectories(path);
```
XiangRongLin, line 37: Todo for me

TobiGr:
> 1. `RecordingDownloader` relays request to the existing DownloaderTestImpl, but saves the request/response pair into a json file inside a folder specific to the test class. This should be done locally when the requests that are made by a test class change.
2. `MockDownloader` relies on the json files created by the `RecordingDownloader`. PR/Push CI jobs should then use this downloader.
3. To change the downloader type during development on the local machine change `DownloaderFactory.DEFAULT_DOWNLOADER`
4. To change the downloader depending on CI jobs, add a system property to gradle commands: `-Ddownloader=ABCD`, where ABCD is one of `DownloaderType`

Can you put that info to a place where it can be found later? [To be honest, I don't know a good place for that]

XiangRongLin:
@TobiGr I added to the classes/methods themselves. Anywhere else it would just get out of date.

Only thing thats missing is that CI jobs should use `MockDownloader`, but i would add that to the CI job itself.

extractor/build.gradle
Stypox: Mmmh ok, this is anyway `testImplementation` so no need to bother about it ;-)

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
```diff
@@ -353,6 +353,11 @@ public static String getKey() throws IOException, ExtractionException {
         return key;
     }
 
+    static void resetClientVersionAndKey() {
```
Stypox, line 356: Maybe add a comment saying that this is only to be used for tests.

Stypox (changes requested):
Almost ready, imo

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
XiangRongLin: I wouldn't add it, since android studio can already show you where a method is used. In this case it's only tests.
Additionally i did not make it public to only allow usage inside the extractor

Stypox: Ok, but I'd like to be able to tell just by looking at it that the method should not be used except for tests

XiangRongLin: added

extractor/build.gradle
XiangRongLin: My initial idea behind saving it as json (human readable) was to be able to search through it. Because i remember that when i first made the mix extractor i saved the request with postman to do just that.

If that proves to not be needed, then `ObjectInputStream` `ObjectOutputStream` can be used.

Stypox: No, we can stick to json, human readability is useful

Stypox (approved):
Thank you, this now looks good to me. Do you still want to implement something else or is this ready @XiangRongLin ? @TobiGr I'll let you merge, since you also reviewed this PR.

XiangRongLin:
I'll just recreate the json files later. Then it can be merged

Stypox (approved):
Looking good to me, and the relevant tests succeed. Great :-D

Issue #485, opened 2020-12-17
Support for 4320p60 / 8K


FireMasterK:
title,
The adaptive stream is as follows:
```
{
"itag":272,
"url":"https://r5---sn-npoe7ner.googlevideo.com/videoplayback?expire=1608236718&ei=TmrbX_StM8364-EP2aiVyAY&ip=94.198.43.77&id=o-AFxDb1xUEfU_bkofMhgL0SaNnUGf60hQvJlnMu5viQFM&itag=272&aitags=133%2C134%2C135%2C136%2C160%2C242%2C243%2C244%2C247%2C272%2C278%2C298%2C299%2C302%2C303%2C308%2C315%2C330%2C331%2C332%2C333%2C334%2C335%2C336%2C337%2C394%2C395%2C396%2C397%2C398%2C399&source=youtube&requiressl=yes&mh=XR&mm=31%2C26&mn=sn-npoe7ner%2Csn-i3belnlz&ms=au%2Conr&mv=m&mvi=5&pl=24&initcwndbps=321250&vprv=1&mime=video%2Fwebm&ns=Fqk8atfteNvT4Q1N_6mo5msF&gir=yes&clen=849820003&dur=268.966&lmt=1576096789562597&mt=1608214613&fvip=5&keepalive=yes&c=WEB&txp=5311222&n=9HlmABm6R8Kg_Z6mj&sparams=expire%2Cei%2Cip%2Cid%2Caitags%2Csource%2Crequiressl%2Cvprv%2Cmime%2Cns%2Cgir%2Cclen%2Cdur%2Clmt&sig=AOq0QJ8wRQIgWbVaNbH-ku1MLBnKpYzXc0M7B9YV7VXGtgmhpcBNEcQCIQDYY2uCaOVhdfm42fhjXEYQCqXxgSgEmFw_BfUrZYNnVA%3D%3D&lsparams=mh%2Cmm%2Cmn%2Cms%2Cmv%2Cmvi%2Cpl%2Cinitcwndbps&lsig=AG3C_xAwRQIgUG6ap635XRkXNE7hStW_yOWeeJcdcz4DW6M2CTcHUr8CIQCy-hLeQgk6qLCQKUzaJhs6BTg3fRc2F6WKcbUrh46rMQ%3D%3D",
"mimeType":"video/webm; codecs="vp9"",
"bitrate":43338259,
"width":7680,
"height":4320,
"initRange":{
"start":"0",
"end":"219"
[...]

Issue #39, opened 2017-11-06
HDR Support


karyogamy:
Although tunneled playback is enabled on the player in the client, without proper extractor support, HDR will not work. 

Here's a snippet from youtube-dl:
> youtube-dl.exe -F https://www.youtube.com/watch?v=WW2DKBGCvEs
[youtube] WW2DKBGCvEs: Downloading webpage
[youtube] WW2DKBGCvEs: Downloading video info webpage
[youtube] WW2DKBGCvEs: Extracting video information
[youtube] WW2DKBGCvEs: Downloading MPD manifest
[info] Available formats for WW2DKBGCvEs:
format code  extension  resolution note
139          m4a        audio only DASH audio   48k , m4a_dash container, mp4a.40.5@ 48k (22050Hz), 683.89KiB
249          webm       audio only DASH audio   55k , opus @ 50k, 771.41KiB
250          webm       audio only DASH audio   80k , opus @ 70k, 1.07MiB
140          m4a        audio only DASH audio  127k , m4a_dash container, mp4a.40.2@128k (44100Hz), 1.78MiB
171          webm       audio only DASH audio  132k , vorbis@128k, 1.73MiB
251          webm       audio only DASH audio  151k , opus @160k, 2.03MiB
160          mp4        256x144    DASH video   47k , avc1.4d400c, 24fps, video only, 470.93KiB
278          webm       256x144    144p   97k , webm container, vp9, 24fps, video only, 1.24MiB
330          webm       256x144    144p HDR  147k , vp9.2, 24fps, video only, 1.73MiB
242          webm       426x240    240p  173k , vp9, 24fps, video only, 1.82MiB
133          mp4        426x240    DASH video  182k , avc1.4d4015, 24fps, video only, 1.83MiB
331          webm       426x240    240p HDR  241k , vp9.2, 24fps, video only, 2.96MiB
134          mp4        640x360    DASH video  280k , avc1.4d401e, 24fps, video only, 2.81MiB
243          webm       640x360    360p  364k , vp9, 24fps, video only, 3.96MiB
332          webm       640x360    360p HDR  457k , vp9.2, 24fps, video only, 5.80MiB
135          mp4        854x480    DASH video  631k , avc1.4d401e, 24fps, video only, 6.27MiB
244          webm       854x480    480p  662k , vp9, 24fps, video only, 7.10MiB
333          webm       854x480    480p HDR  864k , vp9.2, 24fps, video only, 10.99MiB
[...]

PR #488, opened 2020-12-19
8K/HDR/360-degree videos - Add dynamic itag support.

FireMasterK:

This pr adds support for dynamic itag loading and therefore 8K/HDR/360-degree videos.


FireMasterK:
This obviously requires extensive testing, a lot of hardcoded itags aren't covered but I have never seen them in the wild, do send me such videos here so I can handle those as well.

Test APK: [app-debug.zip](https://github.com/TeamNewPipe/NewPipeExtractor/files/5720089/app-debug.zip)

FireMasterK:
Closes https://github.com/TeamNewPipe/NewPipeExtractor/issues/485, https://github.com/TeamNewPipe/NewPipe/issues/5118, https://github.com/TeamNewPipe/NewPipe/issues/1892, https://github.com/TeamNewPipe/NewPipeExtractor/issues/39, https://github.com/TeamNewPipe/NewPipe/issues/2842.

opusforlife2:
You need to add the 'closes' line in the OP for Github to link the issues. Doesn't work in a comment.

FireMasterK:
> You need to add the 'closes' line in the OP for Github to link the issues. Doesn't work in a comment.

Changed 👍

opusforlife2:
You also need the 'closes' keyword in front of each link. :P

FireMasterK:
So after some testing, I found out that HDR webm streams and 8K videos won't play on exoplayer (on my device atleast).

Any suggestions on what should be done on the NewPipe side?

Stypox:
@FireMasterK you could take a look at https://exoplayer.dev/supported-formats.html

fareszr:
> So after some testing, I found out that HDR webm streams and 8K videos won't play on exoplayer (on my device atleast).
> 
> Any suggestions on what should be done on the NewPipe side?

So i tested th apk, on a 4k video with HDR.
The video player crashes when trying to play any resolution higher than 1080p with HDR(mpeg).
And when playing in 1080 mpeg-4 HDR, its playing in like 5fps with everything kind of grey.

Now when switching to Webm it works(on all resolutions) but i barely notice the differences. Tried multiple videos and its the same issue.

FireMasterK:
> In addition to Android’s platform decoders, ExoPlayer can also make use of software decoder extensions. These must be manually built and included in projects that wish to make use of them. We currently provide software decoder extensions for AV1, VP9, FLAC, Opus and FFmpeg.

It looks like this may be required for unsupported devices that don't support AV1 and VP9(.2).

> Device implementations must support dynamic video resolution and frame rate switching through the standard Android APIs within the same stream for all VP8, VP9, H.264, and H.265 codecs in real time and **up to the maximum resolution supported by each codec on the device**. src: https://developer.android.com/guide/topics/media/media-formats#core

This suggests that software encoding is required for higher resolution.

Thoughts? @Stypox 
Is there something I missed?

Stypox:
A while ago it was decided not to include code requiring NDK into NewPipe, idk if this relates

FireMasterK:
What's the solution then? I'm not sure what can be done on devices with no hardware decoding..

orangpelupa:
Hiya, the "bad contrast" when playing HDR youtube video thru the debug newpipe posted above could be due to

1. It displays HDR to SDR directly, without proper tone mapping. Thus it looks to have a bad contrast. On the other hand if you play a HDR video thru 3rd party player (VLC, mx player, etc) it'll be displayed in SDR with proper tone mapping. So you loss its "white sky is brighter than a white paper" but the contrast and colors looks fine.
2. 
2. It did not switch the display mode from SDR to HDR, thus it was displayed in SDR. You can easily check this by swiping down the notification bar or opening the recent menu. In HDR mode (invoked by playing an HDR video in a stock Samsung video player), those overlays/screens will have wonky colors.

TobiGr:
We need to implement something like "supported formats". Pass a List of supported formats to [NewPipe](https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/main/java/org/schabi/newpipe/extractor/NewPipe.java) when initializing the extractor (empty list => all formats are supported).

Stypox:
> We need to implement something like "supported formats". Pass a List of supported formats to NewPipe when initializing the extractor (empty list => all formats are supported).

I'd go with the opposite: let the NewPipe app filter out the unsupported formats

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/ItagItem.java
```diff
@@ -57,6 +57,9 @@
             new ItagItem(137, VIDEO_ONLY, MPEG_4, "1080p"),
             new ItagItem(299, VIDEO_ONLY, MPEG_4, "1080p60", 60),
             new ItagItem(266, VIDEO_ONLY, MPEG_4, "2160p"),
+            new ItagItem(402, VIDEO_ONLY, MPEG_4, "4320p"), // can be 4320p60 as well
```
Co0olboi, line 60:
```suggestion
            new ItagItem(402, VIDEO_ONLY, MPEG_4, "4320p"), // can be 4320p60 as well
```

Co0olboi (approved):

Co0olboi:
- [ ]

TiA4f8R:
Sorry @Co0olboi, your suggestion was indentical to what FireMasterK made, and GitHub dosen't allow to hide the comment, so I deleted it before I saw that I can dismiss your review, sorry :(.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/ItagItem.java
```diff
@@ -66,28 +69,19 @@
...
-        for (ItagItem item : ITAG_LIST) {
-            if (itag == item.id) {
-                return true;
-            }
-        }
-        return false;
-    }
-
+    @Deprecated
     public static ItagItem getItag(int itagId) throws ParsingException {
```
Stypox, line 85: Why not removing this function and static itags directly if they are not used anymore? This is an internal class, so it won't do any harm to backwards compatibility

Stypox (changes requested):
Add `final` when possible ;-)

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/ItagItem.java
FireMasterK: They're actually used in the DashMpdParser, that's why, the qualityLabel isn't provided there iirc.

Stypox: Mmmh, then why mark it deprecated if it will be still used?

FireMasterK: I was questioning whether static itags should really be used/encouraged. Even though the qualityLabel isn't passed, I think it  can be removed in the future with something like "%heightP"

Stypox: Mmmh ok, then leave it with `@Deprecated`

Stypox (approved):
Looks good to me, though my opinion on this is not much valuable

FireMasterK:
Note: There are some changes in NewPipe to be made before merging this. (I need some help with Exoplayer for this)

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/ItagItem.java
```diff
@@ -56,6 +56,9 @@
             new ItagItem(137, VIDEO_ONLY, MPEG_4, "1080p"),
             new ItagItem(299, VIDEO_ONLY, MPEG_4, "1080p60", 60),
             new ItagItem(266, VIDEO_ONLY, MPEG_4, "2160p"),
+            new ItagItem(402, VIDEO_ONLY, MPEG_4, "4320p"), // can be 4320p60 as well
+            new ItagItem(571, VIDEO_ONLY, MPEG_4, "4320p"), // can be 4320p60 HDR as well (1La4QzGeaaQ)
+            new ItagItem(402, VIDEO_ONLY, MPEG_4, "4320p60"),
```
litetex, line 61: @FireMasterK 
Why is there no additional parameter for the 60fps? 
Like above in ``new ItagItem(299, VIDEO_ONLY, MPEG_4, "1080p60", 60)``

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/ItagItem.java
```diff
@@ -97,6 +91,52 @@ public static ItagItem getItag(int itagId) throws ParsingException {
...
+            if (fileType.equals("mp4") && (codec.startsWith("m4a") || codec.startsWith("mp4a")))
+                format = M4A;
+            if (fileType.startsWith("webm") && codec.equals("opus"))
+                format = WEBMA_OPUS;
+        }
+
+        if (itagType == null || format == null)
+            throw new ParsingException("Unknown mimeType: " + mimeType);
+
+        return itagType == AUDIO ? new ItagItem(itagId, itagType, format, Math.round(averageBitrate / 1024f)) : new ItagItem(itagId, itagType, format, qualityLabel, fps);
```
litetex, line 137: Code-style technically I would clearly break this up into multiple lines
I can't even see the line ending when reviewing this here on GitHub 😆

May also be relevant for the header on Line 94

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/ItagItem.java
```diff
@@ -66,28 +69,19 @@
...
-    public static boolean isSupported(int itag) {
-        for (ItagItem item : ITAG_LIST) {
-            if (itag == item.id) {
-                return true;
-            }
-        }
-        return false;
-    }
-
+    @Deprecated
```
litetex, line 84: A reason for deprecating this/what to use instead in JavaDoc would be great here 😄 

I also noticed that https://github.com/TeamNewPipe/NewPipeExtractor/blob/6db4bea8ca51ac40a8f831527b274ad50895070e/extractor/src/main/java/org/schabi/newpipe/extractor/utils/DashMpdParser.java#L150 still relies on that code, so maybe some fixing has to be done there

Stypox: See https://github.com/TeamNewPipe/NewPipeExtractor/pull/488#discussion_r688385684

litetex:
I also tested this together with #706 on an experimental branch with the NewPipe app and got the same problems as stated above (black screen on resolutions >1080 etc) on my emulator - Pixel 3a Android 11.0/API 30.

Personally I would also currently not recommend implementation:
* <10% of all android devices even support AV1, which is [*optionally* (decoder) shipped with Android 10+](https://developer.android.com/guide/topics/media/media-formats#video-codecs)
grafik
grafik
* there could be certain changes from YT's side - this PR is now open since about 9 months; in likely 2-4 years we will see a noteworthy market share of Android 10+ with >20-40% - that may render it useless and it has to be (completely) reworked again

For now, I'll stay out of this topic as it's too subject-heavy.

FireMasterK:
The solution obviously would be to hide resolutions that aren't supported by NewPipe/Exoplayer.

I haven't made any changes in the NewPipe app yet, which is essential to even consider merging this.

From prior conversations, Stypox said it was decided to not allow NDK/software decoding for unsupported devices, so selectively filtering formats on the app is the only suitable solution.

0Karakurt0:
Mmmm... Where can I find new testing .apk or sources AND instruction?
One in the beginning of the thread is quite outdated and can't load videos fast enough to viev HDR

Stypox:
@0Karakurt0 @FireMasterK I rebased [here](https://github.com/Stypox/NewPipeExtractor/tree/pr488) but couldn't push here since I don't have permission. Anyway, here is the apk built upon the latest version of the app and the extractor: [app-debug.zip](https://github.com/TeamNewPipe/NewPipeExtractor/files/7576673/app-debug.zip)

0Karakurt0:
> here is the apk built upon the latest version of the app

Mmm. Idk if I'm lacking hardware, but I belive that on earlier version I did not had that issue.
I am expiriencing frame skipping and video slowdowns on 1080p HDR.
Hardware is Mi 9T, I'll write later if the same issue is present on earlier build. And add more testing overall, tho I gave no idea what information would be helpful in solving this.
May the OS have a part in preformance penalty? I'm running DoT OS on my phone

0Karakurt0:
Another strange issue noticed is that sometimes audio is doubled. I am using Bluetooth headphones and sometimes during the queue change phone's loudspeakers get activated in parallel to headphones

0Karakurt0:
Just right now same doubled sound happened in main build, so it's either not NewPipe specific or a bug from mainstream.
What it had in common was HDR, I was watching downloaded HDR video in Just Player right before opening a video in NewPipe.

DavidH-2022:
Any updates on HDR?

Valmar33:
@FireMasterK Is there a more recent test APK that we can test? Cheers.

synchrone:
Here's an 
[app-debug.zip](https://github.com/user-attachments/files/18970192/app-debug.zip) built with this PR and with aa673635530e131c48762c9eb8b610bea9249495 merged in. 
In my testing it does not show HDR in player quality dropdown, but at least lets you download the HDR video.

PR #494, opened 2020-12-22
[media.ccc.de] Add "recent" kiosk

TobiGr:

Adds the recent page which can be found at https://media.ccc.de/recent

Tests are missing. I don't have time to add them now.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCRecentKioskExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCRecentKioskExtractor.java
@@ -1,22 +1,34 @@
 package org.schabi.newpipe.extractor.services.media_ccc.extractors;
 
 import com.grack.nanojson.JsonObject;
+
+import org.schabi.newpipe.extractor.ServiceList;
+import org.schabi.newpipe.extractor.channel.ChannelExtractor;
+import org.schabi.newpipe.extractor.exceptions.ExtractionException;
 import org.schabi.newpipe.extractor.exceptions.ParsingException;
 import org.schabi.newpipe.extractor.localization.DateWrapper;
 import org.schabi.newpipe.extractor.services.media_ccc.linkHandler.MediaCCCConferenceLinkHandlerFactory;
 import org.schabi.newpipe.extractor.stream.StreamInfoItemExtractor;
 import org.schabi.newpipe.extractor.stream.StreamType;
 
-import javax.annotation.Nullable;
+import java.io.IOException;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
+import java.util.Map;
+
+import javax.annotation.Nullable;
+
+import static org.schabi.newpipe.extractor.utils.Utils.isNullOrEmpty;
 
 public class MediaCCCRecentKioskExtractor implements StreamInfoItemExtractor {
 
     private final JsonObject event;
+    private final Map<String, String> conferenceNames;
 
-    public MediaCCCRecentKioskExtractor(final JsonObject event) {
+    public MediaCCCRecentKioskExtractor(final JsonObject event,
+                                        final Map<String, String> conferenceNames) {
         this.event = event;
+        this.conferenceNames = conferenceNames;
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCRecentKioskExtractor.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Override
+    public long getViewCount() throws ParsingException {
+        return event.getInt("view_count");
+    }
+
+    @Override
+    public String getUploaderName() throws ParsingException {
+        return "";
```
B0pol, line 58: event.getArray("persons").getString(0)

TobiGr: person contains people who participated in a talk / video. That's more like a tag.

The uploader is in our case the conference, but we only have the URL for that conference and making 100 requests to get conference names is a little out of scope.

TobiGr: From the [readme](https://github.com/voc/voctoweb):
>  Every talk (alias **event**, in other systems also called lecture or session) is assigned to exactly one **conference** (e.g. the _congress_ or lecture series like _datengarten_ or _openchaos_) and consists of multiple files alias **recordings**. These files can be video or audio recordings of the talk in different formats and languages (live-translation), subtitle tracks as srt or slides as pdf.

extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCRecentKioskExtractor.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Override
+    public boolean isAd() throws ParsingException {
+        return false;
+    }
+
+    @Override
+    public long getDuration() throws ParsingException {
+        return event.getInt("duration");
```
B0pol, line 48: do you know what's the difference between length and duration?

extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCRecentKioskExtractor.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Override
+    public String getUploaderName() throws ParsingException {
+        return "";
+    }
+
+    @Override
+    public String getUploaderUrl() throws ParsingException {
+        return event.getString("conference_url");
```
B0pol, line 63: this returns an api link. Are you sure it is ok?
Without an api request you can build the uploader url, get the last part from this link and add it here: `https://media.ccc.de/c/<uploader_id>`

extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCRecentKioskExtractor.java
TobiGr: no. both values are the same in all 100 entries.

extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCRecentKioskExtractor.java
TobiGr: done the problem is still that the uploader is not shown, because we don't have the uploader name / conference name.
We could make an API request to get the conference name and store the value in some static map to reduce the number of requests.

TobiGr: I now get the conference URL.
Disadvantage: ~560kB are downloaded, 7 requests made. The number of requests and the size of the retrieve JSON documents can vary extremely. This is because every conference documents contains massive data about its events.
@Stypox @B0pol @TheAssassin

PR #496, opened 2020-12-24
Migrate to GitHub actions from Travis.

FireMasterK:

I have not tested this yet, but everything should be fine I think.

@TheAssassin


.github/workflows/docs.yml
```diff
@@ -0,0 +1,33 @@
+name: Build and deploy JavaDocs
+
+on:
+  push:
+    branches:
+      - dev
```
TobiGr, line 6: Docs should only change when releasing a new version
```suggestion
      - master
```

TobiGr (commented):
Thank you. Code looks good to me but I did not test it yet.

.github/workflows/docs.yml
FireMasterK: Changed 👍

TobiGr:
Please change the badge in README.md
See TeamNewPipe/NewPipe#5278

FireMasterK:
Updated the README badge.

TobiGr:
I used that code in my fork to test it. See https://github.com/TobiGr/NewPipeExtractor/runs/1630857048

As you can see, the workflow is cancelled as soon as the test job fails. That is bad, because there are always failing tests.

TobiGr:
@FireMasterK Can you give us access to your branch so we can apply some fixes and merge afterwards? (E.g. some JDoc needs to be fixed before deployment works).
You can give us access to that branch by ticking a checkbox in the right sidebar.

FireMasterK:
I don't have that option, likely due to my fork being in an org, I'll close this PR in favor of #506 which has those fixes 😃

PR #505, opened 2020-12-30
[media.ccc.de] Fix NPE in search results if they contain a future talk

TobiGr:

If the `release_date` is null, the talk is going to be held in the future. Thus the result's StreamInfo will empty video and audio stream lists. For this reason, do not include future talks in the search results.

Fixes TeamNewPipe/NewPipe#5306

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCParsingHelper.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCParsingHelper.java
@@ -28,9 +28,9 @@ public final class MediaCCCParsingHelper {
         }
     }
 
-    public static boolean isLiveStreamId(final String url) {
+    public static boolean isLiveStreamId(final String id) {
         final String pattern = "\\w+/\\w+";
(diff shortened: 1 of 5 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCParsingHelper.java
```diff
@@ -28,9 +28,9 @@ public static OffsetDateTime parseDateFrom(final String textualUploadDate) throw
         }
     }
 
-    public static boolean isLiveStreamId(final String url) {
+    public static boolean isLiveStreamId(final String id) {
         final String pattern = "\\w+/\\w+";
```
Stypox, line 32: This pattern should be compiled and made `static final`

Stypox (changes requested):
Looks good

extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCParsingHelper.java
TobiGr: Done

Stypox (approved):

PR #506, opened 2020-12-31
 Migrate to GitHub actions from Travis

TobiGr:
 
This PR supersedes #496 and therefore closes #496.
Thanks to @FireMasterK for doing the initial (and most) work!

I fixed the JDoc deployment and updated the link for Java SE components.
This PR targets `master` for technical reasons (such as getting the JDoc deployed asap).

### To Do / Needs to be fixed later
- [ ] Cache builds although the tests fail.


FireMasterK:
You can't really fix the caching on a build failure, see: https://github.com/actions/cache/issues/92.

However, a workaround is to fork the cache action and then do something like [gerbal/always-cache](https://github.com/gerbal/always-cache), but that's going to require some maintenance work.

However, it shouldn't be too big of an issue considering these builds last 2 minutes.

TobiGr:
Yes, you are right.

I'll merge this now. We can improve this later if necessary

PR #507, opened 2021-01-04
Add durationPerFrame to Frameset

vkay94:

This PR adds `durationPerFrame` field to the existing `Frameset` class. This property is necessary to calculate the correct frame depending on the timestamp.  Before there was no way to crop the correct frame bounds from the storyboard ("`framesPerPageX` x `framesPerPageY`"-matrix). 

#### Short example with this change to calculate the correct frame coordinates (pseudo code):
-- Goal: get frame bounds for 3:25:500 (= 180.500 milliseconds)´
-- Given: storyboard-Urls in 5x5-matrix, duration per frame = 5.000, frame height = 160 and frame width = 90

=> `180.500 / 5.000` = 36.1 => 37. Frame
=> `(37 - 1 * 25)` = 12 => 12. frame in 2. storyboard
=> `(12 - 2 * 5)` = 2 => 3. row, 2. column => row index = 2, column index = 1

=> Bounds / rectangle of the frame at the position: `left top (x, y) = (0 + 1 * 160 = 160, 0 + 2 * 90 = 180)`
=> `(left, top, right, bottom) = (160, 180, 320, 270)`

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/stream/Frameset.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/stream/Frameset.java
@@ -8,12 +8,14 @@ public final class Frameset {
     private int frameWidth;
     private int frameHeight;
     private int totalCount;
+    private int durationPerFrame;
     private int framesPerPageX;
     private int framesPerPageY;
 
-    public Frameset(List<String> urls, int frameWidth, int frameHeight, int totalCount, int framesPerPageX, int framesPerPageY) {
+    public Frameset(List<String> urls, int frameWidth, int frameHeight, int totalCount, int durationPerFrame, int framesPerPageX, int framesPerPageY) {
         this.urls = urls;
         this.totalCount = totalCount;
+        this.durationPerFrame = durationPerFrame;
         this.frameWidth = frameWidth;
         this.frameHeight = frameHeight;
         this.framesPerPageX = framesPerPageX;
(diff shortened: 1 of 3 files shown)


TobiGr:
Thank you. Could you put that info into to the class JDoc so the class is easier to use?

vkay94:
@TobiGr Okay, I'll add it. Actually, I'd prefer an open/public default implementation - something like

```java 
public int[] getFrameBoundsAt(long timestamp)
```

where the `int`-array  returns `left, top, right, bottom` in that order. It would reduce code on the client-side.

I just checked the other services and PeerTube doesn't support frames and media.ccc.de uses the same logic as YouTube (grid) (but here you've got to load an additional `vtt` -file to get the count and duration per frame).

Url: https://static.media.ccc.de/media/events/rc3/150-8e6fd90f-f358-518f-95dc-19dbec1dcaea.thumbnails.vtt
Contents: https://pastebin.com/tSgFNBrV

What do you think?

TobiGr:
Sounds good.

vkay94:
I've added the default implementation. It should work correctly (even if the "testing" has basically been with the eyes :') )

extractor/src/main/java/org/schabi/newpipe/extractor/stream/Frameset.java
```diff
@@ -61,4 +63,48 @@ public int getFrameWidth() {
...
+     * @return duration per frame in millis
+     */
+    public int getDurationPerFrame() {
+        return durationPerFrame;
+    }
+
+    /**
+     * Returns the information for the frame at stream position.
+     *
+     * @param position Position of the stream
```
TobiGr, line 77: position in seconds?

vkay94: Yeah, it's definitely better ;)

TobiGr (approved):
Can be merged after this small change

PR #509, opened 2021-01-10
Add new exceptions to be able to display different error messages in apps

TiA4f8R:
This PR add six new exceptions to be able to display different error messages in NewPipe when user tries to play an unplayable content. All of these exceptions extend the `ContentNotAvailableException`.
- `AgeRestrictedContentException`, for age-restricted content which has no video or audio streams.
- `GeographicRestrictionException`, for georestricted content (used for SoundCloud and YouTube georestricted videos, not working on YouTube Music georestricted tracks because YouTube shows "This video is not available")
- `PaidContentException`, for paid content like YouTube movies or YouTube videos that require a channel subscription
- `PrivateContentException`, for private YouTube videos (can be used for other services)
- `SoundCloudGoPlusContentException`, for SoundCloud Go+ tracks
- `YoutubeMusicPremiumContentException`, for YouTube videos restricted to YouTube Music Premium members

- [x] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API. TeamNewPipe/NewPipe#5385


TiA4f8R:
Here is my code which isn't working:

https://github.com/TeamNewPipe/NewPipeExtractor/blob/ea780f268141096bbe254edda2fcca4125718e6d/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java#L276-L298

Can someone explain me what is wrong with my code (`StreamExtractException: Could not get any stream. See error variable to get further details.`)?

Stypox:
That's normal for soundcloud. It happens since there there are no valid audio formats to play

TiA4f8R:
@Stypox So, what I should do?

Stypox:
You could do that in another PR and not touch anything about streams here

TiA4f8R:
@Stypox Done. Code is ready for review.

extractor/src/main/java/org/schabi/newpipe/extractor/exceptions/GeographicRestrictionException.java
```diff
@@ -0,0 +1,11 @@
+package org.schabi.newpipe.extractor.exceptions;
+
+public class GeographicRestrictionException extends ParsingException {
```
Stypox, line 3: I'd have this extend `ContentNotAvailableException`, in order not to introduce breaking changes in the library

extractor/src/main/java/org/schabi/newpipe/extractor/exceptions/SoundCloudGoPlusException.java
```diff
@@ -0,0 +1,11 @@
+package org.schabi.newpipe.extractor.exceptions;
+
+public class SoundCloudGoPlusException extends ParsingException {
```
Stypox, line 3: Also here

Stypox (changes requested):
Except from that one thing it looks good ;-)

TiA4f8R:
@Stypox Changes requested applied.

Stypox (approved):
Looks good, thank you! This will be merged after 0.20.9 is released

TobiGr (approved):
Looks good.
Can be merge once the conflict is solved.

TiA4f8R:
@TobiGr Unfortunately, it doesn't work for some videos from type of https://www.youtube.com/watch?v=ayI2iBwGdxw (members-only content), so it shouldn't be merged until this is fixed.

opusforlife2:
@TiA4f8R You could always do that in a later PR and let this one be merged. 4 out of 5 or 80% is pretty good, no?

TiA4f8R:
@opusforlife2 You're right. Also, when I updated my PR, I added a check and if the reason message is null, NewPipe shows now the default `Content Unavailable` message. With this check, no `Could not get any stream` exception is thrown so this bug isn't a blocker and the PR can be merged without something strange.

TobiGr:
Some tests fail, because the exceptions changed. Please fix them

TiA4f8R:
@TobiGr Done. Thanks to letting me know this because it will break the extraction of unexisiting YouTube videos. So it was not tests but my code.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -720,9 +724,46 @@ public void onFetchPage(@Nonnull final Downloader downloader)
...
+            }
+            if (status.toLowerCase().equals("unplayable")) {
+                if (reason != null) {
+                    if (reason.equals("This video is only available to Music Premium members")) {
+                        throw new YoutubeMusicPremiumContentException();
+                    }
+                    if (reason.equals("This video requires payment to watch.")) {
+                        throw new PaidContentException("This video is a paid video");
+                    }
+                    if (reason.equals("Join this channel to get access to members-only content like this video and other exclusive perks.")) {
```
TobiGr, line 748:
```suggestion
                    if (reason.equals("Join this channel to get access to members-only content like this video, and other exclusive perks.")) {
```

TobiGr (changes requested):
Code looks good, but we need some proper testing.

Please add
``` Java
        @Test(expected = PrivateContentException.class)
        public void privateContent() throws Exception {
            final StreamExtractor extractor =
                    YouTube.getStreamExtractor(BASE_URL + "8VajtrESJzA");
            extractor.fetchPage();
        }

        @Test(expected = PaidContentException.class)
        public void paidContent() throws Exception {
            final StreamExtractor extractor =
                    YouTube.getStreamExtractor(BASE_URL + "ayI2iBwGdxw");
            extractor.fetchPage();
        }
```
to this class: https://github.com/TeamNewPipe/NewPipeExtractor/blob/9c1a7f7df6ca7fcee52dd61b30dc8a77e2bcbf99/extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/stream/YoutubeStreamExtractorDefaultTest.java#L55

I do not have time to create tests for the other exceptions in SoundCloud and the not yet tested PaidException in YouTube. But I guess, that should not be a problem for you ;)

TiA4f8R:
For `PaidException` I don't know if I can make a test because YouTube movies are geo-restricted and videos which requires a membership, like the one you linked, don't throw this exception, I don't understand what is the problem here.

TobiGr:
The problem is quite simple: you check if the message equals a string. we need to know when this string changes, because otherwise, we cannot detect the problem precisely. For example: YouTube changed the member-only message in the time between you writing the code and me testing it.

TiA4f8R:
@TobiGr Oh sorry, I didn't see your suggestion before I wrote my comment.

rocketinventor:
Hi, I was testing this out with the apk from Pull  Request #5385  by @TiA4f8R (which implements these error messages) and I found another type of error that is not on this list - Restricted Mode errors...

If YouTube's restricted mode is enabled in NewPipe and a "restricted" video is opened in the player then a generic "Content unavailable" message will be displayed. Someone who is unaware of this will think that there is an issue with the video extractor.

There should be a way to detect this... Does anybody know how to do this?

TobiGr:
@rocketinventor That might need an exception. However, we detected that before, but after some refactoring, the detection does not work any more. Good catch!
That should be easy to fix in this PR.

I suggest to do the following (@XiangRongLin, @B0pol, @Stypox any other ideas how to solve this?):

- in https://github.com/TiA4f8R/NewPipeExtractor/blob/f3c311e5ffb12c0be155a45852e50a6d146d8cf6/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java#L720-L723, check if we already have a status and save it before fetching the new video info.We should do this, because info from embedded videos is not that detailed and some videos (especially age-restricted ones) cannot be played via the embedded player. 
- if we did not get a good result with the new info, handle  the previously stored message, e.g. in this case set `ageLimit` to `18`.  We get almost every info, but  `streamingData`. So we cannot extract the duration, subtitles as well as audio and video streams. This will later throw an exception "Could not get any stream", but at least the info about the stream is displayed.
The front-ends know how to handle this. 

For later (maybe not in this PR): we should a add a `AgeRestrictedContentException`. We will need this in the future once YouTube disabled age restricted videos completely for users who are not logged in. As far as I can see, it is not possible to obtain the `streamingData` from the given JSON now. I don't know how much info is going to be available after YouTube applied some more changes to age restricted content.

---------
And why the hell are the tests failing again -_-

TiA4f8R:
> And why the hell are the tests failing again -_-

It's a SoundCloud intermittent test failure (the search test), so don't worry ;)

rocketinventor:
@TiA4f8R would your method work for opening a video through a direct link, or only from within the app?

---

Also, just to be clear, there are _two_  levels of age restricted content (in YouTube) - each corresponding with thier own setting in NewPipe:

**Potentially Mature Content** - Videos
blocked by "Restricted Mode":
> Turn on YouTube's "Restricted Mode"
YouTube provides a "Restricted Mode" which hides potentially mature content

**Restricted (Mature) Content**  - Videos _unblocked_ through the following opt-in:
> Show age restricted content
Show content possibly unsuitable for children because it has an age limit (like 18+)

If there is a way to detect one of these vs the other, then they should probably each have there own exceptions. If not, the error messages, at least, should be different (depending on which content restriction settings are in place from within NewPipe).

I.e. If a video is blocked while "Restricted Mode" is on then the message maybe should suggest to disable "Restricted Mode" to see the video.

rocketinventor:
I just tested this with the video URL for a "restricted" video from #4869 (same apk as before). The "This video is age restricted." error message definitely does work _when "restricted mode" is off_.

TobiGr (approved):

TiA4f8R:
Please do not merge this for now, I have to made some code improvements.

PR #512, opened 2021-01-15
Ignore all failing tests

XiangRongLin:
  
Ignore all currently failing tests.
This way one finally can see again if the own changes broke something (additionally).
I'll unignore the ones where there are only minor string changes, when i adjust them to inject the mocks. The ones that are actually broken, have to be fixed by someone else in the future.

### DIFF
--- a/extractor/src/test/java/org/schabi/newpipe/extractor/services/peertube/PeertubeStreamExtractorTest.java
+++ b/extractor/src/test/java/org/schabi/newpipe/extractor/services/peertube/PeertubeStreamExtractorTest.java
@@ -1,6 +1,7 @@
 package org.schabi.newpipe.extractor.services.peertube;
 
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.schabi.newpipe.downloader.DownloaderTestImpl;
 import org.schabi.newpipe.extractor.NewPipe;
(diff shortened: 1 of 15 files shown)


extractor/src/test/java/org/schabi/newpipe/extractor/services/peertube/PeertubeStreamExtractorTest.java
```diff
@@ -111,35 +113,60 @@ public static void setUp() throws Exception {
...
         @Override public long expectedLikeCountAtLeast() { return 1; }
+
         @Override public long expectedDislikeCountAtLeast() { return 0; }
+
         @Override public int expectedAgeLimit() { return 18; }
+
         @Override public boolean expectedHasAudioStreams() { return false; }
+
         @Override public boolean expectedHasSubtitles() { return false; }
+
```
Stypox, line 159: Please don't add all of these spaces

extractor/src/test/java/org/schabi/newpipe/extractor/services/DefaultStreamExtractorTest.java
```diff
@@ -231,6 +233,7 @@ public void testAgeLimit() throws Exception {
 
     @Test
     @Override
+    @Ignore
```
Stypox, line 236: Forgot to add `("TODO fix")` here?

Stypox (changes requested):
I agree failing tests should be ignored, until proper mocked tests exist

extractor/src/test/java/org/schabi/newpipe/extractor/services/peertube/PeertubeStreamExtractorTest.java
XiangRongLin: ok

extractor/src/test/java/org/schabi/newpipe/extractor/services/DefaultStreamExtractorTest.java
XiangRongLin: Yes, but looking at it again, i should probably not ignore the base class one but instead the one of the sub class that is failing...

XiangRongLin: It occured because in android studio it just jumps to the base class, since the sub class are not overriding the test.

Stypox (approved):

XiangRongLin:
@Stypox So much for ignoring failing tests. There are already new ones :smile:

PR #513, opened 2021-01-15
Adjust CI workflow to run tests with mocks

XiangRongLin:

This means that all tests that are already using mocks (currently only `YoutubeMixPlaylistExtractorTest`) will not test against the real Websites anymore. For that a new job needs to be added.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,4 +26,4 @@ jobs:
           restore-keys: ${{ runner.os }}-gradle
 
       - name: Build and run Tests
-        run: ./gradlew check --stacktrace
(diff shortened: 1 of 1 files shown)


Stypox (approved):
> For that a new job needs to be added.

Would you consider that important?

XiangRongLin:
> 
> 
> Would you consider that important?

Not that much. With big bugs, we will get flooded with issues about it (like new yt-decrypt format). Small ones aren't getting fixed (See the amount of failing tests and for how long they are already failing).

But now i know more about github actions and can easily just combine them into a single file i think.

Stypox:
> Not that much. 

I agree with you. I'm merging this, thank you :-D

PR #514, opened 2021-01-15
Pass on gradle JVM system property to test JVM

XiangRongLin:

Seems like i misunderstood how gradle system properties work and somehow even missed it.
Now the system property is actually accessable from inside the tests, meaning CI jobs with a different downloader passed in via CLI, will actually use those downloaders


XiangRongLin:
@Stypox done

Stypox (approved):

PR #516, opened 2021-01-15
Add if-else statement to run tests against real website on schedule

XiangRongLin:

I realized that running the tests against real website is basically just copying the ci.yml and changing one value.
Which can also be done with an if condition.

Because github actions `if` has no `else`, the shell script ones are used

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,9 @@
 name: CI
 
 on:
+  schedule:
+      # once per day
+    - cron: 0 0 * * *
   push:
     branches:
       - dev
@@ -25,5 +28,11 @@ jobs:
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
           restore-keys: ${{ runner.os }}-gradle
 
+        # See gradle file for difference between downloaders
       - name: Build and run Tests
-        run: ./gradlew check --stacktrace -Ddownloader=MOCK
+        run: |
+          if [[ $GITHUB_EVENT_NAME == 'schedule' ]]; then
+            ./gradlew check --stacktrace -Ddownloader=REAL
+          else
+            ./gradlew check --stacktrace -Ddownloader=MOCK
+          fi


Stypox:
I like this! Is it really that simple? :-D

XiangRongLin:
Should be.
I forgot to link this one: https://github.com/XiangRongLin/NewPipeExtractor/pull/4
where the action does: https://github.com/XiangRongLin/NewPipeExtractor/pull/4/checks

XiangRongLin:
@Stypox let the action on my repo run for a night and tomorrow we can see if it echoed the correct thing

XiangRongLin:
Seeing how there already are new failing tests not even an hour after merging #512, we probably need a way to mark tests to be skipped in the ci pipeline if they are not run with mocks.

Stypox:
@XiangRongLin good idea

XiangRongLin:
> let the action on my repo run for a night and tomorrow we can see if it echoed the correct thing

It works
https://github.com/XiangRongLin/NewPipeExtractor/runs/1711997487?check_suite_focus=true#step:2:8

Stypox:
Great! @TobiGr

.github/workflows/ci.yml
```diff
@@ -1,6 +1,8 @@
 name: CI
 
 on:
+  schedule:
+    - cron: 0 0 * * *
```
TobiGr, line 5: Once a day? Coud you add that as a comment, so it is easeir to understand?

.github/workflows/ci.yml
```diff
@@ -25,5 +27,11 @@ jobs:
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
           restore-keys: ${{ runner.os }}-gradle
 
+        #See gradle file for difference between downloaders
```
TobiGr, line 30:
```suggestion
        # See gradle file for difference between downloaders
```

TobiGr (commented):
Can be merged once review is implemented

.github/workflows/ci.yml
XiangRongLin: done

.github/workflows/ci.yml
XiangRongLin: done

TobiGr (approved):

TobiGr:
Ah wait. wasn't there a complaint regarding automated testing when they took down youtube-dl?

@TheAssassin This introduces sheduled tests against the real websites, whereas some predefined JSONs are tested against PRs. Is that ok?

XiangRongLin:
@TobiGr I think part of it was that there we unit tests, that were downloading copyrighted stuff

see last paragraph of this section https://github.blog/2020-11-16-standing-up-for-developers-youtube-dl-is-back/#youtube-dl

So we would need to check all tests once and see if there are any such tests

PR #517, opened 2021-01-16
Add MockOnlyRule to allow skipping specific tests based on downloader

XiangRongLin:

### -
- addresses point 7 of #481
- from https://github.com/junit-team/junit4/issues/116

### showcase
- changes: https://github.com/XiangRongLin/NewPipeExtractor/commits/flaky_test_showcase
- Run that branch locally with `gradle check -Ddownloader=REAL` and gradle check -Ddownloader=MOCK`
  - Mock will fail because failing test is not skipped
  - REAL will succeed because failing test is skipped

### DIFF
--- a/extractor/src/test/java/org/schabi/newpipe/MockOnly.java
+++ b/extractor/src/test/java/org/schabi/newpipe/MockOnly.java
@@ -6,6 +6,8 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+import javax.annotation.Nonnull;
+
 /**
  * Marker annotation to skip test if it not run with mocks.
  *
@@ -15,4 +17,9 @@ import java.lang.annotation.Target;
 @Target({ElementType.METHOD, ElementType.TYPE})
 @Inherited
 public @interface MockOnly {
+
+    /**
+     * Explanation why this test shold only be run with mocks and not against real websites
+     */
+    @Nonnull String reason();
 }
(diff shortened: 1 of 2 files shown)


XiangRongLin:
After thinking about it a bit more i think adding a custom test runner is better.
- it can be specified globaly in gradle file (i think, could not find it immediatly). If not there is an additional overhead of specifiying it
- Can be easily added on method level, instead of class level

XiangRongLin:
With rules it's also possible to skip tests on method level instead of class level. Other that that it's the same addtional overhead compared to the developer to add it. But it allows to use other test runner (Mockito).

I adjusted the showcase branch

extractor/src/test/java/org/schabi/newpipe/MockOnly.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * {@link MockOnlyRule}
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD, ElementType.TYPE})
+@Inherited
+public @interface MockOnly {
+
+    /**
+     * Explanation why this test shold only be run with mocks and not against real websites
```
B0pol, line 22:
```suggestion
     * Explanation why this test should only be run with mocks and not against real websites
```

XiangRongLin: done

Stypox (approved):

Stypox:
@XiangRongLin can this be merged? If yes, feel free to do it yourself 😉

PR #518, opened 2021-01-17
Use and generate mocks for youtube tests

XiangRongLin:

- I removed some and added some other Ignores.
  - Added Ignores were mostly file encoding problems (i think)
- Some tests are not immediatly compatible, see each commit message


XiangRongLin:
So any opinions about this?

I think #527 is a good example on why this is necessary. Normally a change like that can be easily merged if the tests pass. But the tests suite failed there. No one knows if it's due to the change or unrelated broken/flaky tests.

Stypox:
I like this approach, as I said in #482. The only thing that I think could be improved is that you always have to remember to put `YoutubeParsingHelper.resetClientVersionAndKey();` in every YouTube test. I would implement this function somewhere instead and call it in the setup method of each test:
```java
void initNewpipeForTest(@Nullable final String mockResourcePath) {
    YoutubeParsingHelper.resetClientVersionAndKey();
    if (mockResourcePath == null) {
        NewPipe.init(new DownloaderFactory().getDownloader()); // or whatever the signature is
    } else {
        NewPipe.init(new DownloaderFactory().getDownloader(mockResourcePath));
    }
}

**Comment by XiangRongLin** [comment]
I personally don't see much difference, since one already has to remember to use the `DownloaderFactory` and not `DownloaderTestImpl` like one usually does. If it is absolutly needed, i can add it.

Additionally that specific resetting is only needed for Youtube and not the other services. So if someone used that `initNewpipeForTest` for SoundClound, they would run code completly unrelated and unneeded. If the other services have global state too (haven't checked), which would require a "reset" method, then `initNewpipeForTest` would just get bloated.

`YoutubeParsingHelper.resetClientVersionAndKey()` is a quick and dirty solution, since i didn't want to work on the global state in `YoutubeParsingHelper` while working on this.

**Comment by XiangRongLin** [comment]
@Stypox Any update on your opinion about `YoutubeParsingHelper.resetClientVersionAndKey()`

I would like to get this moving

**Comment by Stypox** [comment]
Sorry, I missed your response above. I didn't think that running code for unrelated services could cause unrelated and misguiding errors, so you are right, it's better to keep things separate. Everything should be well-documented then.

**Comment by XiangRongLin** [comment]
Well this was more annoying than expected. After running with the recording a few times random tests started to fail. I guess youtube was thinking i was doing a DOS against them...

@Stypox Where do you want the information, that the `reset` method has to be used? And has it to be part of this PR or can i address it in a seperate one? I would prefer a seperate one.

**Comment by Stypox** [comment]
@XiangRongLin yes, you can do it in a separate one.

File: extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/stream/YoutubeStreamExtractorDefaultTest.java [hunk 1]
```diff
@@ -249,15 +264,49 @@ public static void setUp() throws Exception {
...
-        @Override public long expectedLength() { return 1010; }
-        @Override public long expectedViewCountAtLeast() { return 815500; }
-        @Nullable @Override public String expectedUploadDate() { return "2020-11-18 00:00:00.000"; }
-        @Nullable @Override public String expectedTextualUploadDate() { return "2020-11-18"; }
-        @Override public long expectedLikeCountAtLeast() { return 48500; }
-        @Override public long expectedDislikeCountAtLeast() { return 20000; }
-        @Override public boolean expectedHasSubtitles() { return true; }
-
-        @Override public int expectedStreamSegmentsCount() { return 7; }
+
```
Stypox, line 267: Please undo this spacing-out

Stypox (changes requested):
Looks good to me, I think this can be merged after 0.20.10 is done with. Thank you!

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/stream/YoutubeStreamExtractorDefaultTest.java
XiangRongLin: done

Stypox (approved):

PR #522, opened 2021-01-22
update invidious instances list

B0pol:

teamnewpipe/newpipe#5479 part 1/2


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeStreamLinkHandlerFactory.java
```diff
@@ -186,7 +186,6 @@ public String getId(String urlString) throws ParsingException, IllegalArgumentEx
             case "WWW.INVIDIO.US":
             case "DEV.INVIDIO.US":
             case "INVIDIO.US":
-            case "VID.ENCRYPTIONIN.SPACE":
```
TobiGr, line 189: I suggest to keep the both lists in the same order

TobiGr (approved):

PR #523, opened 2021-01-22
Extract whether the comment is pinned

B0pol:

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfoItem.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfoItem.java
@@ -17,6 +17,7 @@ public class CommentsInfoItem extends InfoItem {
     private DateWrapper uploadDate;
     private int likeCount;
     private boolean heartedByUploader;
+    private boolean pinned;
 
     public CommentsInfoItem(int serviceId, String url, String name) {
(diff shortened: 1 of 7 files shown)


TobiGr (approved):

extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfoItem.java
```diff
@@ -94,4 +95,12 @@ public void setHeartedByUploader(boolean isHeartedByUploader) {
     public boolean getHeartedByUploader() {
         return this.heartedByUploader;
     }
+
+    public boolean getPinned() {
```
TobiGr, line 99: Ah damn. I was too fast. That should be `isPinned` (and `isHeartedByUploader`)

opusforlife2: You could just click on `Revert`?

B0pol: No, I'll create a new PR for both

B0pol: https://github.com/TeamNewPipe/NewPipeExtractor/pull/525

Issue #273, opened 2020-03-02
Support SoundCloud HLS


wb9688:
Some newer streams on SoundCloud only seem to have a MP3 and Opus HLS URL and no MP3 progressive URL. Here is an example JSON (just a random track from New & Hot):

```json
{
  "comment_count": 337,
  "full_duration": 188656,
  "downloadable": false,
  "created_at": "2020-02-27T16:32:45Z",
  "description": null,
  "media": {
    "transcodings": [
      {
        "url": "https://api-v2.soundcloud.com/media/soundcloud:tracks:767454454/bbe7e1ab-3b8a-4b3e-863f-28f674b5c6aa/stream/hls",
        "preset": "mp3_0_1",
        "duration": 188656,
        "snipped": false,
        "format": {
          "protocol": "hls",
          "mime_type": "audio/mpeg"
        },
        "quality": "sq"
      },
      {
        "url": "https://api-v2.soundcloud.com/media/soundcloud:tracks:767454454/64568e31-dc65-456f-9bce-b70b7786aa8c/stream/hls",
        "preset": "opus_0_0",
        "duration": 188613,
        "snipped": false,
        "format": {
          "protocol": "hls",
          "mime_type": "audio/ogg; codecs=\"opus\""
        },
        "quality": "sq"
      }
    ]
  },
  "title": "No Sucker",
  "publisher_metadata": {
    "p_line_for_display": "℗ Quality Control Music/Motown Records; ℗ 2020 Quality Control Music, LLC, under exclusive license to UMG Recordings, Inc.",
    "artist": "Lil Baby, Moneybagg Yo",
    "isrc": "USUG12000051",
    "c_line": "2020 © 2020 Quality Control Music, LLC, under exclusive license to UMG Recordings, Inc.",
    "upc_or_ean": "00602508756467",
    "p_line": "Quality Control Music/Motown Records; ℗ 2020 Quality Control Music, LLC, under exclusive license to UMG Recordings, Inc.",
    "urn": "soundcloud:tracks:767454454",
    "explicit": true,
    "c_line_for_display": "© 2020 © 2020 Quality Control Music, LLC, under exclusive license to UMG Recordings, Inc.",
    "contains_music": true,
    "id": 767454454,
    "album_title": "My Turn",
    "release_title": "No Sucker"
  },
  "duration": 188656,
  "has_downloads_left": true,
[...]

PR #526, opened 2021-01-23
Support SoundCloud HLS-only tracks by using a workaround

TiA4f8R:
Support SoundCloud HLS-only streams by parsing M3U manifests, get the last segment URL (in order to get track length) and request a segment URL equals to track's duration so it's a single URL (`0/track_length/`).

Related issue: #273

Test APK: [app.zip](https://github.com/TiA4f8R/NewPipe/suites/2255244146/artifacts/46920150) (from my fork with GitHub CI, app based on dev branch)

**To do**: add tests, if there is no risk of takedown of this repo by RIAA or by other music rights management companies by the use of SoundCloud HLS-only tracks.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
@@ -195,6 +195,19 @@ public class SoundcloudStreamExtractor extends StreamExtractor {
         try {
             final JsonArray transcodings = track.getObject("media").getArray("transcodings");
 
+            // Iterate a first time to see if there is a progressive MP3 stream available.
+            // If yes, the MP3 HLS stream will be not added to audioStreams.
+
+            boolean mp3ProgressiveStreamInTranscodings = false;
+
+            for (final Object transcoding : transcodings) {
+                final JsonObject t = (JsonObject) transcoding;
+                if (t.getString("preset").contains("mp3") &&
+                    t.getObject("format").getString("protocol").equals("progressive")) {
+                    mp3ProgressiveStreamInTranscodings = true;
+                }
+            }
+
             // Get information about what stream formats are available
             for (final Object transcoding : transcodings) {
                 final JsonObject t = (JsonObject) transcoding;
@@ -205,6 +218,13 @@ public class SoundcloudStreamExtractor extends StreamExtractor {
 
                 if (!isNullOrEmpty(url)) {
                     if (t.getString("preset").contains("mp3")) {
+                        // Don't add the MP3 HLS stream if there is a progressive stream present
+                        // because the two have the same bitrate
+                        if (t.getObject("format").getString("protocol").equals("hls") &&
+                            mp3ProgressiveStreamInTranscodings) {
+                            continue;
+                        }
+
                         mediaFormat = MediaFormat.MP3;
                         bitrate = 128;
                     } else if (t.getString("preset").contains("opus")) {
@@ -222,13 +242,14 @@ public class SoundcloudStreamExtractor extends StreamExtractor {
                         final String res = dl.get(url).responseBody();
 
                         try {
-                            JsonObject mp3UrlObject = JsonParser.object().from(res);
+                            final JsonObject mp3UrlObject = JsonParser.object().from(res);
                             // Links in this file are also only valid for a short period.
                             mediaUrl = mp3UrlObject.getString("url");
                         } catch (final JsonParserException e) {
                             throw new ParsingException("Could not parse streamable url", e);
                         }
                     } else if (t.getObject("format").getString("protocol").equals("hls")) {
+
                         // This url points to the endpoint which generates a unique and short living url to the stream.
                         url += "?client_id=" + SoundcloudParsingHelper.clientId();
                         final String res = dl.get(url).responseBody();
@@ -236,9 +257,13 @@ public class SoundcloudStreamExtractor extends StreamExtractor {
                         try {
                             final JsonObject mp3HlsUrlObject = JsonParser.object().from(res);
                             // Links in this file are also only valid for a short period.
-
                             // Parsing the HLS manifest to get a single file by requesting a range equal to 0-track_length
-                            final String hlsManifestResponse = dl.get(mp3HlsUrlObject.getString("url")).responseBody();
+                            final String hlsManifestResponse;
+                            try {
+                                hlsManifestResponse = dl.get(mp3HlsUrlObject.getString("url")).responseBody();
+                            } catch (final IOException e) {
+                                continue;
+                            }
                             final List<String> hlsRangesList = new ArrayList<>();
                             final Matcher regex = Pattern.compile("((https?):((//)|(\\\\))+[\\w\\d:#@%/;$()~_?+-=\\\\.&]*)")
                                     .matcher(hlsManifestResponse);
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
@@ -10,7 +10,6 @@ import org.schabi.newpipe.extractor.NewPipe;
 import org.schabi.newpipe.extractor.StreamingService;
 import org.schabi.newpipe.extractor.downloader.Downloader;
 import org.schabi.newpipe.extractor.exceptions.ContentNotAvailableException;
-import org.schabi.newpipe.extractor.exceptions.ContentNotSupportedException;
 import org.schabi.newpipe.extractor.exceptions.ExtractionException;
 import org.schabi.newpipe.extractor.exceptions.GeographicRestrictionException;
 import org.schabi.newpipe.extractor.exceptions.ParsingException;
@@ -29,7 +28,10 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
+import static org.schabi.newpipe.extractor.utils.Utils.HTTPS;
 import static org.schabi.newpipe.extractor.utils.Utils.*;
 
 public class SoundcloudStreamExtractor extends StreamExtractor {
@@ -182,7 +184,7 @@ public class SoundcloudStreamExtractor extends StreamExtractor {
 
     @Override
     public List<AudioStream> getAudioStreams() throws IOException, ExtractionException {
-        List<AudioStream> audioStreams = new ArrayList<>();
+        final List<AudioStream> audioStreams = new ArrayList<>();
         final Downloader dl = NewPipe.getDownloader();
 
         // Streams can be streamable and downloadable - or explicitly not.
@@ -193,41 +195,75 @@ public class SoundcloudStreamExtractor extends StreamExtractor {
         try {
             final JsonArray transcodings = track.getObject("media").getArray("transcodings");
 
-            // get information about what stream formats are available
-            for (Object transcoding : transcodings) {
-
+            // Get information about what stream formats are available
+            for (final Object transcoding : transcodings) {
                 final JsonObject t = (JsonObject) transcoding;
                 String url = t.getString("url");
+                final String mediaUrl;
+                final MediaFormat mediaFormat;
+                final int bitrate;
 
                 if (!isNullOrEmpty(url)) {
+                    if (t.getString("preset").contains("mp3")) {
+                        mediaFormat = MediaFormat.MP3;
+                        bitrate = 128;
+                    } else if (t.getString("preset").contains("opus")) {
+                        mediaFormat = MediaFormat.OPUS;
+                        bitrate = 64;
+                    } else {
+                        continue;
+                    }
 
-                    // We can only play the mp3 format, but not handle m3u playlists / streams.
-                    // what about Opus?
-                    if (t.getString("preset").contains("mp3")
-                            && t.getObject("format").getString("protocol").equals("progressive")) {
+                    // TODO: move this to a separate method to generate valid urls when needed (e.g. resuming a paused stream)
+
+                    if (t.getObject("format").getString("protocol").equals("progressive")) {
                         // This url points to the endpoint which generates a unique and short living url to the stream.
-                        // TODO: move this to a separate method to generate valid urls when needed (e.g. resuming a paused stream)
                         url += "?client_id=" + SoundcloudParsingHelper.clientId();
                         final String res = dl.get(url).responseBody();
 
                         try {
                             JsonObject mp3UrlObject = JsonParser.object().from(res);
                             // Links in this file are also only valid for a short period.
-                            audioStreams.add(new AudioStream(mp3UrlObject.getString("url"),
-                                    MediaFormat.MP3, 128));
-                        } catch (JsonParserException e) {
+                            mediaUrl = mp3UrlObject.getString("url");
+                        } catch (final JsonParserException e) {
                             throw new ParsingException("Could not parse streamable url", e);
                         }
+                    } else if (t.getObject("format").getString("protocol").equals("hls")) {
+                        // This url points to the endpoint which generates a unique and short living url to the stream.
+                        url += "?client_id=" + SoundcloudParsingHelper.clientId();
+                        final String res = dl.get(url).responseBody();
+
+                        try {
+                            final JsonObject mp3HlsUrlObject = JsonParser.object().from(res);
+                            // Links in this file are also only valid for a short period.
+
+                            // Parsing the HLS manifest to get a single file by requesting a range equal to 0-track_length
+                            final String hlsManifestResponse = dl.get(mp3HlsUrlObject.getString("url")).responseBody();
+                            final List<String> hlsRangesList = new ArrayList<>();
+                            final Matcher regex = Pattern.compile("((https?):((//)|(\\\\))+[\\w\\d:#@%/;$()~_?+-=\\\\.&]*)")
+                                    .matcher(hlsManifestResponse);
+
+                            while (regex.find()) {
+                                hlsRangesList.add(hlsManifestResponse.substring(regex.start(0), regex.end(0)));
+                            }
+
+                            final String hlsLastRangeUrl = hlsRangesList.get(hlsRangesList.size() - 1);
+                            final String[] hlsLastRangeUrlArray = hlsLastRangeUrl.split("/");
+
+                            mediaUrl = HTTPS + hlsLastRangeUrlArray[2] + "/media/0/" + hlsLastRangeUrlArray[5] + "/" + hlsLastRangeUrlArray[6];
+                        } catch (final JsonParserException e) {
+                            throw new ParsingException("Could not parse streamable url", e);
+                        }
+                    } else {
+                        continue;
                     }
+
+                    audioStreams.add(new AudioStream(mediaUrl, mediaFormat, bitrate));
                 }
             }
 
-        } catch (NullPointerException e) {
+        } catch (final NullPointerException e) {
             throw new ExtractionException("Could not get SoundCloud's track audio url", e);
-        }
-
-        if (audioStreams.isEmpty()) {
-            throw new ContentNotSupportedException("HLS audio streams are not yet supported");
         }
 
         return audioStreams;
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
@@ -29,10 +29,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
-import static org.schabi.newpipe.extractor.utils.Utils.HTTPS;
 import static org.schabi.newpipe.extractor.utils.Utils.*;
 
 public class SoundcloudStreamExtractor extends StreamExtractor {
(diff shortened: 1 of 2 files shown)


Stypox (commented):
The issue is apparently that the streams that you open always end up in a `continue;` case, and therefore at the end of `getAudioStreams()` `audioStreams.isEmpty()` is `true`.

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
```diff
@@ -190,54 +193,102 @@ public String getHlsUrl() {
...
 
-        } catch (NullPointerException e) {
+        } catch (final NullPointerException e) {
             throw new ExtractionException("Could not get SoundCloud's track audio url", e);
         }
 
-        if (audioStreams.isEmpty()) {
-            throw new ContentNotSupportedException("HLS audio streams are not yet supported");
-        }
-
```
Stypox, line 229: For debugging purposes I suggest you to revert this, so that you are instantly notified if `audioStreams.isEmpty()` is `true`

TiA4f8R:
@Stypox It seems that it is something in my HLS manifest parsing code which doesn't work or throws something because even if I readded the code you suggest me to readd, I always get `Could not get any stream. [...]` exception on any SoundCloud track. I really need help because I don't know what is the problem of my code, moreover I am not a developer but someone who is trying some things.

Stypox:
@TiA4f8R ok, I'm gonna try to debug this myself. Since I'm pretty busy in the next few days, if I forget about this feel free to ping me ;-)

TiA4f8R:
@Stypox I am a bit too, as you can see on my profile. Thanks for taking care of this!

TiA4f8R:
@Stypox I've seen what's wrong: it's my regex.

TiA4f8R:
@Stypox Code ready for review and HLS-only tracks are now streamable.

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
```diff
@@ -190,50 +192,109 @@ public String getHlsUrl() {
...
+                        mediaFormat = MediaFormat.MP3;
+                        bitrate = 128;
+                    } else if (t.getString("preset").contains("opus")) {
+                        mediaFormat = MediaFormat.OPUS;
+                        bitrate = 64;
+                    } else {
+                        continue;
+                    }
+
+                    // TODO: move this to a separate method to generate valid urls when needed (e.g. resuming a paused stream)
```
B0pol, line 234: todo

B0pol (changes requested):

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
TobiGr: that's a todo i left while hotfixing soundcloud. It should be done, but in a separate PR.

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
```diff
@@ -185,43 +187,102 @@ public String getHlsUrl() {
...
 
+            boolean mp3ProgressiveStreamInTranscodings = false;
+
+            for (final Object transcoding : transcodings) {
+                final JsonObject t = (JsonObject) transcoding;
+                if (t.getString("preset").contains("mp3") &&
+                    t.getObject("format").getString("protocol").equals("progressive")) {
+                    mp3ProgressiveStreamInTranscodings = true;
+                }
+            }
```
XiangRongLin, line 201: Can't you just `break` out of the loop after it's `true` the first time?

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
```diff
@@ -185,43 +187,102 @@ public String getHlsUrl() {
...
 
-            // get information about what stream formats are available
-            for (Object transcoding : transcodings) {
+            // Iterate a first time to see if there is a progressive MP3 stream available.
+            // If yes, the MP3 HLS stream will be not added to audioStreams.
 
+            boolean mp3ProgressiveStreamInTranscodings = false;
+
+            for (final Object transcoding : transcodings) {
+                final JsonObject t = (JsonObject) transcoding;
```
XiangRongLin, line 196: No single letter variables

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
```diff
@@ -185,43 +187,102 @@ public String getHlsUrl() {
...
+                final String mediaUrl;
+                final MediaFormat mediaFormat;
+                final int bitrate;
 
                 if (!isNullOrEmpty(url)) {
+                    if (t.getString("preset").contains("mp3")) {
+                        // Don't add the MP3 HLS stream if there is a progressive stream present
+                        // because the two have the same bitrate
+                        if (t.getObject("format").getString("protocol").equals("hls") &&
+                            mp3ProgressiveStreamInTranscodings) {
```
XiangRongLin, line 216: Reversing the order of the two boolean should give a very very minor performance boost, since `mp3ProgressiveStreamInTranscodings` is already calculated

XiangRongLin (changes requested):
Overall the code is very hard to read and there are no unit tests for the added functionality.

There is a `for` loop which spans across nearly 100 lines (I think, cant really see where it ends). Add to that if-else if-else, more loops inside and sudden jumps with `continue´ where you have to know in which loop you are.

----

Try to extract out some methods. You already have comments describing that the section is for a specific type of endpoint. Make a method out of that.

I just looked at it until the my last comment. I did not look at the rest.

XiangRongLin:
Something along the lines of having `getAudioStream` be an integration method and the extracted method containing the actual operation. The integration method can then be "read" like normal text and not programming code, to get to know what is done.

see https://clean-code-developer.com/grades/grade-1-red/#Integration_Operation_Segregation_Principle_IOSP

B0pol:
Also extract the regex pattern to a constant and give it a proper name.

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
```diff
@@ -185,43 +187,102 @@ public String getHlsUrl() {
...
-            for (Object transcoding : transcodings) {
+            // Iterate a first time to see if there is a progressive MP3 stream available.
+            // If yes, the MP3 HLS stream will be not added to audioStreams.
 
+            boolean mp3ProgressiveStreamInTranscodings = false;
+
+            for (final Object transcoding : transcodings) {
+                final JsonObject t = (JsonObject) transcoding;
+                if (t.getString("preset").contains("mp3") &&
+                    t.getObject("format").getString("protocol").equals("progressive")) {
```
XiangRongLin, line 198: Another minor that i personally think helps readability is seperating the parts where json is parsed and the part where the comparison is made. Inside the if I don't need to also know how the data was extracted in order to understand if it should go into the if statement

```java
String preset = t.getString("preset");
String protocol = t.getObject("format").getString("protocol");
if (preset.contains("mp3") && protocol.equals("progressive")) {
```

Some performance is lost because `protocol` is always evaluated unlike before and more objects are created. but it's easier to read. And maybe there is some way to remove that downside, that i can't think of right now.
Whether that tradeoff is worth it I can't say, especially since we are on Android where resources are a lot more previous compared to other plattforms

opusforlife2: A lot more previous?

XiangRongLin: precious!!!

stupid github. Whenever i start reviewing and edit a review comment before submitting it, it does not save the edit

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
TiA4f8R: It was initially this variable name so I kept it. I will change it.

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
```diff
@@ -286,10 +274,42 @@ public String getHlsUrl() {
         return audioStreams;
     }
 
-    private static String urlEncode(String value) {
+    private final static String PATTERN_WEB_URLS_IN_HLS_MANIFESTS = "((http?|https?):((//)|(\\\\))+[\\w\\d:#@%/;$()~_?+-=\\\\.&]*)";
```
B0pol, line 277: Please make the compiled pattern static, not the string. Also (http?|https?) => (https?)? i think

B0pol (changes requested):

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
TiA4f8R: I think all SoundCloud streaming formats are https but if they are http, the extractor can extract them.

XiangRongLin:
@TiA4f8R I'm still not seeing any unit tests, please add those.
Before i review the code for quality, i would like to know beforehand that it actually works by seeing unit tests.

TiA4f8R:
@XiangRongLin Thanks for your comment. What check in a test should I do on an HLS only-track? Assert that the audioStreams for a track are not empty?

XiangRongLin:
The tests need to cover the things that ensure that this feature is working.

If i went ahead an deleted your changes, the tests must fail.  
If i went ahead an refactored you code, but made mistakes, the tests must fail.  
If the format returned by soundcloud changes, the tests must fail.

For starters you could think about what you tested manually (I hope you did this part). Meaning which url did you open in the app and what did you expect to see. Just adjusted for the extrator.

XiangRongLin:
> What check in a test should I do on an HLS only-track? Assert that the audioStreams for a track are not empty?

Easiest would maybe be copying one of the existing tests, replace the requested URL with a HLS one. Then adjust expected values to the ones that should come out.

TiA4f8R:
> **To do**: add tests, if there is no risk of takedown of this repo by RIAA or by other music rights management companies.

Because HLS-only tracks are only copyrighted contents (aren't they?), I can't use in tests because RIAA can takedown the extractor repo for the same reason as youtube-dl, am I right or not? What is your advice about this, dear members of the NewPipe team?

XiangRongLin:
The only thing i can think of is to use the `MockDownloader`, generate the json files for a copyrighted track and manually edit the url in the json, so that the extracted url is the same one as inside the json, but returns a 404 when opened. The test then needs to be marked as `MockOnly` #517.

Thats very unreliable and work intensive, so i'd say just to skip the tests in this case

TiA4f8R:
@XiangRongLin I added a test, using the current Creative Commons SoundCloud test track, what do you think of it? Note that CI is failing because of a wrong subscriber count in a PeerTube test, not because of my code.

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
TiA4f8R: I changed `t` to `transcodingJsonObject`.

XiangRongLin (commented):
The method is still very big und hard to read. But if the test does cover it, then you can refactor that method in subsequent PRs.

As for the test, someone else will have to check if that is a HLS stream since i don't use soundcloud and don't know it

TiA4f8R:
@XiangRongLin What do you think now?

XiangRongLin:
@TiA4f8R Grab yourself the SonarLint plugin. It gives you an objective review of bad practices, code smell and much more. In this case cyclomatic complexity is the important one.

From a quick scan it seems only a minor problem is present in the part that you touched. Decide for yourself if it is worth fixing, since it's probably you that has to touch that code again and try to comprehend it. I would just skip it for now. Come back when you have ideas on how to tackle it, because i don't

https://plugins.jetbrains.com/plugin/7973-sonarlint

XiangRongLin (approved):

TiA4f8R:
@XiangRongLin Thank you.

The only thing that I want to implement but I don't know how to is the support of SoundCloud downloadable files (original files uploaded by the artist, when the artist allows downloads of his/her/its track).

I know how to get the URL and add an audioStream but there are two missing informations: the bitrate and the format. Here is my initial work:

I was thinking of returning `-1` when bitrate is unknown and `MediaFormat.ORIGINAL_FORMAT` when format is unknown.

```
        // If the track can be downloadable on SoundCloud's website, add the original file to the audioStreams only if there is standard streams available in order to prevent a high data consumption
        if (!audioStreams.isEmpty()) {
            if (track.getBoolean("downloadable")) {
                try {
                    final String dl_res = dl.get("https://api-v2.soundcloud.com/tracks/" + track.getInt("id") +
                            "/download?client_id=" + SoundcloudParsingHelper.clientId()).responseBody();
                    final JsonObject downloadUrlObject = JsonParser.object().from(dl_res);
                    audioStreams.add(new AudioStream(downloadUrlObject.getString("redirectUri"), MediaFormat.ORIGINAL_FORMAT, -1));
                } catch (final NullPointerException | IOException | JsonParserException ignored) {
                }
            }
        }
```

 What do you think about what I should do?

TiA4f8R:
I installed the plugin and here are the issues:
Issues

However, I don't understand how to fix it (sorry for my current programming knowledge). Can you help me please? Thank you in advance!

XiangRongLin:
> 
> However, I don't understand how to fix it (sorry for my current programming knowledge). Can you help me please? Thank you in advance!
> 

Most of the issues are not from you, so don't worry about them. 

The issues should all have an explanation with examples breaking and examples adhering to the rules. As for how to fix them: I don't know.

As I already mentioned it in my previous post. Come back to them when you know more.

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
```diff
@@ -181,62 +186,144 @@ public String getHlsUrl() {
...
 
-        if (audioStreams.isEmpty()) {
-            throw new ContentNotSupportedException("HLS audio streams are not yet supported");
+        final List<String> hlsRangesList = new ArrayList<>();
+        final Matcher pattern_matches = PATTERN_HTTPS_URLS_IN_HLS_MANIFESTS
+                .matcher(hlsManifestResponse);
+
+        while (pattern_matches.find()) {
+            hlsRangesList.add(hlsManifestResponse.substring(pattern_matches.start(0),
+                    pattern_matches.end(0)));
```
TobiGr, line 313: For context, [this is a SoundCloud HLS Manifest](https://gist.github.com/TobiGr/eabc45d311a887b3814882c7c25ecbe9).
From [RFC8216](https://tools.ietf.org/html/rfc8216#page-10), the HTTP Live Streaming standard:

> Lines in a Playlist file are terminated by either a single line feed
   character or a carriage return character followed by a line feed
   character.  Each line is a URI, is blank, or starts with the
   character '#'.  Blank lines are ignored.  Whitespace MUST NOT be
   present, except for elements in which it is explicitly specified.

This means, we do not need that big regex, but only something really small and fest to check if the line does not start with `#` and is not empty to know that we have a URL in that line. That should give a big performance improvement, especially for longer tracks. 

e.g. something like this (not tested, just typed into this review)
```Java
final String[] lines = hlsManifestResponse.split("\\r?\\n");
for (final String line : lines) {
    if (line.trim().length() != 0 && !line.startsWith("#") {
        hlsRangesList.add(line);
    }
}
```

@XiangRongLin @TiA4f8R Do you think that this is a good approach?

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
```diff
@@ -181,62 +186,144 @@ public String getHlsUrl() {
...
+
+                try {
+                    mediaUrl = getTranscodingUrl(url, protocol);
+                } catch (final Exception e) {
+                    // something went wrong when parsing this transcoding
+                    continue;
+                }
+                audioStreams.add(new AudioStream(mediaUrl, mediaFormat, bitrate));
             }
+        }
```
TobiGr, line 284: I think this is easier to read:
```suggestion
        for (final Object transcoding : transcodings) {
            final JsonObject transcodingJsonObject = (JsonObject) transcoding;
            final String url = transcodingJsonObject.getString("url");

            if (isNullOrEmpty(url)) {
                continue;
            }

            final String mediaUrl;
            final String preset = transcodingJsonObject.getString("preset");
            final String protocol = transcodingJsonObject.getObject("format").getString("protocol");
            final MediaFormat mediaFormat;
            final int bitrate;

            if (preset.contains("mp3")) {
                // Don't add the MP3 HLS stream if there is a progressive stream present
                // because the two have the same bitrate
                if (mp3ProgressiveInStreams && protocol.equals("hls")) {
                    continue;
                }
                mediaFormat = MediaFormat.MP3;
                bitrate = 128;
            } else if (preset.contains("opus")) {
                mediaFormat = MediaFormat.OPUS;
                bitrate = 64;
            } else {
                // Unknown format
                continue;
            }

            try {
                mediaUrl = getTranscodingUrl(url, protocol);
                audioStreams.add(new AudioStream(mediaUrl, mediaFormat, bitrate));
            } catch (final Exception e) {
                // something went wrong when parsing this transcoding
            }

        }
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
```diff
@@ -181,62 +186,144 @@ public String getHlsUrl() {
...
+        if (protocol.equals("progressive")) {
+            return urlString;
+        } else if (protocol.equals("hls")) {
+            return getSingleUrlFromHlsManifest(urlString);
+        }
+        // else, unknown protocol
+        return "";
+    }
+
+    private static void setUpAudioStreams(final JsonArray transcodings,
```
TobiGr, line 247: rename this to `extractAudioStreams`

TobiGr (commented):
Structure looks good. Thank you for having a look!

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
TobiGr: you could make this even shorter by removing the last `continue` and `else` branch and adding a check whether `mediaFormat` is not empty.

TiA4f8R:
Thank you for your suggestions @TobiGr. I will apply it soon.

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
TiA4f8R: @TobiGr The last line of the manifest is `#EXT-X-ENDLIST`, we can get the second last line, which is the last segment of the track and contains the range of the song, can't we?

Here is what it can looks:

```Java
        final String[] lines = hlsManifestResponse.split("\\r?\\n");
        final String hlsLastRangeUrl = hlsRangesList.get(lines.size() - 2);
        final String[] hlsLastRangeUrlArray = hlsLastRangeUrl.split("/");

        return HTTPS + hlsLastRangeUrlArray[2] + "/media/0/" + hlsLastRangeUrlArray[5] + "/"
                + hlsLastRangeUrlArray[6];
```

What do you think?

TobiGr: The `#EXT-X-ENDLIST` is not required. We do not know for sure that the last URL is in the second last line. That's why I'd go through the the array reverse. You can start at the last index and go decrease it.
``` Java
for (int l = lines.length - 1; l >= 0; l--;) {
    if (line.trim().length() != 0 && !line.startsWith("#") {
        // you could also check , whether the line contains a valid URL
        // do what you wand to do here,
    }
}
```

TiA4f8R: And for the first URL found, can I break the loop because it should be the last segment URL?

TobiGr: Yes, you cloud do that. I'd do that to increase readbillity and also make error detection easier.
``` Java
for (int l = lines.length - 1; l >= 0; l--;) {
    final String line = lines[l];
    // get the last URL from manifest, because [add more info here]
    if (line.trim().length() != 0 && !lines.startsWith("#") {
        // you could also check , whether the line contains a valid URL
        final String[] hlsLastRangeUrlArray = line.split("/");
        return HTTPS + hlsLastRangeUrlArray[2] + "/media/0/" + hlsLastRangeUrlArray[5] + "/"
                + hlsLastRangeUrlArray[6];
    }
}
throw new ParsingException("Could not get any URL from HLS manifest");
```

TiA4f8R: Your code perfectly works, I ran tests and URLs of HLS transcodings works and are equals to the track length!

TobiGr (approved):

PR #528, opened 2021-01-27
[YouTube Music] Fix extracting search item view/subscriber count when = 0

Stypox:
- [no] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).
- [not needed] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API.

Fixes the comment https://github.com/TeamNewPipe/NewPipe/issues/5455#issuecomment-767818647

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicSearchExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicSearchExtractor.java
@@ -20,9 +20,8 @@ import org.schabi.newpipe.extractor.localization.TimeAgoParser;
 import org.schabi.newpipe.extractor.search.InfoItemsSearchCollector;
 import org.schabi.newpipe.extractor.search.SearchExtractor;
 import org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper;
-import org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubePlaylistLinkHandlerFactory;
-import org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeStreamLinkHandlerFactory;
 import org.schabi.newpipe.extractor.utils.JsonUtils;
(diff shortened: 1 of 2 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicSearchExtractor.java
```diff
@@ -105,7 +108,7 @@ public void onFetchPage(@Nonnull final Downloader downloader) throws IOException
                 .end()
                 .value("query", getSearchString())
                 .value("params", params)
-            .end().done().getBytes("UTF-8");
+            .end().done().getBytes(StandardCharsets.UTF_8);
```
TobiGr, line 111: Is that going to break compatibility with the legacy version?
Or can we use that safely, because this is a precoompiled lib?

Stypox: Oh, I have no idea, I can revert that commit if needed

TobiGr: See https://github.com/TeamNewPipe/NewPipeExtractor/pull/269
We should remove the other occurencies of that class, too.

Isira-Seneviratne: There's an issue to provide a backport: https://issuetracker.google.com/issues/169047771

Stypox: Reverted @TobiGr

TobiGr: thanks

TobiGr (approved):

PR #539, opened 2021-02-12
[YouTube]: Fixed channel continuation

B0pol:

Closes https://github.com/TeamNewPipe/NewPipe/issues/5550

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
@@ -20,9 +20,7 @@ import org.schabi.newpipe.extractor.utils.Utils;
 import javax.annotation.Nonnull;
 import java.io.IOException;
 
-import static org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper.fixThumbnailUrl;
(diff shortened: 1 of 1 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
```diff
@@ -252,36 +250,47 @@ public String getParentChannelAvatarUrl() throws ParsingException {
         final JsonArray ajaxJson = getJsonResponse(page.getUrl(), getExtractorLocalization());
 
         JsonObject sectionListContinuation = ajaxJson.getObject(1).getObject("response")
-                .getObject("continuationContents").getObject("gridContinuation");
+                .getArray("onResponseReceivedActions").getObject(0).getObject("appendContinuationItemsAction");
```
TobiGr, line 253: is it safe to assume that the `appendContinuationItemsAction` object is always in the first array element?

B0pol: I don't know

B0pol: youtube-dl also assume it's the first element. https://github.com/ytdl-org/youtube-dl/blob/master/youtube_dl/extractor/youtube.py#L2507
but if you wanna be sure the safest way is a simple for loop.

TobiGr (approved):

PR #540, opened 2021-02-12
fix tests, fix links in YouTube description

B0pol:

no need

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
@@ -429,10 +429,16 @@ public class YoutubeParsingHelper {
         return youtubeMusicKeys = new String[]{key, clientName, clientVersion};
     }
 
+
+
     @Nullable
     public static String getUrlFromNavigationEndpoint(JsonObject navigationEndpoint) throws ParsingException {
         if (navigationEndpoint.has("urlEndpoint")) {
             String internUrl = navigationEndpoint.getObject("urlEndpoint").getString("url");
+            if (internUrl.startsWith("https://www.youtube.com/redirect?")) {
+                internUrl = internUrl.substring(23);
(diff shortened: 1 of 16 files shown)


B0pol:
Actual fixes: links in description for YouTube, getName of youtube Trending extractor, link handler get id from null URL

The other ones were tests working but ignored

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
```diff
@@ -429,10 +429,16 @@ public static boolean areHardcodedYoutubeMusicKeysValid() throws IOException, Re
...
+
     @Nullable
     public static String getUrlFromNavigationEndpoint(JsonObject navigationEndpoint) throws ParsingException {
         if (navigationEndpoint.has("urlEndpoint")) {
             String internUrl = navigationEndpoint.getObject("urlEndpoint").getString("url");
+            if (internUrl.startsWith("https://www.youtube.com/redirect?")) {
+                internUrl = internUrl.substring(23);
+            }
+
             if (internUrl.startsWith("/redirect?")) {
```
TobiGr, line 442:
```suggestion
            } else if (internUrl.startsWith("/redirect?")) {
```

B0pol: No, it's expected. When it starts with `https://youtube.com/redirect?`, I only remove the `https://youtube.com` part so that it falls in this condition.

TobiGr: oh ok. please add that info as comment. I would not expect that behaviour and a comment is better to understand than counting letters.

B0pol: done

TobiGr (approved):

PR #552, opened 2021-02-19
Add dependabot

FireMasterK:

This will automatically create PRs to keep dependencies up to date.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  # Maintain dependencies for Gradle
+  - package-ecosystem: "gradle"
+    directory: "/"
+    schedule:
+      interval: "daily"
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"


Stypox:
Wouldn't this flood newpipe with PRs to merge?

FireMasterK:
In a way, yes, but you can always ignore a dependency if you don't want to upgrade it.

XiangRongLin:
Did you check if the test coverage and quality is high enough to merge those PRs without additional manual work?

FireMasterK:
I'm not sure what exactly you mean, I would say the best way to evaluate these PRs is to check the changelog, check if it compiles, check if there's any breaking change, and then decide accordingly.

I'm not sure what tests would help in evaluating such PRs 🤔.

XiangRongLin:
So that if there are any changes in the behavior of the library, we know it.

- Like if they accidently break their library and release it, we don't also break our library too.
- Or if we relied on some unintended behaviour, which got fixed in the library, but then breaks our functionality.

FireMasterK:
That's a difficult question, I usually read the changelog to determine that. The only way to surely know is to test it otherwise.

Stypox:
I'd merge this. Upgrading one thing at a time is a good idea amd we'd have to check changelogs anyway

B0pol (approved):

PR #572, opened 2021-03-05
Fixes missing comments under videos

golfinq:

Fixes TeamNewPipe/NewPipe#5585, TeamNewPipe/NewPipe#5757, TeamNewPipe/NewPipe#5759.

It seems youtube is in the process of slightly changing their API, and the comment extractor could no longer find the ctoken. Based on reports that the old way sometimes worked, I left it in as an option. I have tested it on my phone and the app works correctly with this change.


golfinq:
The failed tests seem to concern parts of the code which this PR does not alter; I am not sure why it is failing.

TobiGr (approved):

TobiGr:
Thank you!
The other test are failing, because YouTube made further changes to the API, e.g. to the continuations. SoundCloud is failing as well, but that will be fixed by nother PR, too.

B0pol:
You should swap the test, as you expect the second one to be working and the first probably not (for few days it's not working)

PR #574, opened 2021-03-05
[YouTube] Mobile API for extracting video info.

FireMasterK:

Needs testing. Once I test, I will mark it as ready.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
@@ -1,9 +1,6 @@
 package org.schabi.newpipe.extractor.services.youtube.extractors;
 
-import com.grack.nanojson.JsonArray;
-import com.grack.nanojson.JsonObject;
-import com.grack.nanojson.JsonParser;
-import com.grack.nanojson.JsonParserException;
+import com.grack.nanojson.*;
 import org.jsoup.Jsoup;
 import org.jsoup.nodes.Document;
 import org.jsoup.nodes.Element;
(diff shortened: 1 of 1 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -702,20 +699,48 @@ public String getErrorMessage() {
...
-            initialData = initialAjaxJson.getObject(2).getObject("response", null);
+        final String url = "https://youtubei.googleapis.com/youtubei/v1/player?key=" + getKey();
+
+        // @formatter:off
+        byte[] json = JsonWriter.string()
+                .object()
+                .object("context")
+                    .object("client")
+                        .value("clientName", "Android")
+                        .value("clientVersion", "16.02")
```
TiA4f8R, line 711: Can you try updating the client version to the latest release version (16.08 from APKMirror) and see if URLs are not encrypted?

FireMasterK: 16.02 is the version that YouTube Vanced uses, but yes, 16.06 does work as well.

FireMasterK:
From further testing, it looks like an additional request is needed to get data such as likes.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -702,20 +699,48 @@ public String getErrorMessage() {
...
             throws IOException, ExtractionException {
-        initialAjaxJson = getJsonResponse(getUrl() + "&pbj=1", getExtractorLocalization());
 
-        initialData = initialAjaxJson.getObject(3).getObject("response", null);
-        if (initialData == null) {
-            initialData = initialAjaxJson.getObject(2).getObject("response", null);
+        final String url = "https://youtubei.googleapis.com/youtubei/v1/player?key=" + getKey();
+
+        // @formatter:off
+        byte[] json = JsonWriter.string()
```
Stypox, line 706: I'd just write this as a string, it would be more readable in the end.

Stypox (commented):
Great! This is really good news! I opened your files in ARC, everything seems fine. It's not a problem if two requests are needed, though we should probably restructure the extractor so that metadata fetching can be done separately from video stream fetching.

Stypox:
@FireMasterK that's not a big issue

FireMasterK:
Closing in favor of #604

PR #575, opened 2021-03-05
Fix Peertube account subscribers extraction

TiA4f8R:
Fix Peertube account subscribers extraction by adding the subscribers of all channels of an account to the subscribers of this account.

Minimal subscriber count for Framasoft account has been increased to 750 in the test.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubeAccountExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubeAccountExtractor.java
@@ -79,7 +79,6 @@ public class PeertubeAccountExtractor extends ChannelExtractor {
         } catch (final IOException | JsonParserException | ReCaptchaException ignored) {
             // something went wrong during video channels extraction, only return subscribers of ownerAccount
         }
-        System.out.println(subscribersCount);
         return subscribersCount;
     }
 
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubeAccountExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubeAccountExtractor.java
@@ -1,5 +1,6 @@
 package org.schabi.newpipe.extractor.services.peertube.extractors;
 
+import com.grack.nanojson.JsonArray;
 import com.grack.nanojson.JsonObject;
 import com.grack.nanojson.JsonParser;
 import com.grack.nanojson.JsonParserException;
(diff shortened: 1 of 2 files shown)


TobiGr:
I already took a look at that one, too. But there seems to be a huge difference between the API result for this channel and the follower count [on the website](https://framatube.org/accounts/framasoft): 425 vs 952 
To be honest, I have no idea, why we get different data here.

TiA4f8R:
When I loaded the first time the website, it shows 491 followers first briefly and after 952. I don't understand why.

B0pol:
I noticed something : it is what we've previously called a "parent channel". The channel, owner channel has 491 subscribers, but the subchannels, video channels have 461 subscribers if you sum them all. That makes 952 subscribers

TiA4f8R:
@B0pol I implement what you said, this should be done now.
Tests are failing but it's not related to the subscribers extraction.

extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubeAccountExtractor.java
```diff
@@ -55,8 +58,29 @@ public String getFeedUrl() throws ParsingException {
...
+            final JsonObject jsonResponse = JsonParser.object().from(responseBody);
+            final JsonArray videoChannels = jsonResponse.getArray("data");
+            for (final Object videoChannel : videoChannels) {
+                final JsonObject videoChannelJsonObject = (JsonObject) videoChannel;
+                subscribersCount += videoChannelJsonObject.getInt("followersCount");
+            }
+        } catch (final IOException | JsonParserException | ReCaptchaException ignored) {
+            // something went wrong during video channels extraction, only return subscribers of ownerAccount
+        }
+        System.out.println(subscribersCount);
```
Stypox, line 82: Remove this

TiA4f8R: I was thinking I remove all prints but no, sorry. Will fix it later.

Stypox (approved):

Stypox:
@B0pol @TobiGr could you merge this if you think it's ok?

extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubeAccountExtractor.java
```diff
@@ -55,8 +58,28 @@ public String getFeedUrl() throws ParsingException {
     }
 
     @Override
-    public long getSubscriberCount() {
-        return json.getLong("followersCount");
+    public long getSubscriberCount() throws ParsingException {
+        long subscribersCount = json.getLong("followersCount");
```
TobiGr, line 62:
```suggestion
        // The subscriber count cannot be retrieved directly. It needs to be calculated.
        // An accounts subscriber count is the number of the channel owner's subscriptions
        // plus the sum of all sub channels subscriptions.
        long subscribersCount = json.getLong("followersCount");
```

TobiGr (approved):
Looks good to me. We should add a comment about what is done here and why it is done, so other people going through the code will know.

TiA4f8R:
A Peertube test is failing because of a new thumbnail URL. Do not merge this PR, I will fix it soon.

PR #580, opened 2021-03-22
Add AccountTerminatedException for terminated channels

TobiGr:

Adds a new `ContentNotAvailableException`: `AccountTerminatedException` to distinguish between terminated channels. The field `reason` provides more info on why the account has been terminated. This fixes a NPE when getting the feed for a terminated channel in YouTubeFeedExtractor.
- [ ] ~~add `CLOSED_BY_OWNER` reason (I need a dedicated channel for this~~
- [ ] ~~implement in other services~~
- [x] add mocks for tests


TobiGr:
I could not find more info on deleted and terminated channels in the docs of the other services.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeChannelExtractorTest.java
```diff
@@ -48,6 +61,91 @@ public void nonExistentFetch() throws Exception {
...
+
+        @Test(expected = AccountTerminatedException.class)
+        public void accountTerminatedCopyrightFetch() throws Exception {
+            final ChannelExtractor extractor =
+                    YouTube.getChannelExtractor("https://www.youtube.com/channel/UCpExuV8qJMfCaSQNL1YG6bQ");
+            try {
+                extractor.fetchPage();
+            } catch (AccountTerminatedException e) {
+                assertEquals(e.getMessage(),
+                        "This account has been terminated because we received multiple third-party claims of copyright infringement regarding material that the user posted.");
```
XiangRongLin, line 143: `assertEquals` has expected first, actual second 
Same for the other asserts.

see https://junit.org/junit4/javadoc/latest/org/junit/Assert.html#assertEquals(java.lang.Object[],%20java.lang.Object[])

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeChannelExtractorTest.java
```diff
@@ -48,6 +61,91 @@ public void nonExistentFetch() throws Exception {
...
+
+        @Test(expected = AccountTerminatedException.class)
+        public void accountTerminatedTOSFetch() throws Exception {
+            final ChannelExtractor extractor =
+                    YouTube.getChannelExtractor("https://www.youtube.com/channel/UCTGjY2I-ZUGnwVoWAGRd7XQ");
+            try {
+                extractor.fetchPage();
+            } catch (AccountTerminatedException e) {
+                assertEquals(e.getMessage(),
+                        "This account has been terminated for a violation of YouTube's Terms of Service.");
```
XiangRongLin, line 73: I would remove the assertions for the error message. They are bound to fail very soon, if they change the wording. Only usecase i can think of for them, is directly displaying them to the end-user in which case, the exact wording does not change the correctness of the implementation.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
```diff
@@ -692,6 +689,23 @@ public static void defaultAlertsCheck(final JsonObject initialData) throws Parsi
             final String alertText = getTextFromObject(alertRenderer.getObject("text"));
             final String alertType = alertRenderer.getString("type", EMPTY_STRING);
             if (alertType.equalsIgnoreCase("ERROR")) {
+                if (alertText != null && alertText.contains("This account has been terminated")) {
+                    if (alertText.contains("violation") || alertText.contains("violating")
+                            || alertText.contains("infringement")) {
```
XiangRongLin, line 694: what happens if the localization is changes to non-english. I would assume the error messages are localized

TiA4f8R: @XiangRongLin Yes, they are, it's the same thing for the error messages from YouTube videos.
I think in the future we need to implement something which stores strings of the error messages for each localization and compare with the value provided by YouTube in the language used for the extraction of YouTube datas.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeChannelExtractorTest.java
TobiGr: oh yes

Stypox (approved):
LGTM

Stypox:
Oh wait, this has to be addressed: `I would remove the assertions for the error message. They are bound to fail very soon, if they change the wording. Only usecase i can think of for them, is directly displaying them to the end-user in which case, the exact wording does not change the correctness of the implementation.`

Stypox (approved):

Issue #593, opened 2021-03-31
Display suggestions


fynngodau:
When viewing an album or track on the web, Bandcamp offers suggestion under the heading "If you like …, you may also like". These should be parsed and displayed.

Stypox:
This should be changed (I don't remember to what), but also requires a change on NewPipe's side.

PR #597, opened 2021-03-31
Rename getRelatedStreams to getRelatedItems and change return type

fynngodau:

Preparations for #593.

* Change return type of the previous `getRelatedStreams()` calls from `StreamInfoItemsCollector` to `InfoItemsCollector`
* Rename all occurrences of "related streams" to "related items"

This works without any major changes to the NewPipe app for these reasons:
* `ExtractorHelper.getRelatedVideosOrLogError` is the only caller of `StreamExtrctor.getRelatedItems()` and already returns a more generic `List<InfoItem>`
* NewPipe's `StreamInfo` object has the value `List<InfoItem> relatedStreams`, which is also generic.
* `PlayerHelper.autoQueueOf` already tests whether the items it handels are `StreamInfoItems`. (We might later want to change this to also support playlists in some form.)
* `RelatedStreamInfo extends ListInfo<InfoItem>`
* `RelatedVideosFragment extends BaseListInfoFragment<RelatedStreamInfo>`

I tried it with Streams, Channels and Playlist, and all were displayed correctly.

This NewPipe PR changes the NewPipe code to keep naming entirely consistent (and to remove inconsistencies like "related videos"): TeamNewPipe/NewPipe#5974

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampStreamExtractor.java
@@ -245,7 +245,7 @@ public class BandcampStreamExtractor extends StreamExtractor {
     }
 
     @Override
-    public StreamInfoItemsCollector getRelatedStreams() {
+    public StreamInfoItemsCollector getRelatedItems() {
         return null;
     }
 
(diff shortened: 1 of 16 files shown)


XiangRongLin (changes requested):
@fynngodau Can you make this api change per deprecations instead of outright removing it. With the current way it's going, there is going to be a breaking API change in a patch version. You can add the "forRemoval" field (or whatever it's called) set to v0.22.0 or v0.23.0 

Anyone who uses the library (me too) will love you for that.

fynngodau:
@XiangRongLin Thanks for the great suggestion, I added deprecated calls that wrap the new ones.

XiangRongLin (approved):

PR #598, opened 2021-04-02
Fix a test and ignore rest

XiangRongLin:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [ ] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).
- [ ] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API.

PR #599, opened 2021-04-02
Add body field to Page and use it

XiangRongLin:

Is based on #598 so that i can see if it runs in CI


TiA4f8R (commented):
Can you also switch the API domain from `https://www.youtube.com/youtubei/v1` to `https://youtubei.googleapis.com/youtubei/v1` for the POST requests? It can reduce the consent cookies redirects and responses are the same.

extractor/src/main/java/org/schabi/newpipe/extractor/Page.java
```diff
@@ -59,4 +66,8 @@ public static boolean isValid(final Page page) {
         return page != null && (!isNullOrEmpty(page.getUrl())
                 || !isNullOrEmpty(page.getIds()));
     }
+
+    public byte[] getBody() {
```
TobiGr, line 70: Please annotate that method as nullable

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubePlaylistExtractor.java
```diff
@@ -259,9 +245,16 @@ private Page getNextPageFrom(final JsonArray contents) throws IOException, Extra
                     .getObject("continuationEndpoint")
                     .getObject("continuationCommand")
                     .getString("token");
+
```
TiA4f8R, line 248: Remove this empty line.

TiA4f8R (approved):
Looks good, just a nit below.

XiangRongLin:
Test failure is from soundcloud and not from youtube

PR #602, opened 2021-04-06
Ignore unreliable SoundcloudGoPlusTrack.testRelatedItems

XiangRongLin:


TiA4f8R:
I don't understand SoundCloud logic to return sometimes related tracks to Go+ tracks and sometimes not. Does they are testing a new layout for Go+ tracks in their website, which doesn't include related tracks?

TiA4f8R (approved):

PR #617, opened 2021-04-25
Fix Google URL detection

TobiGr:

Some google search URLs were not recognized correctly.


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
```diff
@@ -96,7 +96,7 @@ private static boolean isGoogleURL(String url) {
         try {
             final URL u = new URL(url);
             final String host = u.getHost();
-            return host.startsWith("google.") || host.startsWith("m.google.");
+            return host.startsWith("google.") || host.startsWith("m.google.") || host.startsWith("www.google.");
```
TiA4f8R, line 99:
```suggestion
            return host.startsWith("google.") || host.startsWith("m.google.")
                    || host.startsWith("www.google.");
```

So you are in the 100 characters limit.

TiA4f8R (approved):

Issue #586, opened 2021-03-28
Bandcamp comments


fynngodau:
Users can leave comments on Bandcamp, this is called `fanthanks` in Bandcamp's code. Comments can be left for both albums as well as tracks by users who bought them.

Comments on individual tracks could be parsed and displayed in the app.

fynngodau:
Implemented through #620. Next issue: #621.

PR #620, opened 2021-04-26
Add Bandcamp comments (first page)

fynngodau:

Implements #586.

This PR only parses the first set of comments (max. 3), because for those we can parse the website while we need to use some sort of API for additional pages (which the JS on the website usually calls). I might open an additional PR in the future that adds the capability of loading additional pages.

(Here would be an album where you can click "more" to load more comments: https://alfamist.bandcamp.com/album/bring-backs)

Demo (https://c418.bandcamp.com/track/wet-hands):

Website | NewPipe
---|---
image | Screenshot_1619452412

### DIFF
--- a/extractor/src/test/java/org/schabi/newpipe/extractor/services/bandcamp/BandcampCommentsExtractorTest.java
+++ b/extractor/src/test/java/org/schabi/newpipe/extractor/services/bandcamp/BandcampCommentsExtractorTest.java
@@ -40,7 +40,6 @@ public class BandcampCommentsExtractorTest {
 
         DefaultTests.defaultTestListOfItems(Bandcamp, comments.getItems(), comments.getErrors());
         for (CommentsInfoItem c : comments.getItems()) {
-            assertFalse(Utils.isBlank(c.getUploaderUrl()));
             assertFalse(Utils.isBlank(c.getUploaderName()));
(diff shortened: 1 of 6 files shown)


TobiGr (approved):

extractor/src/test/java/org/schabi/newpipe/extractor/services/bandcamp/BandcampCommentsExtractorTest.java
```diff
@@ -0,0 +1,53 @@
...
+        assertTrue(extractor.getInitialPage().getItems().size() >= 3);
+    }
+
+    @Test
+    public void testGetCommentsAllData() throws IOException, ExtractionException {
+        ListExtractor.InfoItemsPage<CommentsInfoItem> comments = extractor.getInitialPage();
+
+        DefaultTests.defaultTestListOfItems(Bandcamp, comments.getItems(), comments.getErrors());
+        for (CommentsInfoItem c : comments.getItems()) {
+            assertFalse(Utils.isBlank(c.getUploaderUrl()));
```
TobiGr, line 43: That fails, because you return an empty String  in the extractor

fynngodau: Thanks, sorry for the oversight.

PR #625, opened 2021-05-12
Add streamPosition for comments

TobiGr:
SoundCloud is the only service which supports adding comments at a specific timestamp in the stream.

- [ ] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API. __Not sure if this is necessary__

Add `CommentsInfoItem.NO_LIKE_COUNT` and `CommentsInfoItem.NO_STREAM_POSITION`.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfoItem.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfoItem.java
@@ -20,6 +20,9 @@ public class CommentsInfoItem extends InfoItem {
     private String textualLikeCount;
     private boolean heartedByUploader;
     private boolean pinned;
+    private int streamPosition;
+
+    public static final int NO_STREAM_POSITION = -1;
 
     public CommentsInfoItem(int serviceId, String url, String name) {
         super(InfoType.COMMENT, serviceId, url, name);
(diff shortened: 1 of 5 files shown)


opusforlife2:
How will this work in the UI?

TobiGr:
Not sure if it is needed for NewPipe, but we could display a "at 4:20" next to the comment's author name

opusforlife2:
Oh. So you've added this generally, and not specifically for Newpipe?

That is fine. I can't think of a neat way to include this in the app. The UI is too different from how Soundcloud presents it.

TobiGr:
> So you've added this generally, and not specifically for Newpipe?

Correct

extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampCommentsInfoItemExtractor.java
```diff
@@ -86,7 +86,12 @@ public boolean isPinned() {
...
 
     @Override
-    public boolean isUploaderVerified() throws ParsingException {
+    public boolean isUploaderVerified() {
         return false;
     }
+
+    @Override
+    public int getStreamPosition() {
+        return -1;
```
Stypox, line 95: Create a constant for this in `CommentsInfoItem`, something like `NO_STREAM_POSITION`

Stypox (changes requested):
Looking good

extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampCommentsInfoItemExtractor.java
TobiGr: Good idea :+1: 
I also added `NO_LIKE_COUNT` for comment likes

Stypox (approved):

PR #626, opened 2021-05-14
[SoundCloud] Fix getting ID from mobile URLs

TobiGr:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).
- [x] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API.

Fixes TeamNewPipe/NewPipe#6284

PR #627, opened 2021-05-15
[SoundCloud] Use a lightweight request to check if the hardcoded client_id is valid, fix the extraction of mobile URLs and more

TiA4f8R:
- [ ] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe). - See TiA4f8R/NewPipe#4 for an up to date debug APK.
- [ ] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API. - Not needed

- Format the SoundCloud package to be in the limit of 100 caracters per line, use final where possible, fix some code comments and warnings.
- Use a new and lightweight way to check if the hardcoded `client_id` is valid: do a request to `https://api-v2.soundcloud.com/?client_id=HARDCODED_CLIENT_ID`. If the response code is 404, the hardcoded `client_id` is valid; if the response code is 401 (and maybe other error codes), it isn't.
- Rename `removeWWWFromUrl` method of the `Utils` class of the extractor to `removeMAndWWWFromUrl` because it now also removes the `m.` part of the SoundCloud domains, in order to be recognized by the SoundCloud API (the widget API right now when trying to get ids of contents). This fixes the extraction of mobile URLs, like #626 but maybe in a better way.

Fixes TeamNewPipe/NewPipe#6284.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
@@ -119,8 +119,9 @@ public class SoundcloudParsingHelper {
      */
     public static JsonObject resolveFor(@Nonnull final Downloader downloader, final String url)
             throws IOException, ExtractionException {
-        final String apiUrl = SOUNDCLOUD_API_V2_URL + "resolve" + "?url="
-                + URLEncoder.encode(url, UTF_8) + "&client_id=" + clientId();
+        final String apiUrl = SOUNDCLOUD_API_V2_URL + "resolve"
+                + "?url=" + URLEncoder.encode(url, UTF_8) 
+                + "&client_id=" + clientId();
 
         try {
             final String response = downloader.get(apiUrl, SoundCloud.getLocalization())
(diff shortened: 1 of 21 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
```diff
@@ -85,77 +89,82 @@ public static synchronized String clientId() throws ExtractionException, IOExcep
...
      * See https://developers.soundcloud.com/docs/api/reference#resolve
      */
-    public static JsonObject resolveFor(Downloader downloader, String url) throws IOException, ExtractionException {
-        String apiUrl = "https://api-v2.soundcloud.com/resolve"
-                + "?url=" + URLEncoder.encode(url, UTF_8)
-                + "&client_id=" + clientId();
+    public static JsonObject resolveFor(@Nonnull final Downloader downloader, final String url)
+            throws IOException, ExtractionException {
+        final String apiUrl = SOUNDCLOUD_API_V2_URL + "resolve" + "?url="
+                + URLEncoder.encode(url, UTF_8) + "&client_id=" + clientId();
```
TobiGr, line 123:
```suggestion
        final String apiUrl = SOUNDCLOUD_API_V2_URL + "resolve"
                + "?url=" + URLEncoder.encode(url, UTF_8) 
                + "&client_id=" + clientId();
```

TobiGr (approved):
LGTM

PR #635, opened 2021-05-23
Update setup java

litetex:

### What is it?
- [x] Codebase improvement (dev facing)

### Description of the changes in your PR
* Update actions/setup-java v1 to v2 ([migration](https://github.com/actions/setup-java/blob/main/docs/switching-to-v2.md))
* Using adoptopenjdk instead of zulu
  * AdoptOpenJDK is unbranded (Zulu is branded)
  * AdoptOpenJDK is very close related to OpenJDK / less modified than zulu
  * There seem to be no specific commitments regarding the availability of Zulu's free constructions
  * Android studio is shipped with OpenJDK:
  demo
  * See also: https://github.com/actions/setup-java/issues/13, https://dzone.com/articles/java-and-the-jdks-which-one-to-use

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,10 +16,12 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: set up JDK 1.8
-        uses: actions/setup-java@v1.4.3
+      - name: set up JDK 8
+        uses: actions/setup-java@v2
         with:
-          java-version: 1.8
+          java-version: '8'
+          distribution: 'adopt'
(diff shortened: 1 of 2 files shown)


FireMasterK:
Could a matrix be used to test in OpenJ9 as well?

litetex:
> Could a matrix be used to test in OpenJ9 as well?

I wouldn't recommend that because the (default) JVM implementation 'HotSpot' is shipped by Oracle and therefore also included in the OpenJDK, which is used in Android Studio:
pic

PR #638, opened 2021-05-28
Improve exception message thrown when not accepting an URL in LinkHandlerFactory#fromUrl(String url, String baseUrl)

TobiGr:

It is not a malformed URL, the URL is just not accepted.


XiangRongLin (approved):
For the future you can try to shorten the header of the commit messsage. I think it was ~72 characters that will be displayed on GitHub, after which the rest is cut off

PR #651, opened 2021-06-08
[YouTube] Fix getting stream type of live streams

TobiGr:

Live streams are sometimes not recognized as such. That is causing the duration extraction to fail.

Fixes https://github.com/TeamNewPipe/NewPipeExtractor/runs/2769405487?check_suite_focus=true#step:5:169

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
@@ -62,16 +62,21 @@ public class YoutubeStreamInfoItemExtractor implements StreamInfoItemExtractor {
         }
 
         final JsonArray badges = videoInfo.getArray("badges");
-        for (Object badge : badges) {
-            if (((JsonObject) badge).getObject("metadataBadgeRenderer").getString("label", EMPTY_STRING).equals("LIVE NOW")) {
+        for (final Object badge : badges) {
+            if (((JsonObject) badge).getObject("metadataBadgeRenderer")
+                    .getString("label", EMPTY_STRING).equals("LIVE NOW")) {
                 return cachedStreamType = StreamType.LIVE_STREAM;
             }
         }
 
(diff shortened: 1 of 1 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
```diff
@@ -62,16 +62,21 @@ public StreamType getStreamType() {
...
                 return cachedStreamType = StreamType.LIVE_STREAM;
             }
         }
 
-        final String style = videoInfo.getArray("thumbnailOverlays").getObject(0)
-                .getObject("thumbnailOverlayTimeStatusRenderer").getString("style", EMPTY_STRING);
-        if (style.equalsIgnoreCase("LIVE")) {
-            return cachedStreamType = StreamType.LIVE_STREAM;
+        final JsonArray thumbnailOverlays = videoInfo.getArray("thumbnailOverlays");
+        for (final Object object: thumbnailOverlays) {
```
Stypox, line 73:
```suggestion
        for (final Object object : thumbnailOverlays) {
```

Stypox (changes requested):
Code looks good; I trust you that this works ;-)

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
TobiGr: Done

Stypox (approved):

litetex:
@TobiGr 
Code seems to be faulty: 
Now every video is shown as "LIVE":
grafik

Please check:
https://github.com/TeamNewPipe/NewPipeExtractor/blob/0cb48e36bd8ff798090fc2f174f2430959b46a88/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java#L76
Looks like the "or" ``||`` should be an "and" ``&&``

TobiGr:
Uhm. No. That should not be the case. Let me take a look. Thank you for the heads up!

PR #652, opened 2021-06-13
Fix yt comments and add disabled comments functionallity

### LINKED ISSUE #577: ParsingException when comments disabled
## Exception
* __User Action:__ requested comments
* __Request:__ https://m.youtube.com/watch?v=MhReKn8STX8
* __Content Country:__ FR
* __Content Language:__ fr-FR
* __App Language:__ fr_FR
* __Service:__ YouTube
* __Version:__ 0.20.11
* __OS:__ Linux Android 9 - 28

```
org.schabi.newpipe.extractor.exceptions.ParsingException: Could not parse json data for comments
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeCommentsExtractor.getPage(YoutubeCommentsExtractor.java:100)
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeCommentsExtractor.getInitialPage(YoutubeCommentsExtractor.java:55)
	at org.schabi.newpipe.extractor.utils.ExtractorHelper.getItemsPageOrLogError(ExtractorHelper.java:19)
	at org.schabi.newpipe.extractor.comments.CommentsInfo.getInfo(CommentsInfo.java:40)
	at org.schabi.newpipe.extractor.comments.CommentsInfo.getInfo(CommentsInfo.java:25)
	at org.schabi.newpipe.util.ExtractorHelper.lambda$getCommentsInfo$7(ExtractorHelper.java:167)
	at org.schabi.newpipe.util.-$$Lambda$ExtractorHelper$60N_-UL7E5eaxFaFO1bZZmnfwM8.call(Unknown Source:4)
	at io.reactivex.rxjava3.internal.operators.single.SingleFromCallable.subscribeActual(SingleFromCallable.java:43)
	at io.reactivex.rxjava3.core.Single.subscribe(Single.java:4813)
	at io.reactivex.rxjava3.internal.operators.single.SingleDoOnSuccess.subscribeActual(SingleDoOnSuccess.java:35)
	at io.reactivex.rxjava3.core.Single.subscribe(Single.java:4813)
	at io.reactivex.rxjava3.internal.operators.maybe.MaybeFromSingle.subscribeActual(MaybeFromSingle.java:41)
	at io.reactivex.rxjava3.core.Maybe.subscribe(Maybe.java:5330)
	at io.reactivex.rxjava3.internal.operators.maybe.MaybeConcatArray$ConcatMaybeObserver.drain(MaybeConcatArray.java:153)
	at io.reactivex.rxjava3.internal.operators.maybe.MaybeConcatArray$ConcatMaybeObserver.request(MaybeConcatArray.java:78)
	at io.reactivex.rxjava3.internal.operators.flowable.FlowableElementAtMaybe$ElementAtSubscriber.onSubscribe(FlowableElementAtMaybe.java:66)
	at io.reactivex.rxjava3.internal.operators.maybe.MaybeConcatArray.subscribeActual(MaybeConcatArray.java:42)
	at io.reactivex.rxjava3.core.Flowable.subscribe(Flowable.java:15753)
	at io.reactivex.rxjava3.internal.operators.flowable.FlowableElementAtMaybe.subscribeActual(FlowableElementAtMaybe.java:36)
	at io.reactivex.rxjava3.core.Maybe.subscribe(Maybe.java:5330)
	at io.reactivex.rxjava3.internal.operators.maybe.MaybeToSingle.subscribeActual(MaybeToSingle.java:46)
	at io.reactivex.rxjava3.core.Single.subscribe(Single.java:4813)
	at io.reactivex.rxjava3.internal.operators.single.SingleSubscribeOn$SubscribeOnObserver.run(SingleSubscribeOn.java:89)
	at io.reactivex.rxjava3.core.Scheduler$DisposeTask.run(Scheduler.java:614)
	at io.reactivex.rxjava3.internal.schedulers.ScheduledRunnable.run(ScheduledRunnable.java:65)
	at io.reactivex.rxjava3.internal.schedulers.ScheduledRunnable.call(ScheduledRunnable.java:56)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
Caused by: com.grack.nanojson.JsonParserException: Unexpected character: < on line 1, char 1
	at com.grack.nanojson.JsonTokener.createParseException(Unknown Source:44)
	at com.grack.nanojson.JsonTokener.advanceToToken(Unknown Source:118)
	at com.grack.nanojson.JsonParser.advanceToken(Unknown Source:12)
	at com.grack.nanojson.JsonParser.parse(Unknown Source:1)
	at com.grack.nanojson.JsonParser$JsonParserContext.from(Unknown Source:19)
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeCommentsExtractor.getPage(YoutubeCommentsExtractor.java:98)
	... 30 more

```
</details>
<hr>

Note: comments are disabled on this video

litetex:

Fixes #577 (ASAP field was set in march lol)
Also fixes https://github.com/TeamNewPipe/NewPipe/issues/5769

Changes:
* Fixes #577 
  * If no continuation token is sent, the comments are disabled
  * The ``findValue`` in ``YoutubeCommentsExtractor`` was faulty: If no "start" was found (e.g. ``continuation":"`` for the continuation token) it searched nearly everything, found ``apple-itunes-app`` (from ``<!DOCTYPE html><head><meta name="apple-itunes-app" content="app-id=544007664, app-argume...`` and used that as continuation-token, which resulted in the response of an internal server error (500) by YT's servers and caused NewPipe to show the error
* Adds a new field to the commentsExtractor ``disabledComments`` which is true, when the comments are disabled.
* Code clean up, according to the NewPipe Checkstyle config (added a lot of missing ``final`` keywords)
* Added comments

Notice: 
* NewPipe PR for the new ``disabledComments`` is here: https://github.com/TeamNewPipe/NewPipe/pull/6483
* Demo video: https://m.youtube.com/watch?v=MhReKn8STX8

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
@@ -23,9 +23,11 @@ import javax.annotation.Nonnull;
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.regex.Pattern;
 
 import static java.util.Collections.singletonList;
@@ -41,53 +43,99 @@ public class YoutubeCommentsExtractor extends CommentsExtractor {
     private String ytClientName;
     private String responseBody;
 
-    public YoutubeCommentsExtractor(StreamingService service, ListLinkHandler uiHandler) {
+    private Optional<Boolean> optCommentsDisabled = Optional.empty();
+
+    public YoutubeCommentsExtractor(
+            final StreamingService service,
+            final ListLinkHandler uiHandler) {
         super(service, uiHandler);
     }
 
     @Override
-    public InfoItemsPage<CommentsInfoItem> getInitialPage() throws IOException, ExtractionException {
-        String commentsTokenInside = findValue(responseBody, "sectionListRenderer", "}");
-        if (!commentsTokenInside.contains("continuation\":\"")) {
-            commentsTokenInside = findValue(responseBody, "commentSectionRenderer", "}");
+    public InfoItemsPage<CommentsInfoItem> getInitialPage()
+            throws IOException, ExtractionException {
+
+        // Check if the the findInitialCommentsToken was already called and initialized
+        if (optCommentsDisabled.orElse(false)) {
+            return getInfoItemsPageForDisabledComments();
         }
-        final String commentsToken = findValue(commentsTokenInside, "continuation\":\"", "\"");
+
+        // Get the token
+        final String commentsToken = findInitialCommentsToken();
+        // Check if the comments have been disabled
+        if (optCommentsDisabled.get()) {
+            return getInfoItemsPageForDisabledComments();
+        }
+
         return getPage(getNextPage(commentsToken));
     }
 
-    private Page getNextPage(JsonObject ajaxJson) throws ParsingException {
+    /**
+     * Finds the initial comments token and initializes commentsDisabled.
+     * @return
+     */
+    private String findInitialCommentsToken() {
+        final String continuationStartPattern = "continuation\":\"";
+
+        String commentsTokenInside = findValue(responseBody, "sectionListRenderer", "}");
+        if (commentsTokenInside == null || !commentsTokenInside.contains(continuationStartPattern)) {
+            commentsTokenInside = findValue(responseBody, "commentSectionRenderer", "}");
+        }
+
+        // If no continuation token is found the comments are disabled
+        if (commentsTokenInside == null || !commentsTokenInside.contains(continuationStartPattern)) {
+            optCommentsDisabled = Optional.of(true);
+            return null;
+        }
+
+        // If a continuation token is found there are >= 0 comments
+        final String commentsToken = findValue(commentsTokenInside, continuationStartPattern, "\"");
+
+        optCommentsDisabled = Optional.of(false);
+
+        return commentsToken;
+    }
+
+    private InfoItemsPage<CommentsInfoItem> getInfoItemsPageForDisabledComments() {
+        return new InfoItemsPage<>(Collections.emptyList(), null, Collections.emptyList());
+    }
+
+    private Page getNextPage(final JsonObject ajaxJson) throws ParsingException {
         final JsonArray arr;
         try {
             arr = JsonUtils.getArray(ajaxJson, "response.continuationContents.commentSectionContinuation.continuations");
-        } catch (Exception e) {
+        } catch (final Exception e) {
             return null;
         }
         if (arr.isEmpty()) {
             return null;
         }
-        String continuation;
+        final String continuation;
         try {
             continuation = JsonUtils.getString(arr.getObject(0), "nextContinuationData.continuation");
-        } catch (Exception e) {
+        } catch (final Exception e) {
             return null;
         }
         return getNextPage(continuation);
     }
 
-    private Page getNextPage(String continuation) throws ParsingException {
-        Map<String, String> params = new HashMap<>();
+    private Page getNextPage(final String continuation) throws ParsingException {
+        final Map<String, String> params = new HashMap<>();
         params.put("action_get_comments", "1");
         params.put("pbj", "1");
         params.put("ctoken", continuation);
         try {
             return new Page("https://m.youtube.com/watch_comment?" + getDataString(params));
-        } catch (UnsupportedEncodingException e) {
+        } catch (final UnsupportedEncodingException e) {
             throw new ParsingException("Could not get next page url", e);
         }
     }
 
     @Override
     public InfoItemsPage<CommentsInfoItem> getPage(final Page page) throws IOException, ExtractionException {
+        if (optCommentsDisabled.orElse(false)) {
+            return getInfoItemsPageForDisabledComments();
+        }
         if (page == null || isNullOrEmpty(page.getUrl())) {
             throw new IllegalArgumentException("Page doesn't contain an URL");
         }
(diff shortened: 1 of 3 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
```diff
@@ -41,53 +43,99 @@
     private String ytClientName;
     private String responseBody;
 
-    public YoutubeCommentsExtractor(StreamingService service, ListLinkHandler uiHandler) {
+    private Optional<Boolean> optCommentsDisabled = Optional.empty();
```
Stypox, line 46: Why not just set this to `false` by default instead of using an `Optional` and then `orElse(false)`?

Stypox (changes requested):
At a first glance it looks fine

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
litetex: @Stypox 

TL;DR

> Caching mechanism and holder of the commentsDisabled value.
> 
> Initial value = empty -> unknown if comments are disabled or not
> Some method calls {@link YoutubeCommentsExtractor#findInitialCommentsToken()} → value is set.
> If the method or another one that is depending on disabled comments is now called again, the method execution can avoid unnecessary calls.

Example:
* A new ``YoutubeCommentsExtractor`` is initialized
* ``isCommentsDisabled`` is called -> initialization in ``findInitialCommentsToken`` occurs 
* As it turns out the comments are disabled and therefore ``optCommentsDisabled`` is set to ``true``
* Shortly after that``getInitialPage`` is called -> we can instantly return the ``getInfoItemsPageForDisabledComments`` without processing ``findInitialCommentsToken`` again

Another example (which matches your question):
* A new ``YoutubeCommentsExtractor`` is initialized
* ``getInitialPage`` is called -> initialization in ``findInitialCommentsToken`` occurs 
* As it turns out the comments are NOT disabled and therefore ``optCommentsDisabled`` is set to ``false``
* Shortly after that ``isCommentsDisabled`` is called ->  we can instantly return the value of ``optCommentsDisabled`` because we already know it 
* if ``optCommentsDisabled`` would be initialized by default with ``false``, it's unknown if ``findInitialCommentsToken`` was already called and resulted in ``false`` or not and therefore the ``findInitialCommentsToken`` should then be executed again

B0pol (commented):
- PeerTube: commentsEnabled field in video
- SoundCloud: commentable field in video
- BandCamp: ?
- media.ccc.de: comments not implemented

Given SoundCloud and PeerTube give that info in video info and not comments, how should it be handled (I'm also talking to other reviewers)?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
Stypox: Ok, I see, this part looks good then

Stypox:
My proposal is to add `getCommentsExtractor()` to `StreamInfo`.

For services where comments and stream info are completely independent, like YouTube, it would just return the same thing as what is currently used, i.e. `YouTubeService.getCommentsExtractor(url)` (see [`CommentsInfo#getInfo(service, url)`](https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfo.java#L24-L26)).

For the other services, like SoundCloud and PeerTube, it would:
- if comments are disabled, return a custom `DisabledCommentsExtractor` that does nothing except `return`ing `true` when `isCommentsDisabled()` is called.
- otherwise, return the same `CommentsExtractor` as the current `service.getCommentsExtractor(url)`

Even though `service.getCommentsExtractor(url)` would lack some information (i.e. whether comments are disabled or not for that particular service), it should still return a `CommentsExtractor` that is able to work acceptably without that information (i.e. maybe provide an empty list of comments or try to extract whether comments are disabled or not in another way).

litetex:
@Stypox 
@B0pol 

Should I change anything for now?

I have no idea how to implement [Stypox's comment](https://github.com/TeamNewPipe/NewPipeExtractor/pull/652#issuecomment-863983515) the best way without breaking e.g. backwards compatibility, caching (I'm not sure if it's a good idea to return the extractor maybe just a supplier how to build it) and so on.
I also have not found videos/music with disabled comments at Soundcloud/Peertube.

I think a simple new field in ``StreamInfo`` e.g. ``Optional<Boolean> hasComments`` could also do the job:
* isEmpty → We don't know if the video has comments or not → the "user" (code using this) has to ask the ``CommentsExtractor``
* true/false → Video has (no) comments

Anyway currently I see two options:
1. We implement this PR for now and postpone the implementation of PeerTube, SoundCloud, etc (→ new issue and PR)
``CommentsExtractor#isCommentsDisabled`` could be marked as "experimental and may change".
2. We add the changes all in once, however due to the fact that - as mentioned above - I have no idea what's best so someone else with more experience should better work on this

I personally would go with option one to not create a PR "monolith"-monster.

Stypox:
Ok, let's go with option 1. Is this ready to be reviewed again?

litetex:
@Stypox 

> Is this ready to be reviewed again?

It never was not ready 😄 

Should I mark the new methods as ``experimental and may change in the future``?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
```diff
@@ -163,8 +222,28 @@ private String getDataString(Map<String, String> params) throws UnsupportedEncod
     }
 
     private String findValue(final String doc, final String start, final String end) {
```
Stypox, line 224: Note for a future PR: this should be moved to a separate string utils class and be tested

Stypox (changes requested):
> Should I mark the new methods as experimental and may change in the future?

Good idea

Other than that, it looks good to me

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
litetex: → https://github.com/TeamNewPipe/NewPipeExtractor/issues/679

litetex:
Done

Stypox (approved):

PR #653, opened 2021-06-13
Fix/live stream duration

TobiGr:

Follow up of #651

Fixed recognizing all streams as live streams
Also recognize streams with "NOW PLAYING" thumbnail overlay as live streams

Thanks to @litetex for noticing the regression introduced in #651.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
@@ -72,10 +72,14 @@ public class YoutubeStreamInfoItemExtractor implements StreamInfoItemExtractor {
         final JsonArray thumbnailOverlays = videoInfo.getArray("thumbnailOverlays");
         for (final Object object : thumbnailOverlays) {
             final JsonObject thumbnailOverlay = (JsonObject) object;
-            if (thumbnailOverlay.has("thumbnailOverlayNowPlayingRenderer")
-                    && thumbnailOverlay.getObject("thumbnailOverlayTimeStatusRenderer")
-                    .getString("style", EMPTY_STRING).equalsIgnoreCase("LIVE")) {
-                return cachedStreamType = StreamType.LIVE_STREAM;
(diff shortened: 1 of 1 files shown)


TobiGr:
don't merge. and don't test

opusforlife2:
So... mark it as a draft, then?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
```diff
@@ -72,10 +72,14 @@ public StreamType getStreamType() {
...
             final JsonObject thumbnailOverlay = (JsonObject) object;
-            if (thumbnailOverlay.has("thumbnailOverlayNowPlayingRenderer")
-                    || thumbnailOverlay.getObject("thumbnailOverlayTimeStatusRenderer")
-                    .getString("style", EMPTY_STRING).equalsIgnoreCase("LIVE")) {
-                return cachedStreamType = StreamType.LIVE_STREAM;
+            if (thumbnailOverlay.has("thumbnailOverlayNowPlayingRenderer")) {
+                final String overlayText = thumbnailOverlay
+                        .getObject("thumbnailOverlayTimeStatusRenderer")
+                        .getString("style", EMPTY_STRING);
+                if (overlayText.equalsIgnoreCase("LIVE")
```
litetex, line 79: When you reverse the comparing here to ``"LIVE".equalsIgnoreCase(overlayText)`` (and also in the next line) the code will be NPE save 😄

TobiGr: `getString("style", EMPTY_STRING)` returns either a string at the key "style" or an empty String, but not null

TobiGr:
I found out that the "now playing" overlay is not only present on live streams, but all streams. The overlay is shown when there is a play queue in YouTube and the stream is playing. Unfortunately, we cannot tell whether a stream is live stream or not when the "LIVE" badge is missing. That is the case when the stream is a never ending stream like a TV channel or news program and already running for a specific time (> 24 hours).

Stypox:
Oh, ok. Let's revert #651 then?

PR #676, opened 2021-07-03
Fix invalid CONSENT cookie value

TobiGr:


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
```diff
@@ -734,7 +734,8 @@ public static void addCookieHeader(final Map<String, List<String>> headers) {
     }
 
     public static String generateConsentCookie() {
-        return CONSENT_COOKIE + 100 + numberGenerator.nextInt(900);
+        final int statusCode = 100 + numberGenerator.nextInt(900);
+        return CONSENT_COOKIE + statusCode;
```
mhmdanas, line 738: How is the new code different from the old?
<sub>Please don't kill me if it's obvious</sub>

TobiGr: `CONSENT_COOKIE` is a String. That's why the result looks like `CONSENT=100580`.

mhmdanas: So if I understand correctly, an equivalent change would be keeping the method with one statement, but wrapping `100 + numberGenerator.nextInt(900)` in parenthesis, right?

TiA4f8R (approved):
Thank you for fixing this bug!

PR #694, opened 2021-07-28
Faster iframe api based player extraction.

### LINKED ISSUE #693: Make YoutubeJavaScriptExtractor more robust
`YoutubeJavaScriptExtractor` currently sometimes failes to find the javascript file , which contains the multiple decryption functions.  
See TODO in file https://github.com/TeamNewPipe/NewPipeExtractor/blob/14c179f343ab8ed703f81cc2956f4802958c6c30/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeJavaScriptExtractor.java#L74

see crash TeamNewPipe/NewPipe#6780

FireMasterK:
Uses the IFrame API to reduce the required download to less than 1/50 of the size.

Closes #693?

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeJavaScriptExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeJavaScriptExtractor.java
@@ -73,7 +73,7 @@ public class YoutubeJavaScriptExtractor {
         cachedJavaScriptCode = null;
     }
 
-    private static String extractJavaScriptUrl() throws ParsingException {
+    public static String extractJavaScriptUrl() throws ParsingException {
         try {
             final String iframeUrl = "https://www.youtube.com/iframe_api";
             final String iframeContent = NewPipe.getDownloader()
@@ -88,7 +88,7 @@ public class YoutubeJavaScriptExtractor {
         throw new ParsingException("Iframe API did not provide YouTube player js url");
     }
 
-    private static String extractJavaScriptUrl(final String videoId) throws ParsingException {
+    public static String extractJavaScriptUrl(final String videoId) throws ParsingException {
         try {
             final String embedUrl = "https://www.youtube.com/embed/" + videoId;
             final String embedPageContent = NewPipe.getDownloader()
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeJavaScriptExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeJavaScriptExtractor.java
@@ -79,7 +79,6 @@ public class YoutubeJavaScriptExtractor {
             return String.format("https://www.youtube.com/s/player/%s/player_ias.vflset/en_US/base.js", hash);
 
         } catch (Exception i) {
-            i.printStackTrace();
             try {
                 final String embedUrl = "https://www.youtube.com/embed/" + videoId;
                 final String embedPageContent = NewPipe.getDownloader()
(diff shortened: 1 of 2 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeJavaScriptExtractor.java
```diff
@@ -62,29 +62,42 @@ public static String extractJavaScriptCode() throws ParsingException {
...
+            final String iframeUrl = "https://www.youtube.com/iframe_api";
+            final String iframeContent = NewPipe.getDownloader()
+                    .get(iframeUrl, Localization.DEFAULT).responseBody();
+            final String hashPattern = "player\\\\\\/([a-z0-9]{8})\\\\\\/";
+            final String hash = Parser.matchGroup1(hashPattern, iframeContent);
 
+            return String.format("https://www.youtube.com/s/player/%s/player_ias.vflset/en_US/base.js", hash);
+
+        } catch (Exception i) {
+            i.printStackTrace();
```
XiangRongLin, line 74: remove

XiangRongLin (commented):
Do you have some examples to back  up the 1/50 size claim?

FireMasterK:
> Do you have some examples to back up the 1/50 size claim?

Yes, I do, in fact, it's a bit more than that:
image
image

XiangRongLin (commented):
Can you think of a way to somehow automatically test if your new implementation is actually used and it does not just always go to the `catch` block with the old implementation?

I can't think of anything right away, besides making everything public.

Edit: Reasoning being that the current tests either only cover your new implementation or the old one and not both

FireMasterK:
What would you like to happen if the new implementation fails?

Printing an error?

I could move then over to two methods so that they can be individually tested.

XiangRongLin:
> What would you like to happen if the new implementation fails?
>
> Printing an error?

Normally i would say log it, but because there is no logging setup => do nothing

> I could move them over to two methods so that they can be individually tested.

Yeah that's probably the easiest thing

FireMasterK:
Done! Another thing I noticed is that the current tests do not account for the fact that the `playerCode` can be cached, thereby making some tests poinless.

XiangRongLin:
Can you also add tests for the 2 methods

> Another thing I noticed is that the current tests do not account for the fact that the playerCode can be cached, thereby making some tests pointless.

You can create an issue for it. I have a rough idea on how to solve it, but finding the time to do it may prove hard

FireMasterK:
Done!

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeJavaScriptExtractorTest.java
```diff
@@ -19,6 +19,16 @@ public void setup() throws IOException {
...
 
+    @Test
+    public void testExtractJavaScriptUrlIframe() throws ParsingException {
+        YoutubeJavaScriptExtractor.extractJavaScriptUrl();
+    }
+
+    @Test
+    public void testExtractJavaScriptUrlEmbed() throws ParsingException {
+        YoutubeJavaScriptExtractor.extractJavaScriptUrl("d4IGg5dqeO8");
+    }
```
XiangRongLin, line 30: Can you add some assertions? As it stands, I could delete the whole implementation of the methods and these test would still pass.

Otherwise remove the tests/ undo you last commit.

FireMasterK: Oh, I thought it was fine since it would throw an exception in case it can't find an URL. Is it still necessary to add some assertions?

XiangRongLin: But what about the cases where it finds an url, but they don't lead to the player js file?

FireMasterK: Resolved! I've made it check whether it ends with `base.js`, there are a lot of URL formats for the player URL, and hence I avoided using a regex.

XiangRongLin (approved):
Good enough, since i can't think of anything else either. As in better tests/rewriting for better testability

PR #696, opened 2021-07-29
Expand regex to match n param decrypt function

XiangRongLin:

Temporary solution by expanding the regex, which extracts the decrypt function.

Depending on how long/difficult a proper solutions turns out, this could be unneeded.  @Stypox

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/utils/Parser.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/utils/Parser.java
@@ -84,6 +84,11 @@ public class Parser {
         return mat.find();
     }
 
+    public static boolean isMatch(Pattern pattern, String input) {
+        Matcher mat = pattern.matcher(input);
+        return mat.find();
+    }
+
     public static Map<String, String> compatParseMap(final String input) throws UnsupportedEncodingException {
         Map<String, String> map = new HashMap<>();
         for (String arg : input.split("&")) {
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/utils/Parser.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/utils/Parser.java
@@ -79,13 +79,13 @@ public class Parser {
     }
 
     public static boolean isMatch(String pattern, String input) {
-        Pattern pat = Pattern.compile(pattern);
-        Matcher mat = pat.matcher(input);
+        final Pattern pat = Pattern.compile(pattern);
+        final Matcher mat = pat.matcher(input);
         return mat.find();
     }
 
     public static boolean isMatch(Pattern pattern, String input) {
-        Matcher mat = pattern.matcher(input);
+        final Matcher mat = pattern.matcher(input);
         return mat.find();
     }
 
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/utils/StringUtil.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/utils/StringUtil.java
@@ -1,0 +1,47 @@
+package org.schabi.newpipe.extractor.utils;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+public class StringUtil {
+
+    private StringUtil() {
+    }
+
+    /**
+     * @param string The string to search in
+     * @param start A string from which to start searching.
+     * @return A substring where each '{' matches a '}'
+     * @throws IndexOutOfBoundsException If {@ string} does not contain {@code start}
+     */
+    @NonNull
+    public static String matchToClosingParenthesis(@NonNull final String string, @NonNull final String start) {
+        int startIndex = string.indexOf(start);
+        if (startIndex < 0) {
+            throw new IndexOutOfBoundsException();
+        }
+
+        startIndex += start.length();
+        int endIndex = startIndex;
+        while (string.charAt(endIndex) != '{') {
+            ++endIndex;
+        }
+        ++endIndex;
+
+        int openParenthesis = 1;
+        while (openParenthesis > 0) {
+            switch (string.charAt(endIndex)) {
+                case '{':
+                    ++openParenthesis;
+                    break;
+                case '}':
+                    --openParenthesis;
+                    break;
+                default:
+                    break;
+            }
+            ++endIndex;
+        }
+
+        return string.substring(startIndex, endIndex);
+    }
+}
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/utils/StringUtils.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/utils/StringUtils.java
@@ -9,10 +9,11 @@ public class StringUtils {
     }
 
     /**
-     * @param string The string to search in
+     * @param string The string to search in.
      * @param start A string from which to start searching.
-     * @return A substring where each '{' matches a '}'
+     * @return A substring where each '{' matches a '}'.
      * @throws IndexOutOfBoundsException If {@code string} does not contain {@code start}
+     * or parenthesis could not be matched .
      */
     @Nonnull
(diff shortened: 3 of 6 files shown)


XiangRongLin:
@Stypox I took the liberty to add your solution here directly

extractor/src/main/java/org/schabi/newpipe/extractor/utils/Parser.java
```diff
@@ -84,6 +84,11 @@ public static boolean isMatch(String pattern, String input) {
         return mat.find();
     }
 
+    public static boolean isMatch(Pattern pattern, String input) {
+        Matcher mat = pattern.matcher(input);
```
Stypox, line 88:
```suggestion
        final Matcher mat = pattern.matcher(input);
```

extractor/src/main/java/org/schabi/newpipe/extractor/utils/StringUtil.java
```diff
@@ -0,0 +1,47 @@
+package org.schabi.newpipe.extractor.utils;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+public class StringUtil {
```
Stypox, line 5: I think it should be `StringUtils`, looking at other files

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingDecrypterTest.java
```diff
@@ -19,6 +21,22 @@ public void setup() throws IOException {
         NewPipe.init(DownloaderTestImpl.getInstance());
     }
 
+    @Test
+    public void testExtractFunction__success() throws ParsingException {
+        String[] videoIds = {"jE1USQrs1rw", "CqxjzfudGAc", "goH-9MfQI7w", "KYIdr_7H5Yw", "J1WeqmGbYeI"};
```
Stypox, line 26:
```suggestion
        final String[] videoIds = {"jE1USQrs1rw", "CqxjzfudGAc", "goH-9MfQI7w", "KYIdr_7H5Yw", "J1WeqmGbYeI"};
```

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingDecrypterTest.java
```diff
@@ -19,6 +21,22 @@ public void setup() throws IOException {
...
+
+        final String encryptedUrl = "https://r6---sn-4g5ednek.googlevideo.com/videoplayback?expire=1626562120&ei=6AnzYO_YBpql1gLGkb_IBQ&ip=127.0.0.1&id=o-ANhBEf36Z5h-8U9DDddtPDqtS0ZNwf0XJAAigudKI2uI&itag=278&aitags=133%2C134%2C135%2C136%2C137%2C160%2C242%2C243%2C244%2C247%2C248%2C278&source=youtube&requiressl=yes&vprv=1&mime=video%2Fwebm&ns=TvecOReN0vPuXb3j_zq157IG&gir=yes&clen=2915100&dur=270.203&lmt=1608157174907785&keepalive=yes&fexp=24001373,24007246&c=WEB&txp=5535432&n=N9BWSTFT7vvBJrvQ&sparams=expire%2Cei%2Cip%2Cid%2Caitags%2Csource%2Crequiressl%2Cvprv%2Cmime%2Cns%2Cgir%2Cclen%2Cdur%2Clmt&alr=yes&sig=AOq0QJ8wRQIgW6XnUDKPDSxiT0_KE_tDDMpcaCJl2Un5p0Fu9qZNQGkCIQDWxsDHi_s2BEmRqIbd1C5g_gzfihB7RZLsScKWNMwzzA%3D%3D&cpn=9r2yt3BqcYmeb2Yu&cver=2.20210716.00.00&redirect_counter=1&cm2rm=sn-4g5ezy7s&cms_redirect=yes&mh=Y5&mm=34&mn=sn-4g5ednek&ms=ltu&mt=1626540524&mv=m&mvi=6&pl=43&lsparams=mh,mm,mn,ms,mv,mvi,pl&lsig=AG3C_xAwRQIhAIUzxTn9Vw1-vm-_7OQ5-0h1M6AZsY9Bx1FlCCTeMICzAiADtGggbn4Znsrh2EnvyOsGnYdRGcbxn4mW9JMOQiInDQ%3D%3D&range=259165-480735&rn=11&rbuf=20190";
+
+        for (String videoId : videoIds) {
```
Stypox, line 30:
```suggestion
        for (final String videoId : videoIds) {
```

Stypox (changes requested):
Yeah, looks good to me

extractor/src/main/java/org/schabi/newpipe/extractor/utils/StringUtil.java
```diff
@@ -0,0 +1,47 @@
+package org.schabi.newpipe.extractor.utils;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
```
TiA4f8R, line 3: Use the javax annotation instead.

TiA4f8R (changes requested):

extractor/src/main/java/org/schabi/newpipe/extractor/utils/StringUtil.java
XiangRongLin: good old auto import got me good

extractor/src/main/java/org/schabi/newpipe/extractor/utils/Parser.java
XiangRongLin: Im waiting for the day, where we can convert to kotlin. the keyword `final` should die. Takes up so much space for so little extra information

extractor/src/main/java/org/schabi/newpipe/extractor/utils/StringUtils.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    /**
+     * @param string The string to search in
+     * @param start A string from which to start searching.
+     * @return A substring where each '{' matches a '}'
+     * @throws IndexOutOfBoundsException If {@code string} does not contain {@code start}
+     */
+    @Nonnull
+    public static String matchToClosingParenthesis(@Nonnull final String string, @Nonnull final String start) {
```
Stypox, line 18: Mmmh, maybe add a couple of simple tests for this function

XiangRongLin: done

extractor/src/main/java/org/schabi/newpipe/extractor/utils/StringUtils.java
```diff
@@ -0,0 +1,49 @@
...
+        startIndex += start.length();
+        int endIndex = startIndex;
+        while (string.charAt(endIndex) != '{') {
+            ++endIndex;
+        }
+        ++endIndex;
+
+        int openParenthesis = 1;
+        int length = string.length();
+        while (openParenthesis > 0 && endIndex < length) {
```
Stypox, line 33: If you added `endIndex < length` here you should also add it to `while (string.charAt(endIndex) != '{') {` and to the `++endIndex;` three lines below. I left index checks out on purpose because an `IndexOutOfBounds` exception is better than returning a wrongly matched string that would anyway create exceptions down the line (e.g. when interpreting JS).

Stypox (changes requested):

extractor/src/main/java/org/schabi/newpipe/extractor/utils/StringUtils.java
XiangRongLin: I removed it. makes more sense this way, because then the catch block is used

Stypox (approved):

Thewisem:
Hmmm how can I get the .apk

PR #699, opened 2021-07-30
Fix YouTube music search.

### LINKED ISSUE #698: Youtube has changed the json format of results so YoutubeMusicSearchExtractor is no longer returning items
As of a day or two ago, there was a change in the json response from youtube music queries.  There appears to now be a tabbed renderer object inserted in the json tree.  

The tracked down where the parsing misses on the new format and the attached patch corrects for the change.

[NewPipeExtractor-fix-youtube-music-result-parse.patch.txt](https://github.com/TeamNewPipe/NewPipeExtractor/files/6909874/NewPipeExtractor-fix-youtube-music-result-parse.patch.txt)

FireMasterK:

Closes #698 

I've only tested this with the help of the tests.


Stypox:
The CI seems to fail, in part due to music search issues. Maybe you should update the mocks?

FireMasterK:
Something odd is going on, the tests were passing locally, so might be something related to different regions. I've used an EU IP and fixed it now.

Mocks are disabled for this test: https://github.com/TeamNewPipe/NewPipeExtractor/blob/633582384395cf8d267e4df4db452fc0c18b8e63/extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/search/YoutubeMusicSearchExtractorTest.java#L20

Stypox (approved):

PR #576, opened 2021-03-11
[YouTube] Extract replies to YouTube comments

golfinq:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).
- [x] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API.

A pull request which pulls the ctoken for replies to comments, downloads it, and turns them into NewPipeExtractor objects which is then attached to the original comment using a new object called CommentReplyExtractor. 
The associated NewPipe repository is found here: https://github.com/golfinq/NewPipe/tree/comment-replies ; currently is able to download, and display replies to comments. Though due to the messiness of the implementation I will create a draft PR once this extractor is merged instead of an open PR.

EDIT: added additional info

PR #703, opened 2021-08-04
Add support for extracting comment replies continuation

FireMasterK:

This takes an alternative approach to #576

@Stypox could you verify that this approach is okay?


Stypox:
Does it work also with comments continuations? Would this approach allow having both nested replies (for Peertube, I think) and replies continuations (when not all replies are loaded at once)?

FireMasterK:
> Does it work also with comments continuations

Yes! Nothing should break it, although I will check it manually. I would have to check this once again as I have not check if there's a different JSON structure for these continuations.

> Would this approach allow having both nested replies

Yes, each comment would have an option replies `Page`. In YouTube there's only one level, so it would be null there.

To get reply Comments, you just call `getMoreItems` with the replies `Page`.


> replies continuations

This would be a regular continuation / getNextPage(), nothing should change.

Stypox:
@FireMasterK ok, then this approach is good. Thank you :-)

FireMasterK:
Just one small issue - this implementation doesn't support multi-level threads in one page request, but I think that can be done later. (I'm not sure if any services do this either)

Mhowser:
Is this PR ready to be reviewed and merged soon?

extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfoItemExtractor.java
```diff
@@ -107,4 +108,13 @@ default boolean isUploaderVerified() throws ParsingException {
     default int getStreamPosition() throws ParsingException {
         return CommentsInfoItem.NO_STREAM_POSITION;
     }
+
+    /**
+     * The continuation page which is used to get comment replies from.
+     * @return the continuation Page for the replies
```
Stypox, line 114:
```suggestion
     * @return the continuation Page for the replies, or null if replies are not supported
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
```diff
@@ -212,10 +215,12 @@ private void collectCommentsFrom(final CommentsInfoItemsCollector collector,
             contents.remove(index);
         }
 
+        String jsonKey = contents.getObject(0).has("commentThreadRenderer") ? "commentThreadRenderer" : "commentRenderer";
```
Stypox, line 218:
```suggestion
        final String jsonKey = contents.getObject(0).has("commentThreadRenderer") ? "commentThreadRenderer" : "commentRenderer";
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
```diff
@@ -212,10 +215,12 @@ private void collectCommentsFrom(final CommentsInfoItemsCollector collector,
...
+
         final List<Object> comments;
         try {
             comments = JsonUtils.getValues(contents,
-                    "commentThreadRenderer.comment.commentRenderer");
+                    jsonKey);
```
Stypox, line 223: Put this on one line

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsInfoItemExtractor.java
```diff
@@ -29,6 +32,16 @@ public YoutubeCommentsInfoItemExtractor(final JsonObject json,
         this.timeAgoParser = timeAgoParser;
     }
 
+    private JsonObject getCommentRenderer() throws ParsingException {
```
Stypox, line 35: Why is there `throws ParsingException`?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsInfoItemExtractor.java
```diff
@@ -224,10 +237,20 @@ public String getUploaderName() throws ParsingException {
...
         }
     }
+
+    @Override
+    public Page getReplies() throws ParsingException {
+        try {
+            final String id = JsonUtils.getString(JsonUtils.getArray(json, "replies.commentRepliesRenderer.contents").getObject(0), "continuationItemRenderer.continuationEndpoint.continuationCommand.token");
+            return new Page(url, id);
+        } catch (final Exception e) {
+            return null;
```
Stypox, line 253: I guess if `YoutubeCommentsInfoItemExtractor` is used as a reply extractor this `return null` would kick in for YouTube, wouldn't it? Add a comment to explain this.

Stypox (changes requested):

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsInfoItemExtractor.java
FireMasterK: `JsonUtils` throws it, that's why.

Stypox: Okk

Stypox (approved):
Perfect, thank you :-D

Stypox:
This will be merged after release 0.20.10 is released

TobiGr:
Do we need to squash the commits?

Stypox:
I don't think that's needed

PR #707, opened 2021-08-06
Fix view count extraction for YouTube originals content.

### LINKED ISSUE #596: Views aren't extracted from YouTube Originals content
Note: It does not happen when looking at the video directly.
I've seen it on the trending and channel pages.

Example URL: https://www.youtube.com/channel/UCLXo7UDZvByw2ixzpQCufnA or https://piped.kavin.rocks/channel/UCLXo7UDZvByw2ixzpQCufnA

FireMasterK:

I've had a look at the JSON structure on a YouTube channel, it shouldn't cause any issues. However, since this extractor is likely used in every other place too, more testing is necessary.

Closes #596

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
@@ -254,7 +254,7 @@ public class YoutubeStreamInfoItemExtractor implements StreamInfoItemExtractor {
     @Override
     public long getViewCount() throws ParsingException {
         try {
-            if (videoInfo.has("topStandaloneBadge") || isPremium()) {
+            if (isPremium()) {
                 return -1;
             }
 


XiangRongLin:
Is there already an test for this? If not can you add it

XiangRongLin:
I checked out this branch and reverted the bug fix, but the added test are still passing => The tests are testing the wrong thing. 

Since a very specific bug needs to be tested, I personally would not go the whole way and implement `DefaultSearchExtractorTest`. This adds a lot of tests, which don't give more confidence. Instead I would just test that single thing

FireMasterK:
> I checked out this branch and reverted the bug fix, but the added test are still passing => The tests are testing the wrong thing.
> 
> Since a very specific bug needs to be tested, I personally would not go the whole way and implement `DefaultSearchExtractorTest`. This adds a lot of tests, which don't give more confidence. Instead I would just test that single thing

Do you know any other way? This bug only happens when you view the video on the search page, channel page, trending page, etc.

XiangRongLin:
With my current knowlegde I have no idea. I don't even know where to look I can see the bug inside the app/piped, since the issue description is very sparse

FireMasterK:
Ok, I just figured that it doesn't happen in the search result, hence why the test was passing.

To see this in action, open https://piped.kavin.rocks/c/vox and https://www.youtube.com/c/vox/videos

and find this video:
image

XiangRongLin:
@FireMasterK Are you going to adjust the test to your comment? Then it could be merged in my opinion

> Ok, I just figured that it doesn't happen in the search result, hence why the test was passing.

FireMasterK:
I'll do that tomorrow, I have to board a plane today!

litetex:
@FireMasterK 
Ping. Downgraded PR to draft as long as it's not finished.

FireMasterK:
The test has been removed already tho 🤔

litetex:
Sorry I'm not familiar with this issue... Which test? I didn't see one in the PR and the currently implemented seem to be working with mock-data...

XiangRongLin:
I didn't want the test removed, but adjusted.

> Ok, I just figured that it doesn't happen in the search result, hence why the test was passing.

With adjusting the test I meant changing it from checking the search result to checking whereever the problem does occur. 

Probably here
> To see this in action, open https://piped.kavin.rocks/c/vox and https://www.youtube.com/c/vox/videos

FireMasterK:
I don't think that's feasible since I don't know any channel which uploads only YouTube originals content.

XiangRongLin:
Is the problem that the data changes with the time? Because back then you linked a channel were you said that the problem could be observed.

If yes you can be the first on to use the `MockOnly` rule https://github.com/TeamNewPipe/documentation/blob/master/docs/05_Mock_tests.md#mock-only-tests
Then you just need to create the mock once and run the test against that scenario. Maybe also add what characteristics are important in the case that the mocks need to be regenerated

litetex:
Does this problem occur inside playlists? Maybe we can use one for a test. Playlists usually don't change that much.

FireMasterK:
> Does this problem occur inside playlists? Maybe we can use one for a test. Playlists usually don't change that much.

YouTube doesn't provide view counts in playlists, so that won't work!

I'll create a `MockOnly` later.

FireMasterK:
It looks like this now happens on the search page, so I added a test again!

FireMasterK:
Closing this as YouTube is now hiding the views on originals content, so this doesn't matter anymore.

PR #708, opened 2021-08-07
[YouTube] Always get the best thumbnail resolution

FireMasterK:


tsiflimagas:
This seems to be giving an error for channels with no banner.

XiangRongLin:
1. What is the reasoning behind always getting the best resolution?
2. Would this not lead to a higher network usage, which would not be desired on mobile data/bad internet connection?

TobiGr:
The goal should be to provide all image sizes and let the client choose the resolution it needs. See #268

FireMasterK:
I'll close this as this doesn't solve the problem in that case.

PR #709, opened 2021-08-08
 [YouTube] Support csv and zip subscription import (Google Takeout) 

talanc:

Extractor fix for issue TeamNewPipe/NewPipe#6757

This is the extractor PR for supporting the subscriptions.csv file from Google Takeout.
Also supports selecting a ZIP file (which will in turn find the subscriptions.csv file)
I introduced a new method on `SubscriptionExtractor` to determine what the stream is (e.g. json, zip, or csv):
`   List<SubscriptionItem> fromInputStream(@Nonnull final InputStream contentInputStream, String contentType)`

Ready for review and happy to make any changes to get this accepted.


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSubscriptionExtractor.java
```diff
@@ -10,17 +10,22 @@
...
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
 
 import javax.annotation.Nonnull;
 
 import static org.schabi.newpipe.extractor.subscription.SubscriptionExtractor.ContentSource.INPUT_STREAM;
 
 /**
- * Extract subscriptions from a Google takout export (the user has to get the JSON out of the zip)
+ * Extract subscriptions from a Google takeout export (the user has to get the JSON out of the zip)
```
Stypox, line 28:
```suggestion
 * Extract subscriptions from a Google takeout export
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSubscriptionExtractor.java
```diff
@@ -68,4 +97,69 @@ public String getRelatedUrl() {
...
+            throw new InvalidSourceException("Error reading contents of zip file", e);
+        }
+
+        throw new InvalidSourceException("Unable to find a subscriptions.csv file (try extracting and selecting the csv file)");
+    }
+
+    public static final String[] ENTRY_PATHS = {
+            "Takeout/YouTube and YouTube Music/subscriptions/subscriptions.csv",
+            "Takeout/YouTube y YouTube Music/suscripciones/suscripciones.csv" // There is a <NBSP> between YouTube and Music
+    };
```
Stypox, line 127: You can't rely on a list of paths. It will be different for each language. I'd navigate the whole zip searching for a valid file instead.

Stypox (changes requested):
Thank you for this needed usability improvement!

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSubscriptionExtractor.java
tbjgolden: A possible way to find this file is to look for files in the glob `Takeout/YouTube*/*/*.csv` and find one where the file starts with a match for this regex (this regex is a JS one, unsure of the Java equiv):

`/^.*\n([0-9A-Za-z_-]{20,}),[^,]*?\1,.+\n/`

tbjgolden: ^ intuition, this regex skips the first line, checks the second line to see if it starts with something that looks like a channel id, then looks for `/` followed by that same id again followed by a comma `,`

Stypox: Good idea, it should work out fine

talanc: Done--brief summary of changes:

CSV: after 5 lines, checks to make sure it's found subscription items (ensures the 2nd entry is a channel url)
if there's no items, then it exits out of the function, returning an empty list.

ZIP: iterates over every CSV file, gets the subscription items from that CSV file, if there's an error or no items in the list, it goes to the next CSV file in the zip.

talanc:
Thanks for your feedback -- ready for review again.

Stypox (changes requested):
Thank you, this approach looks good to me! Have you tested it? Also, please put `final` in front of every variable you never change after declaration.

talanc:
@Stypox 
Testing:
Yes.
It has unit tests for the CSV and ZIP import.
I've manually tested (via an emulator) with two different subscription files.
And I'm sure there will be another round of testing when it gets merged and the UI layer is tested: https://github.com/TeamNewPipe/NewPipe/pull/6882

Final:
I've added final to variables where appropriate.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSubscriptionExtractor.java
```diff
@@ -68,4 +97,110 @@ public String getRelatedUrl() {
...
+
+        int currentLine = 0;
+        String line = "";
+
+        try (BufferedReader br = new BufferedReader(new InputStreamReader(contentInputStream))) {
+            final List<SubscriptionItem> subscriptionItems = new ArrayList<>();
+
+            // Ignore header
+            currentLine = 1;
+            line = br.readLine();
```
Stypox, line 151:
```suggestion
        try (BufferedReader br = new BufferedReader(new InputStreamReader(contentInputStream))) {
            final List<SubscriptionItem> subscriptionItems = new ArrayList<>();
            int currentLine = 1; // Ignore header
            String line = br.readLine();
```

Stypox (changes requested):
Almost good ;-)

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSubscriptionExtractor.java
talanc: Locals currentLine and line are at the top of the function because the catch throws an exception with a detailed message using those locals.
Moving the locals to the try and I'll need to make the exception message more generic.

Knowing this, please let me know if you still want your suggested changed applied.

Stypox: Oh ok, I see, then leave it like this

Stypox (approved):
Looks good code-wise :-)

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSubscriptionExtractor.java
```diff
@@ -68,4 +97,110 @@ public String getRelatedUrl() {
...
+                final SubscriptionItem newItem = new SubscriptionItem(service.getServiceId(), channelUrl, channelTitle);
+                subscriptionItems.add(newItem);
+            }
+
+            return subscriptionItems;
+        } catch (IOException e) {
+            if (line == null) {
+                line = "<null>";
+            }
+            if (line.length() > 10) {
```
litetex, line 200:
```suggestion
            else if (line.length() > 10) {
```

if the line is ``null`` the check for the length > 10 is irrelevant

litetex (commented):
Code looks good 😄 
Test could be in theory parametrized, however because this is pain in JUnit 4 and creates a bunch of overhead, it's okay for me to go with a ``foreach``-loop.

talanc:
I think this is good to be merged now.

The failed tests are unrelated to my changes (a Peertube test is failing).

Stypox (approved):

Stypox:
@TobiGr please merge this, I don't have permission as apparently `snyk` failed. And I don't have access to Snyk.

XiangRongLin:
@Stypox Synk fails because of outdated jsoup version having a vulnerability. The version was already updated and merged. So if this is rebased, the check should succeed

Stypox:
Rebased, thank you!

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSubscriptionExtractor.java
```diff
@@ -68,4 +97,109 @@ public String getRelatedUrl() {
         }
         return subscriptionItems;
     }
+
+    public List<SubscriptionItem> fromZipInputStream(@Nonnull final InputStream contentInputStream)
+            throws ExtractionException {
+        final ZipInputStream zipInputStream = new ZipInputStream(contentInputStream);
```
Sanjay15693, line 103: This zipInputStream is not closed anywhere. You should close the stream after its use.

Stypox: Oh, right, didn't notice this

opusforlife2: @Stypox Hotfix?

Stypox: Nope, this was not released. We just need a PR with a try-with-resources

opusforlife2: Oh. Yeah. I mixed up the PRs.

Stypox: Thank you @Sanjay15693 :-)
See #721

Simon311:
Importing the CSV file doesn't work in the latest NewPipe available from F-Droid. 
But perhaps more importantly, my Google export only returned 6 channels out of hundreds. I know there's nothing you guys can do about it, but has anyone had this same issue? Is there a way to report this to Google, since this is a clear GDPR violation?

opusforlife2:
> Importing the CSV file doesn't work in the latest NewPipe available from F-Droid.

@Simon311 The version with this fix isn't released yet. You can install and test the Release Candidate by checking the pinned issue.

PR #710, opened 2021-08-09
Keep original exception when there is an error message

Stypox:

Just a small change so that we get exceptions with more information in case of a `ContentNotAvailableException` with an error message. That exception will now contain the original error that caused it.


TobiGr (approved):

XiangRongLin (approved):
Usefull error messages 🚀

XiangRongLin:
Test failure is from peertube => unrelated
So I'm merging

PR #714, opened 2021-08-15
Add verified info to ChannelInfo.

FireMasterK:


XiangRongLin:
Am i seeing it correctly that this is just adding the field and that parsing and filling that field will still need to be done?

FireMasterK:
They're already parsed in the extractor, I'm just adding it since I don't use extractor/extractions API directly in Piped, but I directly use `ChannelInfo.getInfo(url)`.

XiangRongLin (approved):
ah, i missed that the class is creating itself with the static methods.

PR #723, opened 2021-09-01
Extract Uploader's Avatar on YouTube and PeerTube

FireMasterK:

This pull request adds support for Channel Avatars extraction in the `StreamInfoItem` class.


Stypox:
Thank you @FireMasterK! Could you add some description? Does this add support for channel avatars of stream info items?

FireMasterK:
> Thank you @FireMasterK! Could you add some description? Does this add support for channel avatars of stream info items?

Done! It indeed does, this currently works in the Search and Trending pages I believe as YouTube doesn't provide these anywhere else.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/search/YoutubeSearchExtractorTest.java
```diff
@@ -320,4 +311,36 @@ public void testAtLeastOneVerified() throws IOException, ExtractionException {
...
+        @Override public String expectedOriginalUrlContains() { return "youtube.com/results?search_query=" + QUERY; }
+        @Override public String expectedSearchString() { return QUERY; }
+        @Nullable @Override public String expectedSearchSuggestion() { return null; }
+        @Override public InfoItem.InfoType expectedInfoItemType() { return InfoItem.InfoType.STREAM; }
+
+        @Test
+        public void testUploaderAvatar() throws IOException, ExtractionException {
+            final List<InfoItem> items = extractor.getInitialPage().getItems();
+            for (InfoItem item : items) {
+                assertNotNull(((StreamInfoItem) item).getUploaderAvatarUrl());
```
Stypox, line 342: This query could technically also return channel info items and the test would fail, wouldn't it? I'd add an `if (item instanceof StreamInfoItem)`

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamInfoItem.java
```diff
@@ -82,6 +83,15 @@ public void setUploaderUrl(String uploaderUrl) {
         this.uploaderUrl = uploaderUrl;
     }
 
+    @Nullable
+    public String getUploaderAvatarUrl() {
+        return uploaderAvatarUrl;
+    }
+
+    public void setUploaderAvatarUrl(String uploaderAvatarUrl) {
```
Stypox, line 91:
```suggestion
    public void setUploaderAvatarUrl(final String uploaderAvatarUrl) {
```

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/search/YoutubeSearchExtractorTest.java
```diff
@@ -320,4 +311,36 @@ public void testAtLeastOneVerified() throws IOException, ExtractionException {
...
+        @Override public String expectedUrlContains() { return "youtube.com/results?search_query=" + QUERY; }
+        @Override public String expectedOriginalUrlContains() { return "youtube.com/results?search_query=" + QUERY; }
+        @Override public String expectedSearchString() { return QUERY; }
+        @Nullable @Override public String expectedSearchSuggestion() { return null; }
+        @Override public InfoItem.InfoType expectedInfoItemType() { return InfoItem.InfoType.STREAM; }
+
+        @Test
+        public void testUploaderAvatar() throws IOException, ExtractionException {
+            final List<InfoItem> items = extractor.getInitialPage().getItems();
+            for (InfoItem item : items) {
```
Stypox, line 341:
```suggestion
            for (final InfoItem item : items) {
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
```diff
@@ -162,6 +163,17 @@ public String getUploaderUrl() throws ParsingException {
         return url;
     }
 
+    @Nullable
+    @Override
+    public String getUploaderAvatarUrl() throws ParsingException {
+
+        if(videoInfo.has("channelThumbnailSupportedRenderers")) {
+            return JsonUtils.getArray(videoInfo, "channelThumbnailSupportedRenderers.channelThumbnailWithLinkRenderer.thumbnail.thumbnails").getObject(0).getString("url");
```
Stypox, line 171: Better to break it into two
```suggestion
            return JsonUtils.getArray(videoInfo, "channelThumbnailSupportedRenderers.channelThumbnailWithLinkRenderer.thumbnail.thumbnails")
                .getObject(0).getString("url");
```

Stypox (changes requested):
Looks almost good to me :-)

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/search/YoutubeSearchExtractorTest.java
Stypox: Oh, didn't see that one, good

Stypox (approved):

extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubeStreamInfoItemExtractor.java
```diff
@@ -54,6 +56,12 @@ public String getUploaderUrl() throws ParsingException {
                 .fromId("accounts/" + name + "@" + host, baseUrl).getUrl();
     }
 
+    @Nullable
+    @Override
+    public String getUploaderAvatarUrl() throws ParsingException {
+        return null;
```
TobiGr, line 62:
```suggestion
        final JsonObject account = item.getObject("account");
        if (account.has("avatar") && !account.isNull("avatar")) {
            return baseUrl + account.getObject("avatar").getString("path");
        }
        return null;
```

TobiGr (changes requested):
I think this can also be tested roughly in `DefaultTests.defaultTestListOfItems()`.

Please remove  `throws ParsingException` when overriding the method and returning `null`

FireMasterK:
> I think this can also be tested roughly in DefaultTests.defaultTestListOfItems().

What about services which in which it's not extracted in? Another thing to note is that even on YouTube, it's not provided everywhere.

TobiGr:
Testing this should be similar to testing the thumbnail URL
```Java
            final String uploaderAvatarUrl = item.getUploaderAvatarUrl();
            if (!isNullOrEmpty(uploaderAvatarUrl)) {
                assertIsSecureUrl(uploaderAvatarUrl);
            }
```

FireMasterK:
> Testing this should be similar to testing the thumbnail URL
> 
> ```java
>             final String uploaderAvatarUrl = item.getUploaderAvatarUrl();
>             if (!isNullOrEmpty(uploaderAvatarUrl)) {
>                 assertIsSecureUrl(uploaderAvatarUrl);
>             }
> ```

Added, thanks for the clarification on what you meant!

PR #725, opened 2021-09-03
Extract Channel Avatar in related videos.

FireMasterK:

This would be an addition to #723


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
```diff
@@ -172,6 +172,11 @@ public String getUploaderAvatarUrl() throws ParsingException {
                     .getObject(0).getString("url");
         }
 
+        if(videoInfo.has("channelThumbnail")) {
```
TobiGr, line 175: Also in the statement above
```suggestion
        if (videoInfo.has("channelThumbnail")) {
```

FireMasterK: Done!

TobiGr (approved):

PR #731, opened 2021-09-22
Extract Video Short Description in YouTube.

FireMasterK:
In Trending, and Search results.

A test wasn't added for trending, since not all videos have descriptions.


Stypox (dismissed):
I had a doubt about returning `null` or an empty string when there is no short description, but by looking at other extractors there doesn't seem to be a consistent way to do it, so it is ok like this.
Code looks good, thank you!

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
```diff
@@ -288,4 +288,17 @@ private OffsetDateTime getDateFromPremiere() throws ParsingException {
...
+
+    @Nullable
+    @Override
+    public String getShortDescription() throws ParsingException {
+
+        if(videoInfo.has("detailedMetadataSnippets"))
+            return getTextFromObject(videoInfo.getArray("detailedMetadataSnippets").getObject(0).getObject("snippetText"));
+
+        if(videoInfo.has("descriptionSnippet"))
+            return  getTextFromObject(videoInfo.getObject("descriptionSnippet"));
```
TiA4f8R, line 300:
```suggestion
        if (videoInfo.has("detailedMetadataSnippets")) {
            return getTextFromObject(videoInfo.getArray("detailedMetadataSnippets").getObject(0).getObject("snippetText"));
        }

        if (videoInfo.has("descriptionSnippet")) {
            return getTextFromObject(videoInfo.getObject("descriptionSnippet"));
        }
```

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamInfoItem.java
```diff
@@ -92,6 +93,14 @@ public void setUploaderAvatarUrl(final String uploaderAvatarUrl) {
         this.uploaderAvatarUrl = uploaderAvatarUrl;
     }
 
+    public String getShortDescription() {
+        return shortDescription;
+    }
+
+    public void setShortDescription(String shortDescription) {
```
TiA4f8R, line 100:
```suggestion
    public void setShortDescription(final String shortDescription) {
```

TiA4f8R (dismissed):
Just some code style issues and this can be merged.

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamInfoItemExtractor.java
```diff
@@ -116,4 +116,14 @@
...
 
+
+    /**
+     * Get the video's short description.
+     *
+     * @return The video's short description or {@code null} if not provided by the service.
+     * @throws ParsingException if there is an error in the extraction
+     */
+    @Nullable
+    String getShortDescription() throws ParsingException;
```
litetex, line 127: Why don't implement a default method here?
Would remove the need for implementing this in each service
```java
    @Nullable
    @Override
    public String getShortDescription() {
        return null;
    }
```

litetex (changes requested):

litetex (approved):
LGTM

PR #732, opened 2021-09-23
Get IOS HLS manifest through async means

FireMasterK:

Although this makes an additional request that's unnecessary for NewPipe, it's done asynchronously so that there's no performance impact.


Stypox:
This PR aims at adding another method to get the manifest, so that if the other two break, this might still work?

FireMasterK:
Yes, except normally, the manifest is only added on live streams, however with this, all videos will have an HLS manifest.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -491,13 +479,19 @@ public String getDashMpdUrl() throws ParsingException {
...
+
         if (desktopStreamingData != null) {
-            return desktopStreamingData.getString("hlsManifestUrl");
-        } else if (mobileStreamingData != null) {
-            return mobileStreamingData.getString("hlsManifestUrl");
-        } else {
-            return EMPTY_STRING;
+            hlsManifestUrl = desktopStreamingData.getString("hlsManifestUrl", EMPTY_STRING);
+        }
+        if (mobileStreamingData != null && hlsManifestUrl.isEmpty()) {
```
Stypox, line 487: These could as well be `else if`s, can't they?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -706,6 +700,13 @@ public void onFetchPage(@Nonnull final Downloader downloader)
                 .done())
                 .getBytes(UTF_8);
 
+        CompletableFuture iosTask = CompletableFuture.supplyAsync(() -> {
```
Stypox, line 703: I am not sure this is a good solution: what if the task hasn't finished when `getHlsUrl()` is called? If the extractor was all built with async means it would integrate good, but in this case there is nothing waiting for the async task to complete. Also, futures are not used anywhere else in the extractor, by doing a fast search.

Stypox (commented):
I see. I think this increases data usage, though, without apparent improvents on the client side. I don't fully understand when this would become useful, sorry ;-)

TiA4f8R:
Superseeded by #780.

PR #733, opened 2021-09-23
Extract isLive on YouTube.

FireMasterK:


Stypox:
Can't you just check whether the stream type is either a video live or an audio live, instead of having a full new method with duplicate checks?

FireMasterK:
I see, I didn't even know that was even possible!

PR #739, opened 2021-10-13
Changes for "Show Replies"

golfinq:
A small change needed for the associated pull request in NewPipe (https://github.com/TeamNewPipe/NewPipe/pull/7244), which stores whether or not the replies have been cached to prevent re-downloading

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfoItem.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfoItem.java
@@ -4,8 +4,6 @@ import org.schabi.newpipe.extractor.InfoItem;
 import org.schabi.newpipe.extractor.Page;
 import org.schabi.newpipe.extractor.localization.DateWrapper;
 
-import java.util.List;
-
 import javax.annotation.Nullable;
 
 public class CommentsInfoItem extends InfoItem {
@@ -26,8 +24,6 @@ public class CommentsInfoItem extends InfoItem {
     private int streamPosition;
     @Nullable
     private Page replies;
-    @Nullable
-    private List<CommentsInfoItem> repliesInfoList;
     private boolean repliesOpen = false;
 
     public static final int NO_LIKE_COUNT = -1;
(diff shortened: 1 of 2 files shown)


litetex:
> @Redirion I tried it, and I ran into the `NetworkOnMainThreadException` and I am not familiar with android threads

https://stackoverflow.com/questions/6343166/how-can-i-fix-android-os-networkonmainthreadexception

Btw: I'm pretty sure that this PR will cause problems regarding the comment loading performance and resources required for holding the replies in memory.

golfinq:
@TobiGr Oh my god, that is what I was missing. I'll see if I can incorporate RXJava to not have to cache the replies to comments. I am doing some reading to learn about it here: https://github.com/ReactiveX/RxJava/tree/3.x/docs

golfinq:
> 
> 
> I don't understand why you want to do it this way.
> 
> How I imagined it to be used was:
> 
> ```
> CommentsInfo info = CommentsInfo.getInfo(...);
> 
> // Only called when the user clicks on show more
> CommentsInfo replies = info.getReplies() != null ? CommentsInfo.getMoreItems(info, info.getReplies()) : null;
> ```

This was quite helpful in getting off the ground, thanks!

extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfoItem.java
```diff
@@ -149,4 +150,10 @@ public int getStreamPosition() {
     public void setReplies(@Nullable Page replies) { this.replies = replies; }
 
     public Page getReplies() { return this.replies; }
+
+    public void setRepliesOpen(boolean repliesOpen) {
+        this.repliesOpen = repliesOpen;
+    }
+
+    public boolean getRepliesOpen() {return this.repliesOpen; }
```
Stypox, line 158: This should not be done in the Extractor code, since this is just a UI thing

Stypox (changes requested):

opusforlife2:
Closing this for the same reason as #7244.

PR #751, opened 2021-11-17
Attempt to parallelize YouTube Stream extractor

FireMasterK:

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
@@ -50,6 +50,7 @@ import org.mozilla.javascript.ScriptableObject;
 import org.schabi.newpipe.extractor.MediaFormat;
 import org.schabi.newpipe.extractor.MetaInfo;
 import org.schabi.newpipe.extractor.MultiInfoItemsCollector;
+import org.schabi.newpipe.extractor.NewPipe;
 import org.schabi.newpipe.extractor.StreamingService;
 import org.schabi.newpipe.extractor.downloader.Downloader;
 import org.schabi.newpipe.extractor.exceptions.AgeRestrictedContentException;
@@ -99,6 +100,8 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 import java.util.stream.Collectors;
 
 import javax.annotation.Nonnull;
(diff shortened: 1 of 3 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -731,44 +734,73 @@ public void onFetchPage(@Nonnull final Downloader downloader)
...
+        });
+
+        try {
+            playerFetchFuture.get();
+            nextFuture.get();
+            fetchAndroidFuture.get();
+            fetchDesktopSts.get();
+        } catch (InterruptedException | ExecutionException e) {
+            e.printStackTrace();
         }
```
FireMasterK, line 804: I need a better solution for this, should I override the class?

XiangRongLin:
1. Are there currently some performance problems that would warrant such a change and it's developer overhead?
2. If there are, how does it look like with this change. Any statistics or the like?

ps: I did NOT take a real look at the code

Stypox:
Why use Java futures manually? I'd rely on RxJava instead (like we already do in the app)

litetex:
> Why use Java futures manually? I'd rely on RxJava instead (like we already do in the app)

I think there is no RXJava inside this project.

Stypox:
> I think there is no RXJava inside this project.

Yeah, I know. I'd propose to use it, what do you think?

litetex:
> Yeah, I know. I'd propose to use it, what do you think?

I don't know. 
* RxJava was built for event based programming however there are no events here. Just asynch tasks.
* It would add a new dependency here; (Larger) dependencies tend to introduce problems...
* I might be biased because personally the most RxJava code I read so far in NewPipe was [horribly complex](https://github.com/TeamNewPipe/NewPipe/blob/d8236bbedd4713f777e90d249850b31c1f71cf3f/app/src/main/java/org/schabi/newpipe/local/feed/service/FeedLoadService.kt#L177-L252) or caused problems that have been different for debugging

I'm also fine with the current way of using the java native way.
However before we get technical 👉 https://github.com/TeamNewPipe/NewPipeExtractor/pull/751#issuecomment-971826701

FireMasterK:
Was there any conclusion to what library to use? Should we go with the default Java API? I'd like to rework on this, one last time.

extractor/src/main/java/org/schabi/newpipe/extractor/utils/ExceptionUtils.java
```diff
@@ -0,0 +1,16 @@
+package org.schabi.newpipe.extractor.utils;
+
+public class ExceptionUtils {
+
+    // From Apache Commons Lang: https://commons.apache.org/proper/commons-lang/javadocs/api-3.10/org/apache/commons/lang3/exception/ExceptionUtils.html#rethrow-java.lang.Throwable-
+
+    public static <R> R rethrow(final Throwable throwable) {
+        // claim that the typeErasure invocation throws a RuntimeException
+        return ExceptionUtils.<R, RuntimeException>typeErasure(throwable);
```
Stypox, line 9: My mind is blown :exploding_head: 
Is this really needed?

Stypox (commented):
@FireMasterK yes, let's use the default Java API. The current code structure looks good :-)

FireMasterK:
> Are there currently some performance problems that would warrant such a change and its developer overhead?

Yes, currently in NewPipe's case, a minimum of 2 requests are made regardless. We parallelize them, to reduce the time taken by half. Another additional request may be made - ios or android client, depending if it is a live stream or not.

In Piped's case, a minimum of 4 requests are made since I force the android and ios player fetch.

> If there are, how does it look like with this change. Any statistics or the like?

Piped's response times have dropped by ~400-500ms on average (it depends on how fast youtube's server responds), this should benefit NewPipe too.

FireMasterK:
I tried updating the mocks, but it appears to be unable to find some requests, anyone have any idea why this might be happening?

Stypox (approved):
Some time has passed, sorry... On a second thought this implementation is fine. @AudricV do you have anything against merging this, after it has been rebased? ;-)

FireMasterK:
It's still a draft since it would have issues with mocks.

What I'm going to do to fix them is remove the `cpn` variable before checking if a mock request matches a response.

AudricV:
Closing this pull request, as making more parallel requests may increase chances of getting hit by the anti-bot sign in wall of YouTube.

PR #755, opened 2021-11-30
Cleanup: Remove old ways of getting YT dis/likes

litetex:

Some clean/followups after #753; Content
* Removed old way of getting likes (was disabled by YT - as predicted - on [2021-12-13](https://support.google.com/youtube/thread/134791097/update-to-youtube-dislike-counts))
* Removed dislike calculation because it no longer works (no average rating)
* Fixes Exception from #7525
* Deduplicated a ton of code inside the StreamExtractor(s) by using default/non-abstract methods
* Fixed tests


opusforlife2:
I suppose it will have to be a hotfix.

litetex:
I don't think we have to hotfix this as everything should still work as expected.
However this PR removes the code that by then will be (likely) irrelevant.
So this is just an improvement :)

opusforlife2:
Oh, I was confused. The calculation method will still work.

TobiGr (approved):
Looks good.

PR #760, opened 2021-12-08
Fix test

litetex:

Fixes all of the currently broken tests.


extractor/src/test/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelperTest.java
```diff
@@ -20,6 +24,14 @@ public void assertThatHardcodedClientIdIsValid() throws Exception {
...
     }
 
+    @Test
+    public void assertHardCodedClientIdMatchesCurrentClientId() throws IOException, ExtractionException {
+        assertEquals(
+                "Hardcoded client doesn't match extracted clientId",
+                SoundcloudParsingHelper.HARDCODED_CLIENT_ID,
+                SoundcloudParsingHelper.clientId());
+    }
+
```
TobiGr, line 34: Isn't that exactly what `assertThatHardcodedClientIdIsValid()` is doing? If the hardcoded id is not valid, a new one is extracted. In this case, both test will fail. On the other hand, if it is valid, it is used by `clientId()` and thus both tests complete successfully.

litetex: Yes but the difference is it actually shows you the extracted id, so you can fix it on the fly.
No need for an Android studio start and debugging ;)

TobiGr (approved):

PR #767, opened 2021-12-27
[media.ccc.de] Fix live streams extraction

TobiGr:

The API was slightly changed and only HLS streams are provided anymore.

Add a basic test for to check whether the required streamInfo fields are available.


extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCLiveStreamExtractor.java
```diff
@@ -165,13 +165,14 @@ public String getDashMpdUrl() throws ParsingException {
...
-        //       Make getHlsUrl() and getDashMpdUrl() return lists of VideoStreams, so the user can choose a resolution.
+        //       Make getHlsUrl() and getDashMpdUrl() return lists of VideoStreams
+        //       to allow selecting a resolution.
         for (int s = 0; s < room.getArray("streams").size(); s++) {
             final JsonObject stream = room.getArray("streams").getObject(s);
             if (stream.getString("type").equals("video")) {
-                final String resolution = stream.getArray("videoSize").getInt(0) + "x"
-                        + stream.getArray("videoSize").getInt(1);
-                if (stream.has("hls")) {
+                /*final String resolution = stream.getArray("videoSize").getInt(0) + "x"
```
litetex, line 173: Note: Dead code

TobiGr: Removed

extractor/src/test/java/org/schabi/newpipe/extractor/services/media_ccc/MediaCCCLiveStreamExtractorTest.java
```diff
@@ -0,0 +1,45 @@
...
+        liveKiosk = MediaCCC.getKioskList().getExtractorById("live", null);
+        liveKiosk.fetchPage();
+        liveItems = liveKiosk.getInitialPage().getItems();
+        Assume.assumeFalse(
+                "Received an empty list of live streams. Skipping MediaCCCLiveStreamExtractorTest",
+                liveItems.isEmpty());
+    }
+
+    @Test
+    public void testRequiredStreamInfo() throws Exception {
```
litetex, line 36: Shouldn't there be an assert?

TobiGr: see the comment below

litetex: A test that runs into an exception != A test that fails (through an ``AssertionError``).
See also https://rules.sonarsource.com/java/RSPEC-2699

``assertDoesNotThrow`` would be helpful here however as we aren't using JUnit 5 that's not available. (I think I should do a PR for that).

I would solve the problem like this:
```java
try {
    ...
} catch (final Exception e) {
    Assert.fail("An exception occurred. This should never happen.")
}
```

TobiGr: Done

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamInfo.java
```diff
@@ -181,13 +181,16 @@ private static StreamInfo extractStreams(StreamInfo streamInfo, StreamExtractor
 
         // Either audio or video has to be available, otherwise we didn't get a stream
         // (since videoOnly are optional, they don't count).
-        if ((streamInfo.videoStreams.isEmpty()) && (streamInfo.audioStreams.isEmpty())) {
+        if (streamInfo.videoStreams.isEmpty()
+                && streamInfo.audioStreams.isEmpty()
+                && isNullOrEmpty(streamInfo.getHlsUrl())
+                && isNullOrEmpty(streamInfo.getDashMpdUrl())) {
```
Stypox, line 187: Mmmh, I think before only videoStreams and audioStreams were checked because in any other case the stream is a livestream. So NewPipe's player expected to need to use hls/dash urls only on livestreams. Does this change break that workflow, or is the check being done differently by NewPipe and so this is not an issue?

Stypox (changes requested):
Thank you! Could you provide an apk?

Mybeterlife-continue2025mylife (approved):
https://git.savannah.gnu.org/cgit/directory.git/commit/?id=222f7c16acc41bc94b3acafbe719442c8131a73d
https://eu.docworkspace.com/d/sIPnI_KunAtC1g74G?sa=601.1129

PR #769, opened 2022-01-02
Update nanojson

TobiGr:

Update our fork of nanojson. This includes a few dependency updates as well as smaller improvements from upstream.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,7 +27,7 @@ allprojects {
     }
 
     ext {
-        nanojsonVersion = "1d9e1aea9049fc9f85e68b43ba39fe7be1c1f751"
+        nanojsonVersion = "1.7-np"
         spotbugsVersion = "4.5.2"
         junitVersion = "4.13.2"
     }


litetex (commented):
2 points I have to discuss here:
1. All tests are failing there seems to be something wrong
2. What are the changes in the new "version"?

FireMasterK:
I just want to add that the current fork of nanojson uses significantly more CPU than upstream!

In Piped, I'm now using a personal fork since https://github.com/TeamPiped/Piped-Backend/pull/357, the difference is *MASSIVE*! (probably well within )

I did this after seeing that nanojson was using a lot of CPU in the `nextToken` method. The NewPipe fork has changes, calling this method more times (for JS parsing?), which could be the cause of this.

I don't have a before, but here's the after (it was ~1800 seconds before, within 10 minutes):
After:
image

opusforlife2:
@FireMasterK If the only difference between this version and your fork is upstream changes, could they not be added to this fork?

FireMasterK:
> @FireMasterK If the only difference between this version and your fork is upstream changes, could they not be added to this fork?

I suppose they could, but I only tested it on YouTube (in Piped), I'm not sure if other services require the JS json extraction. In my fork, I have the upstream changes and have cherry-picked only this commit: https://github.com/TeamNewPipe/nanojson/commit/d5da581ec79e6cc12962da8b92749038143f8e08

Stypox:
Closing in favor of  #981

PR #786, opened 2022-02-01
[Youtube] Make throttling decryption more resilient to api change

XiangRongLin:
- [] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingDecrypter.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingDecrypter.java
@@ -38,7 +38,9 @@ public class YoutubeThrottlingDecrypter {
     private static final Pattern FUNCTION_NAME_PATTERN = Pattern.compile(
             "b=a\\.get\\(\"n\"\\)\\)&&\\(b=(\\S+)\\(b\\),a\\.set\\(\"n\",b\\)");
 
-    private static final Map<String, String> nParams = new HashMap<>();
+    private static final Map<String, String> N_PARAMS_CACHE = new HashMap<>();
+    private static String FUNCTION;
+    private static String FUNCTION_NAME;
 
     private final String functionName;
     private final String function;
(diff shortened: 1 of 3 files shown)


litetex (commented):
Not sure how using static code is making the API more resilient...

I'm usually always against static stuff because it can't e.g. be overridden or simply reset by throwing away the related object.

XiangRongLin:
The reason for it being static is, that the function and function name is always the same while the extractor is in use. Reason for THAT is that `YoutubeJavaScriptExtractor.extractJavaScriptCode(videoId);` is cached. So subsequent calls will always return the same results.
Knowing that it always is the same, I want to cache it. I don't see a way to cache it, beside making `FUNCTION` and `FUNCTION_NAME` static. And at that point the class has no member variables. So I made it static because I remeber reading an Android guide recommending not creating unecessary objects, since phones are very resource constrained.

---

> Not sure how using static code is making the API more resilient...

The static code has no relation to the resilience. The resilience comes from wrapping the whole decryption process in a try catch. It also does not make the API more resilient. It is making it more resilient against changes in the Youtube API

Stypox (commented):
Code looks good to me. I agree with you about making things static, it's already being done in other places and it's good to cache the things needed for decryption.

Since the `YoutubeThrottlingDecrypter` is only meant for internal use, why deprecate things and not just remove them directly?

XiangRongLin:
> Since the YoutubeThrottlingDecrypter is only meant for internal use, why deprecate things and not just remove them directly?

Because there is the n params cache, which had the reset method provided, which is meant to be used externaly. So who knows if somebody came to the idea to use the rest of the methods.

XiangRongLin:
So anything blocking this? I would like to get this merged, so we don't get another fiasco like before because of this

opusforlife2:
Use the request review button?

Stypox (approved):
Commit my small suggestion and then this is ready

Stypox (approved):

PR #685, opened 2021-07-19
Add getMixUrl method to StreamExtractors

Stypox:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).
- [x] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API.

I added the `getMixUrl(MixType)` method to every `StreamExtractor`, along with an `enum` for mix types (video, music, channel, album), following https://github.com/TeamNewPipe/NewPipe/issues/3583#issuecomment-876712109. This PR is a draft and still needs work, also the frontend UI is still a TODO, but feel free to give feedback ;-).

- Youtube video with mix: https://www.youtube.com/watch?v=l-mPoJe8wZQ
- Youtube video mix: https://www.youtube.com/watch?v=l-mPoJe8wZQ&list=RDl-mPoJe8wZQ
- Youtube music mix: https://music.youtube.com/watch?v=l-mPoJe8wZQ&list=RDAMVMl-mPoJe8wZQ
- Youtube channel mix: ~~https://www.youtube.com/watch?v=l-mPoJe8wZQ&list=RDCMUCf5q0cbFOLbphljteZ9d4Pw~~ (I don't know why it doesn't work, this needs investigation) https://www.youtube.com/watch?v=9iHM6X6uUH8&list=RDCMUC_aEa8K-EOJ3D6gOs7HcyNg
- Soundcloud track with station (click the More button): https://soundcloud.com/jim-yosef/jim-yosef-firefly-ncs-release
- Soundcloud station (still unsupported by the extractor): https://soundcloud.com/stations/track/jim-yosef/jim-yosef-firefly-ncs-release
- Bandcamp track with related album (in NewPipe just considered as a playlist): https://jimyosef.bandcamp.com/track/dawn

Fixes https://github.com/TeamNewPipe/NewPipe/issues/3583
Fixes https://github.com/TeamNewPipe/NewPipe/issues/3250

PR #768, opened 2021-12-27
JUnit 5

litetex:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

* This PR drops JUnit 4 and introduces the new and improved JUnit 5.
* Minor code reworks and fixes on the fly

PR #788, opened 2022-02-02
Add MixInfoItem and extract YouTube mixes in related items

Stypox:
- [no need] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API.

Replaces #685, thank you @opusforlife2 for the ping there ;-)

### Changes
- ~This PR adds a `mix` subpackage that contains a `MixInfoItem`, i.e. a new type of `InfoItem` separated from `PlaylistInfoItem` that adds an abstract `getMixType` function and nothing else to the plain `InfoItem` (since mixes do not have an uploader or a stream count). The mix type can be one of: stream, music, channel (as described [here](https://github.com/TeamNewPipe/NewPipeExtractor/pull/685#issue-948001895)). The `mix` subpackage also contains the `MixInfoItemCollector` and the base `MixInfoItemExtractor`.~ See https://github.com/TeamNewPipe/NewPipeExtractor/pull/788#issuecomment-1043226535
- I renamed `InfoItemsSearchCollector` into `MultiInfoItemsCollector` and added support for also collecting `MixInfoItemExtractor`s. I could do this since the essence of that class was just to combine some collectors into one, so it had nothing to do with "search" but could be useful everywhere we want to return a list with multiple types of items.
- [YouTube] I implemented mixes in related searches in YouTube. Since some code in `YouTubeMixInfoItemExtractor` would have been duplicated with `YoutubeStreamInfoItemExtractor` and classes related to mixes, I created some common functions in `YouTubeParsingHelper` (the TODO there is unrelated to this PR and was just copied over). Related videos have `compactVideoRenderer` as a key in the json object, while related mixes have `compactRadioRenderer` (**note: still untested for channel, music and my mixes**)

@XiangRongLin what do you think about this? :-)


Stypox:
Actually, I'm not so sure whether `MixInfoItem` should be completely separated from `PlaylistInfoItem`: maybe we should make it so that `MixInfoItem extends PlaylistInfoItem`? This way backward compatibility would be better (no need to instantly adapt to a new `InfoItem` type, since `MixInfoItem` could just be used as a `PlaylistInfoItem`) and maybe some services (will) have an uploader and/or a video count (which can just be set to default values in case they don't exist). What do you think?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
```diff
@@ -278,7 +289,7 @@ public static boolean isYoutubeChannelMixId(@Nonnull final String playlistId) {
     @Nonnull
     public static String extractVideoIdFromMixId(@Nonnull final String playlistId)
             throws ParsingException {
-        if (playlistId.startsWith("RDMM")) { // My Mix
+        if (isYoutubeMyMixId(playlistId)) { // My Mix
             return playlistId.substring(4);
 
         } else if (isYoutubeMusicMixId(playlistId)) { // starts with "RDAMVM" or "RDCLAK"
```
XiangRongLin, line 295: The comments here can be removed, the method name or method itself now contains the information

extractor/src/main/java/org/schabi/newpipe/extractor/InfoItem.java
```diff
@@ -68,7 +68,8 @@ public String toString() {
     public enum InfoType {
```
XiangRongLin, line 68: Writing here, so a thread is created:

---

concerning your question https://github.com/TeamNewPipe/NewPipeExtractor/pull/788#issuecomment-1028713839

For me it It depends on what you want mixes to be. Are they a specialization of a playlist or are they a separate concept.  
I see NewPipe through the YouTube only lens, so for me it's a separate concept. A mix does not have a uploader or streamcount, which are the defining elements of playlists.

> maybe some services (will) have an uploader and/or a video count

Does any of the current services have it? If not do we plan on adding new services?

> backward compatibility

This should be the place that is affected: https://github.com/TeamNewPipe/NewPipeExtractor/blob/857b2f39e87353f4617706829dd42231e457a87e/extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamInfo.java#L631

Seeing as how it is not documented, which subtypes are going to be returned, one could hope that users of that functions already check for the type before using items of it. Like NewPipe itself does: https://github.com/TeamNewPipe/NewPipe/blob/9c9730b15217c7e8d3b2ccdecd658361215ccae3/app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java#L212  
Otherwise I find that this is an acceptable breaking change, since one can simply add such a check into their own app. It simply needs to be communicated properly. 
The alternative of adding a feature flag that is turned off by default, which toggles this [if statement](https://github.com/Stypox/NewPipeExtractor/blob/9f9c6cb49a6c4387958090d7112a880a7df8fb8a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java#L623) I find very ugly which is why I'm willing to create the breaking change

XiangRongLin: After thinking about it over the night, scrap most of what I said. Inside the app it will be displayed as a playlist and not something hugely different. So go with the playlist approach. 
A consideration is not using inheriting from `PlaylistInfoItem` but just simply adding the `mixType` field to it as `playlistType` with the type `NORMAL` and the ones already listed.
I can't think of any (dis-)advantes right now for that

opusforlife2:
Yay! Youtube Music mixes! 💃

Stypox:
As TiA4f8R pointed out I need to change the videos being tested to non-copyrighted ones

opusforlife2:
Every Mix gives me this error:

## Exception
* __User Action:__ requested playlist
* __Request:__ Start loading: https://www.youtube.com/watch?v=GYeDTTb4jYI&playnext=1&list=RDCMUCZYTClx2T1of7BRZ86-8fow
* __Content Country:__ CH
* __Content Language:__ en-IN
* __App Language:__ en_IN
* __Service:__ YouTube
* __Version:__ 0.21.16
* __OS:__ Linux Android 10 - 29

```
java.lang.NullPointerException: Attempt to invoke interface method 'java.util.Iterator java.util.List.iterator()' on a null object reference
	at org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper.extractCookieValue(YoutubeParsingHelper.java:1090)
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeMixPlaylistExtractor.onFetchPage(YoutubeMixPlaylistExtractor.java:88)
	at org.schabi.newpipe.extractor.Extractor.fetchPage(Extractor.java:54)
	at org.schabi.newpipe.extractor.playlist.PlaylistInfo.getInfo(PlaylistInfo.java:65)
	at org.schabi.newpipe.util.ExtractorHelper.lambda$getPlaylistInfo$9(ExtractorHelper.java:171)
	at org.schabi.newpipe.util.ExtractorHelper$$ExternalSyntheticLambda11.call(Unknown Source:4)
	at io.reactivex.rxjava3.internal.operators.single.SingleFromCallable.subscribeActual(SingleFromCallable.java:43)
	at io.reactivex.rxjava3.core.Single.subscribe(Single.java:4813)
	at io.reactivex.rxjava3.internal.operators.single.SingleDoOnSuccess.subscribeActual(SingleDoOnSuccess.java:35)
	at io.reactivex.rxjava3.core.Single.subscribe(Single.java:4813)
	at io.reactivex.rxjava3.internal.operators.maybe.MaybeFromSingle.subscribeActual(MaybeFromSingle.java:41)
	at io.reactivex.rxjava3.core.Maybe.subscribe(Maybe.java:5330)
	at io.reactivex.rxjava3.internal.operators.maybe.MaybeConcatArray$ConcatMaybeObserver.drain(MaybeConcatArray.java:153)
	at io.reactivex.rxjava3.internal.operators.maybe.MaybeConcatArray$ConcatMaybeObserver.request(MaybeConcatArray.java:78)
	at io.reactivex.rxjava3.internal.operators.flowable.FlowableElementAtMaybe$ElementAtSubscriber.onSubscribe(FlowableElementAtMaybe.java:66)
	at io.reactivex.rxjava3.internal.operators.maybe.MaybeConcatArray.subscribeActual(MaybeConcatArray.java:42)
	at io.reactivex.rxjava3.core.Flowable.subscribe(Flowable.java:15868)
	at io.reactivex.rxjava3.internal.operators.flowable.FlowableElementAtMaybe.subscribeActual(FlowableElementAtMaybe.java:36)
	at io.reactivex.rxjava3.core.Maybe.subscribe(Maybe.java:5330)
	at io.reactivex.rxjava3.internal.operators.maybe.MaybeToSingle.subscribeActual(MaybeToSingle.java:46)
	at io.reactivex.rxjava3.core.Single.subscribe(Single.java:4813)
	at io.reactivex.rxjava3.internal.operators.single.SingleSubscribeOn$SubscribeOnObserver.run(SingleSubscribeOn.java:89)
	at io.reactivex.rxjava3.core.Scheduler$DisposeTask.run(Scheduler.java:644)
	at io.reactivex.rxjava3.internal.schedulers.ScheduledRunnable.run(ScheduledRunnable.java:65)
	at io.reactivex.rxjava3.internal.schedulers.ScheduledRunnable.call(ScheduledRunnable.java:56)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
	at java.lang.Thread.run(Thread.java:919)

```
</details>
<hr>

Did no one else encounter this? I didn't bother reporting because I thought you would have come across it instantly.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeMixPlaylistExtractorTest.java
```diff
@@ -381,5 +397,100 @@ void getPage() throws Exception {
...
             assertEquals(ListExtractor.ITEM_COUNT_INFINITE, extractor.getStreamCount());
         }
+
+        @Test
+        void getPlaylistType() throws ParsingException {
+            assertEquals(PlaylistInfo.PlaylistType.MIX_CHANNEL, extractor.getPlaylistType());
+        }
+    }
+
+    public static class GenreMix {
```
Stypox, line 407: I could not find another video which suggests me a genre mix, so I won't change these, also because they are autogenerated playlists, not actual possible copyrighted videos

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/stream/YoutubeStreamExtractorRelatedMixTest.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.Random;
+import java.util.stream.Collectors;
+
+import javax.annotation.Nullable;
+
+public class YoutubeStreamExtractorRelatedMixTest extends DefaultStreamExtractorTest {
+    private static final String RESOURCE_PATH = DownloaderFactory.RESOURCE_PATH + "services/youtube/extractor/stream/";
+    static final String ID = "UtF6Jej8yb4";
```
Stypox, line 35: Instead I found an NCS video with a stream mix and a yt music mix in the suggestions, I will use this one: https://www.youtube.com/watch?v=K4DyBUG242c

Stypox:
I changed the copyrighted video to https://www.youtube.com/watch?v=K4DyBUG242c as requested and, since that video also incuded a music mix, I found out music mixes were not being extracted. They were a still different type of recommended item, `compactPlaylistRenderer`, so I added support for that, too. I did so by changing the `MixPlaylistInfoItemExtractor` to `MixOrPlaylistInfoItemExtractor` and by parsing the uploader name and the stream count automatically (this let me remove the hardcoded "YouTube" uploader). I pushed a commit for that. @XiangRongLin please review again ;-)

opusforlife2:
@Stypox The latest APK is from the `delete-large-land-player` branch (just FYI). And Mixes still give the same error.

opusforlife2:
Progress! Tested Adele's [Hello](https://www.youtube.com/watch?v=YQHsXMglC9A). It's the only video where I got **_two_** mixes in related videos instead of just one. One is the regular "Mix - [video name]", which still doesn't work. For all videos that showed me a Mix in related videos, this was the title format.

But the other one does! I don't know if you'll get the same suggestion, but it's called "Hip Hop Essentials" and the uploader is shown as "YouTube Music". There's no "Mix" in the title.

Stypox:
@opusforlife2 yep, the one you found is a YouTube Music Mix and is among the ones I added support for in 098acf2ad8fda7985a97fa3bf6e49751c6a20a87

XiangRongLin:
I'll tkae a look at it on the weekend.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
```diff
@@ -271,33 +284,106 @@ public static boolean isYoutubeChannelMixId(@Nonnull final String playlistId) {
...
+            throw new ParsingException("Video id could not be determined from genre mix id: "
                     + playlistId);
 
-        } else if (isYoutubeMixId(playlistId)) { // normal mix, starts with "RD"
+        } else if (isYoutubeMixId(playlistId)) { // normal mix
+            if (playlistId.length() != 13) {
+                // Stream YouTube mixes are of the form RD{videoId}, but if videoId is not exactly
+                // 11 characters then it can't be a video id, hence we are dealing with a different
+                // type of mix (e.g. genre mixes handled above, of the form RDGMEM{garbage})
+                throw new ParsingException("Video id could not be determined from mix id: "
```
XiangRongLin, line 331: Do you have the 11 character number from that stackexchange post I shared? If yes one could link it here: https://webapps.stackexchange.com/questions/54443/format-for-id-of-youtube-video/101153#101153

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/stream/YoutubeStreamExtractorRelatedMixTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import java.util.Random;
+import java.util.stream.Collectors;
+
+import javax.annotation.Nullable;
+
+public class YoutubeStreamExtractorRelatedMixTest extends DefaultStreamExtractorTest {
```
XiangRongLin, line 33: > Note that the YoutubeStreamExtractorRelatedMixTest only works reliably if the downloader is MOCK, since YouTube does not always send the same mixes each time. Should I do something for this? I don't think there is any way to test reliably this part other than only using MOCK, since suggested items alongside videos are autogenerated by YouTube.

For that usecase I once added  `MockOnlyRule`, but it seems that it was deleted at some point, so yeah...

XiangRongLin (commented):
I just did a cursory review. Since the branch was force pushed I have no idea what exactly is new and old.
Codewise it looks good to me.

> Testing apk: app-debug.zip

What exactly can be tested and how? API compatibility?

Stypox:
> What exactly can be tested and how? API compatibility?

Well, you can open music videos on youtube and you can verify that mixes now appear as related items (which is what this PR is almost all about)

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
Stypox: Even though I read your post, the fact that a youtube video id is 11 characters long was known to me. Anyway, I will add that.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/stream/YoutubeStreamExtractorRelatedMixTest.java
Stypox: It was deleted in https://github.com/TeamNewPipe/NewPipeExtractor/pull/768 since Junit 5 doesn't support rules anymore and that rule was not even used. I found [this article](https://www.codeaffine.com/2016/04/06/replace-rules-in-junit5/) that explains how to implement ~~rules~~ extensions in junit 5, I'm gonna open a PR for it.

Stypox:
>  If yes one could link it here:

Done

@XiangRongLin have you understood what/how you should test?

XiangRongLin (approved):
> have you understood what/how you should test?

Yes. Then LGTM

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/stream/YoutubeStreamExtractorRelatedMixTest.java
Stypox: Added `@MockOnly` to `testRelatedItems`

Stypox:
@XiangRongLin rebased and added `@MockOnly` to `testRelatedItems`, please review again and then merge :-)

opusforlife2:
Woohoo!

Issue #793, opened 2022-02-08
YouTube extract total comment count


FireMasterK:
Requested at: https://github.com/TeamPiped/Piped/issues/800

TiA4f8R:
It should be also implemented for other services, if possible and available.

PR #794, opened 2022-02-08
[YouTube] Add support to extract total comment count

FireMasterK:

Related issue: #793

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfo.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfo.java
@@ -50,7 +50,7 @@ public final class CommentsInfo extends ListInfo<CommentsInfoItem> {
         commentsInfo.setRelatedItems(initialCommentsPage.getItems());
         try {
             commentsInfo.setCommentsCount(commentsExtractor.getCommentsCount());
-        } catch (Exception e) {
+        } catch (final Exception e) {
             commentsInfo.addError(e);
         }
         commentsInfo.setNextPage(initialCommentsPage.getNextPage());
@@ -92,8 +92,6 @@ public final class CommentsInfo extends ListInfo<CommentsInfoItem> {
     }
 
     /**
-     * @return <code>true</code> if the comments are disabled otherwise <code>false</code> (default)
-     * @apiNote Warning: This method is experimental and may get removed in a future release.
      * @return {@code true} if the comments are disabled otherwise {@code false} (default)
      * @see CommentsExtractor#isCommentsDisabled()
      */
@@ -102,8 +100,6 @@ public final class CommentsInfo extends ListInfo<CommentsInfoItem> {
     }
 
     /**
-     * @param commentsDisabled <code>true</code> if the comments are disabled otherwise <code>false</code>
-     * @apiNote Warning: This method is experimental and may get removed in a future release.
      * @param commentsDisabled {@code true} if the comments are disabled otherwise {@code false}
      */
     public void setCommentsDisabled(final boolean commentsDisabled) {
@@ -124,7 +120,7 @@ public final class CommentsInfo extends ListInfo<CommentsInfoItem> {
      *
      * @param commentsCount the commentsCount to set.
      */
-    public void setCommentsCount(int commentsCount) {
+    public void setCommentsCount(final int commentsCount) {
         this.commentsCount = commentsCount;
     }
 }
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfo.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfo.java
@@ -113,7 +113,7 @@ public final class CommentsInfo extends ListInfo<CommentsInfoItem> {
     /**
      * Returns the total number of comments.
      *
-     * @return totalComments
+     * @return commentsCount the total number of comments.
      */
     public int getCommentsCount() {
         return commentsCount;
@@ -122,7 +122,7 @@ public final class CommentsInfo extends ListInfo<CommentsInfoItem> {
     /**
      * Sets the total number of comments.
      *
-     * @param commentsCount
+     * @param commentsCount the commentsCount to set.
      */
     public void setCommentsCount(int commentsCount) {
         this.commentsCount = commentsCount;
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfo.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfo.java
@@ -48,6 +48,11 @@ public final class CommentsInfo extends ListInfo<CommentsInfoItem> {
                 ExtractorHelper.getItemsPageOrLogError(commentsInfo, commentsExtractor);
         commentsInfo.setCommentsDisabled(commentsExtractor.isCommentsDisabled());
         commentsInfo.setRelatedItems(initialCommentsPage.getItems());
+        try {
+            commentsInfo.setCommentsCount(commentsExtractor.getCommentsCount());
+        } catch (Exception e) {
+            commentsInfo.addError(e);
+        }
         commentsInfo.setNextPage(initialCommentsPage.getNextPage());
 
         return commentsInfo;
@@ -76,6 +81,7 @@ public final class CommentsInfo extends ListInfo<CommentsInfoItem> {
 
     private transient CommentsExtractor commentsExtractor;
     private boolean commentsDisabled = false;
+    private int commentsCount;
 
     public CommentsExtractor getCommentsExtractor() {
         return commentsExtractor;
@@ -86,6 +92,7 @@ public final class CommentsInfo extends ListInfo<CommentsInfoItem> {
     }
 
     /**
+     * @return <code>true</code> if the comments are disabled otherwise <code>false</code> (default)
      * @apiNote Warning: This method is experimental and may get removed in a future release.
      * @return {@code true} if the comments are disabled otherwise {@code false} (default)
      * @see CommentsExtractor#isCommentsDisabled()
@@ -95,10 +102,29 @@ public final class CommentsInfo extends ListInfo<CommentsInfoItem> {
     }
 
     /**
+     * @param commentsDisabled <code>true</code> if the comments are disabled otherwise <code>false</code>
      * @apiNote Warning: This method is experimental and may get removed in a future release.
      * @param commentsDisabled {@code true} if the comments are disabled otherwise {@code false}
      */
     public void setCommentsDisabled(final boolean commentsDisabled) {
         this.commentsDisabled = commentsDisabled;
+    }
+
+    /**
+     * Returns the total number of comments.
+     *
+     * @return totalComments
+     */
+    public int getCommentsCount() {
+        return commentsCount;
+    }
+
+    /**
+     * Sets the total number of comments.
+     *
+     * @param commentsCount
+     */
+    public void setCommentsCount(int commentsCount) {
+        this.commentsCount = commentsCount;
     }
 }
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfo.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfo.java
@@ -109,7 +109,7 @@ public final class CommentsInfo extends ListInfo<CommentsInfoItem> {
     /**
      * Returns the total number of comments.
      *
-     * @return commentsCount the total number of comments.
+     * @return the total number of comments
      */
     public int getCommentsCount() {
         return commentsCount;
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
@@ -33,12 +33,7 @@ public class YoutubeCommentsExtractor extends CommentsExtractor {
     /**
      * Whether comments are disabled on video.
      */
-    private boolean commentsDisabled = true;
-
-    /**
-     * The total number of comments on video.
-     */
-    private int commentsCount = (int) ITEM_COUNT_UNKNOWN;
+    private boolean commentsDisabled;
 
     /**
      * The second ajax <b>/next</b> response.
@@ -268,6 +263,10 @@ public class YoutubeCommentsExtractor extends CommentsExtractor {
         final String initialToken =
                 findInitialCommentsToken(getJsonPostResponse("next", body, localization));
 
+        if (initialToken == null) {
+            return;
+        }
+
         // @formatter:off
         final byte[] ajaxBody = JsonWriter.string(
                         prepareDesktopJsonBuilder(localization, getExtractorContentCountry())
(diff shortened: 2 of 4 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
```diff
@@ -257,4 +257,17 @@ public boolean isCommentsDisabled() throws ExtractionException {
...
+
+    @Override
+    public int getCommentsCount() throws ExtractionException {
+        final JsonObject commentsHeaderRenderer = ajaxJson
+                .getArray("onResponseReceivedEndpoints").getObject(0)
+                .getObject("reloadContinuationItemsCommand")
+                .getArray("continuationItems").getObject(0)
+                .getObject("commentsHeaderRenderer");
+
+        final String text = getTextFromObject(commentsHeaderRenderer.getObject("countText"));
```
litetex, line 269: I think we don't need a variable here.

* ``commentsHeaderRenderer.getObject("countText")`` can be done in the above code block
* ``getTextFromObject`` can be done inline in the below code

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
```diff
@@ -257,4 +257,17 @@ public boolean isCommentsDisabled() throws ExtractionException {
...
+
+        return Integer.parseInt(Utils.removeNonDigitCharacters(text));
```
litetex, line 271: This is not error resistant.
Should be wrapped into a ``ExtractionException``

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
```diff
@@ -45,6 +45,7 @@
      */
     @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
     private Optional<Boolean> optCommentsDisabled = Optional.empty();
+    private JsonObject ajaxJson;
```
litetex, line 48: No documentation.
I'm also not sure if it's good to keep this in the memory, because it might be rather big.

I think it would be better when you instantly fetch the comment count when the extractor is called (like the optCommentsDisabled).
This also hides the parameter inside ``getNextPage`` below.

FireMasterK: > I'm also not sure if it's good to keep this in the memory, because it might be rather big.

It's a JSON object, and these classes shouldn't be reused, right?

> I think it would be better when you instantly fetch the comment count when the extractor is called (like the optCommentsDisabled).

Wouldn't that result in 2 API calls? That's exactly what I want to avoid.

litetex: > It's a JSON object

Not sure how that changes the situation

> and these classes shouldn't be reused

The class itself is not reused but the comments extractor is e.g. hold in memory by https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfo.java#L73


> Wouldn't that result in 2 API calls? That's exactly what I want to avoid.

No, ``optCommentsDisabled`` also does exactly one API call.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
```diff
@@ -257,4 +260,21 @@ public boolean isCommentsDisabled() throws ExtractionException {
...
+    public int getCommentsCount() throws ExtractionException {
+        final JsonObject countText = ajaxJson
+                .getArray("onResponseReceivedEndpoints").getObject(0)
+                .getObject("reloadContinuationItemsCommand")
+                .getArray("continuationItems").getObject(0)
+                .getObject("commentsHeaderRenderer")
+                .getObject("countText");
+
+        try {
+            final String text = getTextFromObject(countText);
```
litetex, line 274: (optional) It's better now but I think it could still be done inline

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
FireMasterK: > No, optCommentsDisabled also does exactly one API call.

That's because it's extracted in findInitialCommentsToken(), which can be called without any stateless parameters, unlike here where we require a continuation token.

I don't what's the problem since we do exactly this in the `YoutubeStreamExtractor` class: https://github.com/TeamNewPipe/NewPipeExtractor/blob/65129e65a81563c81cd8dbcb6c64e7570bd2910c/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java#L93-L101

I had a look at implementing your suggestion, but I couldn't find a simple way, which wouldn't create 2 requests.

litetex: How I would have implemented it:

<details><summary>CommentsExtractor.java</summary>

```diff
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsExtractor.java
@@ -6,9 +6,12 @@ import org.schabi.newpipe.extractor.exceptions.ExtractionException;
 import org.schabi.newpipe.extractor.exceptions.ParsingException;
 import org.schabi.newpipe.extractor.linkhandler.ListLinkHandler;
 
+import java.util.OptionalInt;
+
 import javax.annotation.Nonnull;
 
 public abstract class CommentsExtractor extends ListExtractor<CommentsInfoItem> {
+    public static int NO_COMMENT_COUNT = -1;
 
     public CommentsExtractor(final StreamingService service, final ListLinkHandler uiHandler) {
         super(service, uiHandler);
@@ -22,6 +25,14 @@ public abstract class CommentsExtractor extends ListExtractor<CommentsInfoItem>
         return false;
     }
 
+    /**
+     *
+     * @return total number of comments or {@link #NO_COMMENT_COUNT} if no count was extracted.
+     */
+    public int getCommentsCount() {
+        return NO_COMMENT_COUNT;
+    }
+
     @Nonnull
     @Override
     public String getName() throws ParsingException {
```

</details>

<details><summary>YoutubeCommentsExtractor.java</summary>

```diff
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
@@ -1,7 +1,6 @@
 package org.schabi.newpipe.extractor.services.youtube.extractors;
 
-import static org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper.getJsonPostResponse;
-import static org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper.prepareDesktopJsonBuilder;
+import static org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper.*;
 import static org.schabi.newpipe.extractor.utils.Utils.UTF_8;
 import static org.schabi.newpipe.extractor.utils.Utils.isNullOrEmpty;
 
@@ -9,6 +8,7 @@ import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
+import java.util.OptionalInt;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -25,6 +25,7 @@ import org.schabi.newpipe.extractor.exceptions.ParsingException;
 import org.schabi.newpipe.extractor.linkhandler.ListLinkHandler;
 import org.schabi.newpipe.extractor.localization.Localization;
 import org.schabi.newpipe.extractor.utils.JsonUtils;
+import org.schabi.newpipe.extractor.utils.Utils;
 
 import com.grack.nanojson.JsonArray;
 import com.grack.nanojson.JsonObject;
@@ -46,6 +47,15 @@ public class YoutubeCommentsExtractor extends CommentsExtractor {
     @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
     private Optional<Boolean> optCommentsDisabled = Optional.empty();
 
+    /**
+     * Caching mechanism and holder of the commentsCount value.
+     * <br/>
+     * Initial value = empty -> unknown if comment a comment count is present or not<br/>
+     * Some method calls {@link #determineCommentsCount(JsonObject)}
+     * -> value is set<br/>
+    * If the method is called again the cached value will be used.
+     */
+    @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
+    private OptionalInt optCommentsCount = OptionalInt.empty();
+
     public YoutubeCommentsExtractor(
             final StreamingService service,
             final ListLinkHandler uiHandler) {
@@ -177,12 +182,36 @@ public class YoutubeCommentsExtractor extends CommentsExtractor {
 
         final JsonObject ajaxJson = getJsonPostResponse("next", body, localization);
 
-        final CommentsInfoItemsCollector collector = new CommentsInfoItemsCollector(
-                getServiceId());
+        final CommentsInfoItemsCollector collector =
+                new CommentsInfoItemsCollector(getServiceId());
+        determineCommentsCount(ajaxJson);
         collectCommentsFrom(collector, ajaxJson);
         return new InfoItemsPage<>(collector, getNextPage(ajaxJson));
     }
 
+    private void determineCommentsCount(final JsonObject ajaxJson) {
+        // Check if commentscount was already determined
+        if (optCommentsCount.isPresent()) {
+            return;
+        }
+
+        final JsonObject countText = ajaxJson
+                .getArray("onResponseReceivedEndpoints").getObject(0)
+                .getObject("reloadContinuationItemsCommand")
+                .getArray("continuationItems").getObject(0)
+                .getObject("commentsHeaderRenderer")
+                .getObject("countText");
+
+        try {
+            this.optCommentsCount = OptionalInt.of(
+                    Integer.parseInt(Utils.removeNonDigitCharacters(getTextFromObject(countText)))
+            );
+        } catch (final Exception e) {
+            // If no count could be determined use the default
+            this.optCommentsCount = OptionalInt.of(NO_COMMENT_COUNT);
+        }
+    }
+
     private void collectCommentsFrom(final CommentsInfoItemsCollector collector,
                                      @Nonnull final JsonObject ajaxJson) throws ParsingException {
 
@@ -257,4 +286,11 @@ public class YoutubeCommentsExtractor extends CommentsExtractor {
 
         return optCommentsDisabled.get();
     }
+
+    @Override
+    public int getCommentsCount() {
+        assertPageFetched();
+
+        return optCommentsCount.orElse(NO_COMMENT_COUNT);
+    }
 }
```

</details>

prateekmedia:
Any progress?

FireMasterK:
I'm currently a bit busy, I may be able to resume work on this the next or the week after.

FireMasterK:
Closing in favor of #936

AudricV:
This PR is about implementing replies count of a comment, while yours is about a different feature, the comments count of a stream!

Reopening.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
Stypox: I second @litetex's suggestion, unless I'm missing something

FireMasterK: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Optional.html

> Optional is primarily intended for use as a method return type where there is a clear need to represent "no result," and where using null is likely to cause errors. A variable whose type is Optional should never itself be null; it should always point to an Optional instance.

I want to remove the optional altogether, it's not meant to be used as a field at all!

FireMasterK: Yes, I think so too. But then, exception handling will get a little trickier, but that's okay.

FireMasterK: I've done something different: I've removed the `nextResponse` JSON object and replaced it with `initialToken` which is a String extracted on page fetch.

I really don't think we should be concerned about memory usage since I don't think anyone should be holding extractor objects for long periods. The `CommentInfo` class drops them right after it's used for example, so any performance differences would be _very very_ minimal.

We store response objects anyways in `YoutubePlaylistExtractor`, `YoutubeChannelExtractor`, `YoutubeStreamExtractor`, etc. So, I don't think this is an issue.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
```diff
@@ -24,26 +14,37 @@
...
-     * Initial value = empty -> unknown if comments are disabled or not<br/>
-     * Some method calls {@link #findInitialCommentsToken()}
-     * -> value is set<br/>
-     * If the method or another one that is depending on disabled comments
-     * is now called again, the method execution can avoid unnecessary calls
+     * The second ajax <b>/next</b> response.
      */
-    @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
-    private Optional<Boolean> optCommentsDisabled = Optional.empty();
+    private JsonObject ajaxJson;
```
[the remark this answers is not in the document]
FireMasterK: > I see an issue with this approach: if the client make a call to getCommentsCount() after loading some pages, it will use the comments count from latest next continuation response. YouTube provides only the comments section header for the first continuation.

I don't think this should be a problem since extractor objects shouldn't be reused? 

> This approach may drop however the ability to know extraction errors when trying to get comments count.

Yeah this is the issue with that approach, it would silently fail.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeCommentsExtractorTest.java
```diff
@@ -344,6 +345,12 @@ public void testGetCommentsReplyCount() throws IOException, ExtractionException
             assertNotEquals(UNKNOWN_REPLY_COUNT, firstComment.getReplyCount(), "Could not get the reply count of the first comment");
             assertGreater(300, firstComment.getReplyCount());
         }
+
+        @Test
+        public void testCommentsCount() throws IOException, ExtractionException {
+            extractor.getInitialPage(); // Needs to be called first
```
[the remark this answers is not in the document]
FireMasterK: This is no longer necessary in my new changes

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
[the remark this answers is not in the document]
FireMasterK: Removed it  :+1:

I added it before since other extractors store it like that in a field (likely for caching reasons), but seeing there's no consistency in how it's done, I removed it.

extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfo.java
```diff
@@ -86,7 +92,6 @@ public void setCommentsExtractor(final CommentsExtractor commentsExtractor) {
     }
 
     /**
-     * @apiNote Warning: This method is experimental and may get removed in a future release.
```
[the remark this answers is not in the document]
Stypox: I don't think that was the reason. The `Optional` was in YouTube's specific implementation, not here. Anyway, I think it does not make sense to keep this as experimental. It has not been changed [since one and a half year](https://github.com/TeamNewPipe/NewPipeExtractor/commit/fdebf3c6cdf88fed8a372d7d7376acfa3a9f8c10) and will probably not be changed again any time soon. So this is as experimental as all other extractor features, which may change in case we need to change it, but generally things don't change.

Stypox (changes requested):
The structure is fine for me. @AudricV please merge this if you agree.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
```diff
@@ -254,24 +252,59 @@ private void collectCommentsFrom(final CommentsInfoItemsCollector collector,
...
+    public boolean isCommentsDisabled() {
+        return commentsDisabled;
+    }
+
+    @Override
+    public int getCommentsCount() throws ExtractionException {
+        assertPageFetched();
+
+        if (commentsDisabled) {
+            return -1;
```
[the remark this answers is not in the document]
FireMasterK: I disagree, since when comments are disabled, there's technically no count available. So, I feel -1 is the best value to represent this.

PR #663, opened 2021-06-20
Add a proper support of other delivery methods than progressive HTTP

TiA4f8R:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).
- [x] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API.

This is the work made by @wb9688 (thank you for your work, I wouldn't have been able to do it myself) in #367 but rebased and with some changes:

<h3>Changes in Stream classes</h3>

- Rename `url` (now deprecated) to `content` and add `isUrl` to indicate whether content contains the URL or the file itself (which is useful when we have to split DASH/HLS manifests);
- Add `id` to identify a specific stream (there could be multiple Streams with the same ID if the same stream is offered through multiple delivery methods; in YouTube's case the ID would be the itag);
- Add a `baseUrl` field, required for some players (like ExoPlayer) to play manifests.

<h3>Other changes:</h3>

- Add `DeliveryFormat` enum with `PROGRESSIVE_HTTP`, `DASH`, `HLS` and `TORRENT` values;
- Don't support SoundCloud HLS only-streams with the HLS delivery method, but with the current method which is progressive HTTP (so they are downloadable by NewPipe);
- Extract the downloadable file of SoundCloud tracks (on tracks on which download is available);
- Fix parsing of YouTube DASH manifests (not used now by the extractor);
- Generate DASH manifests for YouTube OTF streams and post live streams, by using the data returned in the first sequence;
- Format the `ItagItem` class;
- Update `DefaultStreamExtractorTest` test with the changes made in Stream classes;
- Update `SoundcloudStreamExtractor` test to test if streams returned by the extractor are progressive HTTP (test will fail if that's not the case);
- Return the HLS URL of PeerTube streams and their HLS variant, if available;
- Return PeerTube audio streams as audio streams (and not as video streams);
- Add `POST_LIVE_STREAM` and `POST_LIVE_AUDIO_STREAM` to StreamType enum;
- Return captions for YouTube age-restrcited videos (because we are now able to extract them, probably since the InnerTube migration for age-restricted contents).

### Fixed issues

Closes #461, fixes #648

PR #796, opened 2022-02-11
Add support for extracting channel subscriber count in StreamInfo

### LINKED ISSUE #795: [YouTube] Extract Subscriber Count in StreamInfo
I don't think this can be implemented in all services since I checked on Peetube, it doesn't seem to include it.

FireMasterK:

Closes #795

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
@@ -180,7 +180,7 @@ public class YoutubeStreamExtractor extends StreamExtractor {
             // TODO: this parses English formatted dates only, we need a better approach to parse
             //  the textual date
             final LocalDate localDate = LocalDate.parse(getTextFromObject(
-                    getVideoPrimaryInfoRenderer().getObject("dateText")),
+                            getVideoPrimaryInfoRenderer().getObject("dateText")),
                     DateTimeFormatter.ofPattern("dd MMM yyyy", Locale.ENGLISH));
             return DateTimeFormatter.ISO_LOCAL_DATE.format(localDate);
         } catch (final Exception ignored) {
@@ -440,14 +440,13 @@ public class YoutubeStreamExtractor extends StreamExtractor {
     @Override
     public long getUploaderSubscriberCount() throws ParsingException {
         final JsonObject videoOwnerRenderer = JsonUtils.getObject(videoSecondaryInfoRenderer, "owner.videoOwnerRenderer");
-        if (videoOwnerRenderer.has("subscriberCountText")) {
-            try {
-                return Utils.mixedNumberWordToLong(getTextFromObject(videoOwnerRenderer.getObject("subscriberCountText")));
-            } catch (final NumberFormatException e) {
-                throw new ParsingException("Could not get uploader subscriber count", e);
-            }
-        } else {
+        if (!videoOwnerRenderer.has("subscriberCountText")) {
             return UNKNOWN_SUBSCRIBER_COUNT;
+        }
+        try {
+            return Utils.mixedNumberWordToLong(getTextFromObject(videoOwnerRenderer.getObject("subscriberCountText")));
+        } catch (final NumberFormatException e) {
+            throw new ParsingException("Could not get uploader subscriber count", e);
         }
     }
 
@@ -680,9 +679,9 @@ public class YoutubeStreamExtractor extends StreamExtractor {
         final Localization localization = getExtractorLocalization();
         final ContentCountry contentCountry = getExtractorContentCountry();
         final byte[] body = JsonWriter.string(prepareDesktopJsonBuilder(
-                localization, contentCountry)
-                .value("videoId", videoId)
-                .done())
+                        localization, contentCountry)
+                        .value("videoId", videoId)
+                        .done())
                 .getBytes(UTF_8);
 
         // Put the sts string if we already know it so we don't have to fetch again the player
@@ -731,8 +730,8 @@ public class YoutubeStreamExtractor extends StreamExtractor {
 
         if (ageRestricted) {
             final byte[] ageRestrictedBody = JsonWriter.string(prepareDesktopEmbedVideoJsonBuilder(
-                    localization, contentCountry, videoId)
-                    .done())
+                            localization, contentCountry, videoId)
+                            .done())
                     .getBytes(UTF_8);
             nextResponse = getJsonPostResponse("next", ageRestrictedBody, localization);
         } else {
@@ -814,7 +813,7 @@ public class YoutubeStreamExtractor extends StreamExtractor {
                                               final String videoId)
             throws IOException, ExtractionException {
         final byte[] mobileBody = JsonWriter.string(prepareAndroidMobileJsonBuilder(
-                localization, contentCountry)
+                        localization, contentCountry)
                         .value("videoId", videoId)
                         .done())
                 .getBytes(UTF_8);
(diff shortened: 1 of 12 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -437,6 +438,20 @@ public String getUploaderAvatarUrl() throws ParsingException {
...
+    public long getUploaderSubscriberCount() throws ParsingException {
+        final JsonObject videoOwnerRenderer = JsonUtils.getObject(videoSecondaryInfoRenderer, "owner.videoOwnerRenderer");
+        if (videoOwnerRenderer.has("subscriberCountText")) {
+            try {
+                return Utils.mixedNumberWordToLong(getTextFromObject(videoOwnerRenderer.getObject("subscriberCountText")));
+            } catch (final NumberFormatException e) {
+                throw new ParsingException("Could not get subscriber count", e);
+            }
+        } else {
+            return ITEM_COUNT_UNKNOWN;
```
TobiGr, line 451: this should be either -1 or have its own constant.
I'd suggest to add an own constant, because that's the way we went with in #663

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -437,6 +438,20 @@ public String getUploaderAvatarUrl() throws ParsingException {
...
     }
 
+    @Override
+    public long getUploaderSubscriberCount() throws ParsingException {
+        final JsonObject videoOwnerRenderer = JsonUtils.getObject(videoSecondaryInfoRenderer, "owner.videoOwnerRenderer");
+        if (videoOwnerRenderer.has("subscriberCountText")) {
+            try {
+                return Utils.mixedNumberWordToLong(getTextFromObject(videoOwnerRenderer.getObject("subscriberCountText")));
+            } catch (final NumberFormatException e) {
+                throw new ParsingException("Could not get subscriber count", e);
```
TobiGr, line 448:
```suggestion
                throw new ParsingException("Could not get uploader subscriber count", e);
```

TobiGr (commented):
Thanks!

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
TobiGr: Yes.

FireMasterK: Done, I haven't added any documentation for the variable though.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -437,6 +439,20 @@ public String getUploaderAvatarUrl() throws ParsingException {
...
+        final JsonObject videoOwnerRenderer = JsonUtils.getObject(videoSecondaryInfoRenderer, "owner.videoOwnerRenderer");
+        if (videoOwnerRenderer.has("subscriberCountText")) {
+            try {
+                return Utils.mixedNumberWordToLong(getTextFromObject(videoOwnerRenderer.getObject("subscriberCountText")));
+            } catch (final NumberFormatException e) {
+                throw new ParsingException("Could not get uploader subscriber count", e);
+            }
+        } else {
+            return UNKNOWN_SUBSCRIBER_COUNT;
+        }
```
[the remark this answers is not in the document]
FireMasterK: Please feel free to push changes to my branch the next time :)

extractor/src/test/java/org/schabi/newpipe/extractor/services/DefaultStreamExtractorTest.java
```diff
@@ -45,6 +45,7 @@
     public abstract String expectedUploaderName();
     public abstract String expectedUploaderUrl();
     public boolean expectedUploaderVerified() { return false; }
+    public long expectedUploaderSubscriberCountAtLeast() { return -1; }
```
Stypox, line 48: This test isn't the best: if somebody forgets to override it, the related test will always pass, so it wouldn't be testing anything. The other methods with default values provide only one possible value, so the related test will pass only if the value is the default one or otherwise fail (hence notifying the user that he should override the default function). In this case this doesn't happen. Here's my suggestion (1):
```suggestion
    public long expectedUploaderSubscriberCountAtLeast() { return UNKNOWN_SUBSCRIBER_COUNT; }
```

extractor/src/test/java/org/schabi/newpipe/extractor/services/DefaultStreamExtractorTest.java
```diff
@@ -105,6 +106,12 @@ public void testUploaderVerified() throws Exception {
         assertEquals(expectedUploaderVerified(), extractor().isUploaderVerified());
     }
 
+    @Test
+    @Override
+    public void testSubscriberCount() throws Exception {
+        assertGreaterOrEqual(expectedUploaderSubscriberCountAtLeast(), extractor().getUploaderSubscriberCount());
```
Stypox, line 112: ... continues (2)
```suggestion
        if (expectedUploaderSubscriverCountAtLeast() == UNKNOWN_SUBSCRIBER_COUNT) {
            assertEquals(UNKNOWN_SUBSCRIBER_COUNT, extractor().getUploaderSubscriberCount());
        } else {
            assertGreaterOrEqual(expectedUploaderSubscriberCountAtLeast(), extractor().getUploaderSubscriberCount());
        }
```
Also, this way if `expectedUploaderSubscriverCountAtLeast()` returns `UNKNOWN_SUBSCRIBER_COUNT`, i.e. its default value, we will make sure the extractor actually returns an unknown subscriber count.

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamExtractor.java
```diff
@@ -202,6 +203,17 @@ public boolean isUploaderVerified() throws ParsingException {
...
 
+    /**
+     * The subscriber count of the uploader.
+     * If the subscriber count is not implemented, or is unavailable, return <code>-1</code>.
+     *
+     * @return the subscriber count of the uploader or -1 if not available
+     * @throws ParsingException
+     */
+    public long getUploaderSubscriberCount() throws ParsingException {
+        return -1;
```
Stypox, line 214: (also update docs)
```suggestion
        return UNKNOWN_SUBSCRIBER_COUNT;
```

Stypox (changes requested):

extractor/src/test/java/org/schabi/newpipe/extractor/services/DefaultStreamExtractorTest.java
FireMasterK: > This test isn't the best: if somebody forgets to override it, the related test will always pass, so it wouldn't be testing anything.

That's intended, I didn't want to modify all existing tests by looking for values for each video, hence why.

Stypox: Oh, would it be a problem for you to solve the tests? The CI already tells you all of the expected values: https://github.com/TeamNewPipe/NewPipeExtractor/runs/5284419766

FireMasterK: Done!

Stypox (approved):
Looks good to me, thank you :-)

litetex (approved):
LGTM

PR #797, opened 2022-02-11
Fix NullPointerException in YouTubeStreamExtractor.getStreamSegments()

TobiGr:

Fixes https://github.com/TeamNewPipe/NewPipe/issues/7792

For more info see https://github.com/TeamNewPipe/NewPipe/issues/7792#issuecomment-1030900188

APK: [app-debug.zip](https://github.com/TeamNewPipe/NewPipeExtractor/files/8049631/app-debug.zip)

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
@@ -1272,8 +1272,7 @@ public class YoutubeStreamExtractor extends StreamExtractor {
                 final String panelIdentifier = panels.getObject(i)
                         .getObject("engagementPanelSectionListRenderer")
                         .getString("panelIdentifier");
-                if (panelIdentifier != null && panelIdentifier.equals(
-                        "engagement-panel-macro-markers-description-chapters")) {
+                if ("engagement-panel-macro-markers-description-chapters".equals(panelIdentifier)) {
                     segmentsArray = panels.getObject(i)
                             .getObject("engagementPanelSectionListRenderer").getObject("content")
                             .getObject("macroMarkersListRenderer").getArray("contents");
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
@@ -1272,7 +1272,7 @@ public class YoutubeStreamExtractor extends StreamExtractor {
                 final String panelIdentifier = panels.getObject(i)
                         .getObject("engagementPanelSectionListRenderer")
                         .getString("panelIdentifier");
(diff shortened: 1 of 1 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -1272,7 +1272,7 @@ public Locale getLanguageInfo() {
                 final String panelIdentifier = panels.getObject(i)
                         .getObject("engagementPanelSectionListRenderer")
                         .getString("panelIdentifier");
-                if (panelIdentifier.equals(
+                if (panelIdentifier != null && panelIdentifier.equals(
                         "engagement-panel-macro-markers-description-chapters")) {
```
litetex, line 1276:
```suggestion
                if ("engagement-panel-macro-markers-description-chapters".equals(panelIdentifier)) {
```

litetex (commented):
This would fix the problem with less code

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
TobiGr: That's way better.

litetex (approved):

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -1272,8 +1272,7 @@ public Locale getLanguageInfo() {
                 final String panelIdentifier = panels.getObject(i)
                         .getObject("engagementPanelSectionListRenderer")
                         .getString("panelIdentifier");
-                if (panelIdentifier.equals(
-                        "engagement-panel-macro-markers-description-chapters")) {
+                if ("engagement-panel-macro-markers-description-chapters".equals(panelIdentifier)) {
```
mhmdanas, line 1275: It's not obvious at all why the comparison is flipped if you don't know `panelIdentifier` can be `null`. How about adding a comment?

TobiGr: Something like that?
``` java
                // panelIdentifier might be null if the panel has something to do with ads
                // See https://github.com/TeamNewPipe/NewPipe/issues/7792#issuecomment-1030900188
                @Nullable
                if ("engagement-panel-macro-markers-description-chapters".equals(panelIdentifier)) {
```

TobiGr: because that var can be null

mhmdanas: Ah, I saw the comment before you edited it; the `@Nullable` was directly before the `if` statement then. xD

PR #799, opened 2022-02-13
Refactored YoutubeCommentsExtractor

litetex:

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
@@ -73,7 +73,8 @@ public class YoutubeCommentsExtractor extends CommentsExtractor {
 
     /**
      * Finds the initial comments token and initializes commentsDisabled.
-     * Also set
+     * <br/>
+     * Also sets {@link #optCommentsDisabled}.
      *
      * @return the continuation token or null if none was found
      */
@@ -85,14 +86,12 @@ public class YoutubeCommentsExtractor extends CommentsExtractor {
                 // Only use JsonObjects
                 .filter(JsonObject.class::isInstance)
                 .map(JsonObject.class::cast)
-                // Only process JsonObjects that have a itemSectionRenderer
-                .filter(jObj -> jObj.has("itemSectionRenderer"))
                 // Check if the comment-section is present
                 .filter(jObj -> {
                     try {
                         return "comments-section".equals(
                                 JsonUtils.getString(jObj, "itemSectionRenderer.targetId"));
-                    } catch (final ParsingException ex) {
+                    } catch (final ParsingException ignored) {
                         return false;
                     }
                 })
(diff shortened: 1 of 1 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
```diff
@@ -74,45 +73,46 @@ public YoutubeCommentsExtractor(
...
-
-        optCommentsDisabled = Optional.of(false);
+        final String token = JsonUtils.getArray(nextResponse,
+                "contents.twoColumnWatchNextResults.results.results.contents")
+                .stream()
+                // Only use JsonObjects
+                .filter(JsonObject.class::isInstance)
+                .map(JsonObject.class::cast)
+                // Only process JsonObjects that have a itemSectionRenderer
+                .filter(jObj -> jObj.has("itemSectionRenderer"))
```
Stypox, line 89: This is probably unneeded since the `.filter` just below already returns false if the `JsonObject` has no `"itemSectionRenderer"`

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
```diff
@@ -74,45 +73,46 @@ public YoutubeCommentsExtractor(
...
+                // Check if the comment-section is present
+                .filter(jObj -> {
+                    try {
+                        return "comments-section".equals(
+                                JsonUtils.getString(jObj, "itemSectionRenderer.targetId"));
+                    } catch (final ParsingException ex) {
```
Stypox, line 95: Otherwise Android Studio gives warning for unused variable
```suggestion
                    } catch (final ParsingException ignored) {
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
```diff
@@ -224,24 +242,23 @@ private void collectCommentsFrom(final CommentsInfoItemsCollector collector,
...
-                final CommentsInfoItemExtractor extractor = new YoutubeCommentsInfoItemExtractor(
-                        (JsonObject) c, getUrl(), getTimeAgoParser());
-                collector.commit(extractor);
-            }
-        }
+        final String url = getUrl();
+        comments.stream()
+                .filter(JsonObject.class::isInstance)
+                .map(JsonObject.class::cast)
+                .map(jObj -> new YoutubeCommentsInfoItemExtractor(jObj, url, getTimeAgoParser()))
```
Stypox, line 249:
```suggestion
        comments.stream()
                .filter(JsonObject.class::isInstance)
                .map(JsonObject.class::cast)
                .map(jObj -> new YoutubeCommentsInfoItemExtractor(jObj, getUrl(), getTimeAgoParser()))
```

Stypox (changes requested):
Thanks!

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
litetex: FYI: My Android studio shows no warning here:
grafik

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
litetex: That's not possible since it throws an exception which can't be handled in the lambda (because it's not a RuntimeException).
grafik

PR #802, opened 2022-02-18
[YouTube] Add a comment clarifying why the panelIdentfier can be null

TobiGr:

Follow up of #797


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -1269,6 +1269,8 @@ public Locale getLanguageInfo() {
 
             // Search for correct panel containing the data
             for (int i = 0; i < panels.size(); i++) {
+                // panelIdentifier might be null if the panel has something to do with ads
+                // See https://github.com/TeamNewPipe/NewPipe/issues/7792#issuecomment-1030900188
```
mhmdanas, line 1273: Maybe move this to before the `if` statement? I'm not really sure though, so you can choose whichever you want.

mhmdanas (approved):

PR #808, opened 2022-03-08
Turn FUNCTION_NAME_PATTERN and N_PARAM_PATTERN in YoutubeThrottlingDecrypter not final and create setter method

fabiano-araujo:

I created a setter method for the variables nParamPattern and functionNamePattern that were previously declared as "final". So that when the decryption code is changed it is not necessary to update the library in the app to be able to work normally, being able to use some kind of api to store the decryption code and get it working again at the same time, without the user having to update the app .


Stypox:
Mmmh, I don't think this is a good idea:
- Those are internal constants and the library user is not supposed to know how they are used
- Those variables are not the only things that YouTube could change, so why those ones and not others?
- If we decide to make them public we'd need to make them consistent across versions (otherwise people's code using them would break), but those variables and variable names are supposed to be changed or removed

opusforlife2:
@fabiano-araujo You haven't replied to Stypox's comments above.

PR #811, opened 2022-03-12
[YouTube] Fix the extraction of series playlists and don't return the view count as the stream count for learning playlists

TiA4f8R:

These changes need testing on every playlist type available, because I removed some old code related to the extraction of a playlist type, which seems to be not used anymore by YouTube.

This PR fixes the extraction of YouTube series playlists and the extraction of the stream count for learning playlists: instead of returning the view count of the playlist as the stream count, `ITEM_COUNT_UNKNOWN` is now returned for these playlists.

Because some of the methods in the `PlaylistExtractor` abstract class are mostly overriden with the same return value in inherited classes, I implemented them in the base class instead and they are now only overridden where it's relevant.

I also improved and refactored a bit the classes which extends the `PlaylistExtractor` one (use of the Java 8 Streaming API and the `StandardCharsets` class, use of the `final` keyword where possible, etc.).

This PR fixes TeamNewPipe/NewPipe#5004 and fixes TeamNewPipe/NewPipe#7609.

Debug APK to test the changes: [app-debug.zip](https://github.com/TeamNewPipe/NewPipeExtractor/files/8238149/app-debug.zip)


TiA4f8R:
@SovietScout @yashpalgoyal1304 Could you test this APK and see if it fixes the issues? Thank you in advance.

Stypox (changes requested):
Looks almost good to me, thank you!

<sub>As always the real fix is in between a hundred other unrelated changed lines :-/</sub>

TiA4f8R:
> Please remove the "fixes" for suggestions and mixes

Done.

Stypox (approved):
Code looks good and tests succeed, lgtm

PR #815, opened 2022-03-15
Removed hardcoded soundcloud HARDCODED_CLIENT_ID

### LINKED ISSUE #781: Separate CI + test for hardcoded SoundCloud client ID
As clarified here: https://github.com/TeamNewPipe/NewPipeExtractor/pull/777#issuecomment-1021422300, it is possible, but needs time and effort.

Benefit: CI will stop failing on unrelated PRs because of an outdated client ID.

litetex:
It never works (long enough) so let's simply remove it...

More why we should remove it points:
1. It always fails after a month (because new id)
2. We always have to update it and it breaks our tests
3. No one benefits from this not even the users - maybe for ~2 weeks (since we usually have a 2 week pre-release testing period) but then it's useless as it's not up-to-date anymore...

Fixes #781


Stypox (approved):
I would also remove the hardcoded client id

PR #816, opened 2022-03-16
Add `@MockOnly` Junit 5 extension

Stypox:

The `@MockOnly` rule was introduced by #517 and then removed when migrating to Junit 5 in #768 because it was not used. Now I would need it again for #788, so this PR reintroduces it as a Junit 5 extension.

In #517 the rule also had a `reason()`, to make sure devs give a reason for making a test mock only, but I don't seem to be able to have Android Studio / Gradle print out ignore reasons for me, so I skipped adding that.

This PR also makes `DownloaderFactory`'s methods static, so that there is no need to create a new instance each time and also because I needed to extract `getDownloaderType` from the `DownloaderFactory.getDownloader` method.

@XiangRongLin


XiangRongLin (approved):
I haven't tested it, but looks good.

PR #822, opened 2022-03-26
More refactors

Stypox:

While doing the checkstyle PR I found some more things that could be improved. This PR tries to use Java 8 streams in the bandcamp extractor (which should, among other things, improve resiliency), and fixes some other warnings here and there as reported by Android Studio. Check the commits for more info.

Btw if you leave comments/reviews don't expect me to answer until a few days/weeks ;-)

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampSearchExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampSearchExtractor.java
@@ -50,40 +50,31 @@ public class BandcampSearchExtractor extends SearchExtractor {
 
     public InfoItemsPage<InfoItem> getPage(final Page page)
             throws IOException, ExtractionException {
-        final String html = getDownloader().get(page.getUrl()).responseBody();
-
         final MultiInfoItemsCollector collector = new MultiInfoItemsCollector(getServiceId());
+        final Document d = Jsoup.parse(getDownloader().get(page.getUrl()).responseBody());
 
-
-        final Document d = Jsoup.parse(html);
-
-        final Elements searchResultsElements = d.getElementsByClass("searchresult");
-
-        for (final Element searchResult : searchResultsElements) {
-
-            final String type = searchResult.getElementsByClass("result-info").first()
-                    .getElementsByClass("itemtype").first().text();
+        for (final Element searchResult : d.getElementsByClass("searchresult")) {
+            final String type = searchResult.getElementsByClass("result-info").stream()
+                    .flatMap(element -> element.getElementsByClass("itemtype").stream())
+                    .map(Element::text)
+                    .findFirst()
+                    .orElse("");
 
             switch (type) {
                 default:
-                    continue;
                 case "FAN":
                     // don't display fan results
                     break;
-
                 case "ARTIST":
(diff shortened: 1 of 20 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampSearchExtractor.java
```diff
@@ -50,40 +50,31 @@ public boolean isCorrectedSearch() {
...
+        for (final Element searchResult : d.getElementsByClass("searchresult")) {
+            final String type = searchResult.getElementsByClass("result-info").stream()
+                    .flatMap(element -> element.getElementsByClass("itemtype").stream())
+                    .map(Element::text)
+                    .findFirst()
+                    .orElse("");
 
             switch (type) {
                 default:
-                    continue;
```
litetex, line 69: 1. shouldn't be the ``default`` at the end? (Code-Style)
2. this will cause the ``default`` to behave in the same ways as  ``case "FAN"`` which is wrong

extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampSearchExtractor.java
```diff
@@ -50,40 +50,31 @@ public boolean isCorrectedSearch() {
...
-        for (final Element searchResult : searchResultsElements) {
-
-            final String type = searchResult.getElementsByClass("result-info").first()
-                    .getElementsByClass("itemtype").first().text();
+        for (final Element searchResult : d.getElementsByClass("searchresult")) {
+            final String type = searchResult.getElementsByClass("result-info").stream()
+                    .flatMap(element -> element.getElementsByClass("itemtype").stream())
+                    .map(Element::text)
+                    .findFirst()
+                    .orElse("");
```
litetex, line 61: Consider using ``Utils.EMPTY_STRING``

extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampStreamExtractor.java
```diff
@@ -128,11 +128,10 @@ public String getThumbnailUrl() throws ParsingException {
...
     public String getUploaderAvatarUrl() {
-        try {
-            return document.getElementsByClass("band-photo").first().attr("src");
-        } catch (final NullPointerException e) {
-            return "";
-        }
+        return document.getElementsByClass("band-photo").stream()
+                .map(element -> element.attr("src"))
+                .findFirst()
+                .orElse("");
```
litetex, line 134: Consider using ``Utils.EMPTY_STRING``

litetex (changes requested):
Thank you for the cleanup!

Minor changes:

extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampSearchExtractor.java
Stypox: I just want `default:` to do the same thing as `"FAN"`, that is, exiting from the `switch` (`continue;` and `break;` result in the same thing happening here). But yes, I will move the default at the end.

litetex (approved):
LGTM 🥳

PR #825, opened 2022-04-01
Improve empty string so that it is meaningful for clients

Stypox:
- [no need] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API.

This allows clients not to worry about empty strings: they can just show to the user the string being provided by the extractor and the user will see a nice "Nothing here but crickets" message as usual. Tell me what you think about this. Since this PR is so small I'd like to get it merged by next version.

I tested this in NewPipe and it worked without issues. Here is an apk: [app-debug.zip](https://github.com/TeamNewPipe/NewPipeExtractor/files/8395006/app-debug.zip)

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
@@ -19,7 +19,7 @@ public final class Utils {
     public static final String HTTP = "http://";
     public static final String HTTPS = "https://";
     public static final String UTF_8 = "UTF-8";
-    public static final String EMPTY_STRING = "";
+    public static final String EMPTY_STRING = "Nothing here but crickets";
     private static final Pattern M_PATTERN = Pattern.compile("(https?)?:\\/\\/m\\.");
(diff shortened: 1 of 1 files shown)


TobiGr:
I do mit like the idea, to be honest. We really need to check, where the constant is used.
If we agree to set a default string, we should make it customizae. Either via the `NewPipe(...)` constructor or via a setter

Stypox:
As I said on IRC, in my opinion this PR would simplify the use of the extractor for clients using it without diving too much into what can and cannot be empty

Stypox:
Ha ha ha I fooled you XD

This PR does not make any sense, why would the empty string be anything different than `""`? Which makes me question why we even use `EMPTY_STRING` instead of `""`...

On a side note, I would have expected this joke to have more drastic consequences for NewPipe (e.g. crashes or strange data in fields) but I couldn't notice anything, which made the joke less funny...

mhmdanas:
Haha, you caught me off guard!

PR #832, opened 2022-04-15
Fix YouTube throttling decrypter function parsing

Stypox:
Tests pass. Can somebody (@TiA4f8R) confirm that this really works?
See https://github.com/yt-dlp/yt-dlp/blob/master/yt_dlp/extractor/youtube.py#L2535
Here is an apk of NewPipe: [app-debug.zip](https://github.com/TeamNewPipe/NewPipeExtractor/files/8495561/app-debug.zip)

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingDecrypter.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingDecrypter.java
@@ -8,6 +8,7 @@ import org.schabi.newpipe.extractor.utils.StringUtils;
 import javax.annotation.Nonnull;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 /**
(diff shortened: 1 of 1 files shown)


Stypox:
Yes, I can confirm this works. Without this fix I can only download at 125kB/s, with this fix I get up to 3MB/s on my home wifi.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingDecrypter.java
```diff
@@ -97,21 +98,24 @@ public static String apply(final String url, final String videoId) throws Parsin
...
-        return functionName;
+
+        final String functionName = matcher.group(1);
+        if (matcher.groupCount() == 1) {
+            return functionName;
+        }
+
+        final int arrayNum = Integer.parseInt(matcher.group(2));
+        final Pattern arrayPattern = Pattern.compile(
+                "var " + Pattern.quote(functionName) + "\\s*=\\s*\\[(.+?)];");
```
pukkandan, line 115:
```suggestion
                "var " + Pattern.quote(functionName) + "\\s*=\\s*\\[(.+?)][,;]");
```

Otherwise player https://www.youtube.com/s/player/b7910ca8/player_ias.vflset/en_US/base.js breaks

PR #841, opened 2022-04-26
[YouTube] Use correct order of clients to get better streams first

TiA4f8R:

In https://github.com/TeamNewPipe/NewPipeExtractor/pull/780#discussion_r837863231, I didn't noticed when applying the requested change, that by using a `HashMap`, the order of `streamingData` JSON objects wanted will be not respected (because in a `Map`, values are not ordered). This PR fixes this unwanted change, by using an `ArrayList` instead of a `HashMap`.

Because we need to get at the same time the content playback nonce of a client and its corresponding `streamingData` JSON object and `List`s can contain only type of an element, I copied the `Pair` class of #810 and used it in my fix.

To test the fix: get streams of a non age-restricted YouTube video (either audio or video), and check if most of streaming URLs contains `&c=ANDROID`, instead of `&c=WEB`.


Stypox (approved):
Looks good to me, thanks :-)
I tested and I can confirm it works as expected.

PR #843, opened 2022-05-02
Remove Checkstyle suppressions file and fix Checkstyle issues

TiA4f8R:
In 24e8399 and 8c1041d, violations of the Checkstyle line length rule have been added (because the lines were not matching the lines in the Checkstyle suppressions file). These changes causes right now a compile error.

This PR fixes these errors by splitting or reducing the corresponding lines in the character limit per line.

It also removes the usage of Checkstyle suppressions file and allows now usage of `// CHECKSTYLE:OFF` and `// CHECKSTYLE:ON` comments to suppress remaining Checkstyle false errors when usage of `@SuppressWarnings("checkstyle:CheckstyleRule")` is not possible (like in #833).


checkstyle/suppressions.xml
```diff
@@ -1,15 +0,0 @@
...
-<suppressions>
-  <!-- Use @SuppressWarnings("...") if it is possible, only use this file if it is not -->
-
-  <suppress checks="LineLength"
-    files="BandcampExtractorHelper.java"
-    lines="54"/>
-
-  <suppress checks="LineLength"
-    files="ItagItem.java"
-    lines="19"/>
```
Stypox, line 14: Keep this file, just remove the lines I highlighted

extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampExtractorHelper.java
```diff
@@ -54,7 +54,8 @@ public static String getStreamUrlFromIds(final long bandId,
 
     /**
      * Fetch artist details from mobile endpoint.
-     * <a href=https://notabug.org/fynngodau/bandcampDirect/wiki/rewindBandcamp+%E2%80%93+Fetching+artist+details>
+     * <a href="https://notabug.org/fynngodau/bandcampDirect/wiki/
+     * rewindBandcamp+%E2%80%93+Fetching+artist+details">
```
Stypox, line 58: Does it really still work?!?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/ItagItem.java
```diff
@@ -18,7 +18,7 @@
 public class ItagItem {
     /**
      * List can be found here
-     * https://github.com/ytdl-org/youtube-dl/blob/9fc5eafb8e384453a49f7cfe73147be491f0b19d/youtube_dl/extractor/youtube.py#L1071
+     * https://github.com/ytdl-org/youtube-dl/blob/9fc5eaf/youtube_dl/extractor/youtube.py#L1071
```
Stypox, line 21: Nice catch ;-)

Stypox (changes requested):
Also, if you need to readd it in `delivery-methods-v2`, why bother removing it?

extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampExtractorHelper.java
TiA4f8R: In IDEA, yes, because of the `"` characters I added. In other IDEs, I don't know.

checkstyle/suppressions.xml
TiA4f8R: Why? Because of your comment above?

TiA4f8R:
> Also, if you need to readd it in delivery-methods-v2, why bother removing it?

I can use the `// CHECKSTYLE:OFF` and `// CHECKSTYLE:ON` so it can be safely removed (and it will allow us to not have to update the suppressions file directly each time we made changes on code or documentation with false positives).

Stypox (approved):
Ok then ;-)

PR #846, opened 2022-05-09
Remove unused methods

litetex:

After working on https://github.com/TeamNewPipe/NewPipe/pull/7458 I noticed that these methods (and corresponding tests) are no longer required and can be removed.

This PR can be merged once https://github.com/TeamNewPipe/NewPipe/pull/7458 is done

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/NewPipe.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/NewPipe.java
@@ -25,9 +25,10 @@ import org.schabi.newpipe.extractor.exceptions.ExtractionException;
 import org.schabi.newpipe.extractor.localization.ContentCountry;
 import org.schabi.newpipe.extractor.localization.Localization;
 
+import java.util.List;
+
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import java.util.List;
 
 /**
  * Provides access to streaming services supported by NewPipe.
@@ -90,14 +91,6 @@ public final class NewPipe {
             }
         }
         throw new ExtractionException("No service can handle the url = \"" + url + "\"");
-    }
-
-    public static int getIdOfService(final String serviceName) {
-        try {
-            return getService(serviceName).getServiceId();
-        } catch (final ExtractionException ignored) {
-            return -1;
-        }
     }
 
     public static String getNameOfService(final int id) {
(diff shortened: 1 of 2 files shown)


Stypox (approved):

AudricV:
@litetex You should have checked with more caution because they are still used by NewPipe on the `dev` branch!

```
> Task :app:compileDebugKotlin
e: [...]/app/src/main/java/org/schabi/newpipe/error/ErrorInfo.kt: (68, 45): Unresolved reference: getNameOfService
e: [...]/app/src/main/java/org/schabi/newpipe/error/ErrorInfo.kt: (76, 45): Unresolved reference: getNameOfService
e: [...]/app/src/main/java/org/schabi/newpipe/error/ErrorInfo.kt: (98, 57): Unresolved reference: getNameOfService
e: [...]/app/src/main/java/org/schabi/newpipe/error/ErrorPanelHelper.kt: (109, 29): Unresolved reference: getNameOfService

> Task :app:compileDebugJavaWithJavac

[...]

[...]/app/src/main/java/org/schabi/newpipe/util/ServiceHelper.java:142: error: cannot find symbol
        final int serviceId = NewPipe.getIdOfService(serviceName);
                                     ^
  symbol:   method getIdOfService(String)
  location: class NewPipe
```

This PR should be reverted in my opinion.

litetex:
@AudricV 
Thank you for noticing, looks like I overlooked the Kotlin code.

A fix is provided with https://github.com/TeamNewPipe/NewPipe/pull/8531

TobiGr:
I am a little late on this. Can you please clarify why you removed the methods?
The extractor is not only used by NewPipe, but other applications, too. Removing methods without a prior warning / deprecation is not good practice.

litetex:
> Can you please clarify why you removed the methods?

1. I thought they were just used inside https://github.com/TeamNewPipe/NewPipe/pull/7458 but apparently I was wrong they were also used on some other places
2. These methods are just shorthand methods for ``getService`` with an additional exception-check because somehow ``getService`` throws an checked Exception instead of simply returning ``null`` which would be a lot better.
3. They are never used anywhere inside the extractor.
5. grafik
6. The tests are completely pointless
   a. The methods are in in my opinion way to simple to even have tests for
   b. If you write tests don't just write it for one testcase...

> Removing methods without a prior warning / deprecation is not good practice.

I think people can handle removal of 2 barely used methods. Especially when the complete extractor architecture was/will be reworked with #810 and #862.

Also there was no [prior warning when the behavior of returning null was changed to throwing a exception](https://github.com/TeamNewPipe/NewPipeExtractor/pull/19/files#diff-fda29b004e321b18a1036445762ebcd7ec3bcd5ba575b55b0d4cba480f080393).

> as you already started to move one method into the app.

I moved the method into the app, because a) it's used there and b) the new method doesn't have a use the unnecessary ``ExtractionException``.

PR #848, opened 2022-05-17
[YouTube] Fix extracting YouTube Music client version and key in the EU

TiA4f8R:

Google returns now the consent page of YouTube when accessing YouTube Music in EU, which can be avoided by adding the `ucbcb` parameter to the URL with the value `1` (`?ucbcb=1`/`&ucbcb=1`), like on YouTube.

This extraction of the website is used as a last resort to get the client version and the API key of YouTube Music (which should not happen because the hardcoded client version and API key are always valid, and the JavaScript service worker is fetch first in order to try to get them), but is broken in the EU because of these changes. This PR fixes it by addding the parameter above with its value.


Stypox (approved):
I can confirm it works. Thank you!

PR #850, opened 2022-05-19
Add test for Localization.getCountryCode and Localization.getLanguageCode

lacinoire:

My PR is not changing the API, so I did not perform manual tests against NewPipe.

Hey 😊
I want to contribute the following test:

Test that a `java.lang.IllegalArgumentException` is thrown when `getTimeAgoParser` is called with the parameter `localization = new org.schabi.newpipe.extractor.localization.Localization("W4JN-$nw<}7EGpwmm(EQ", "ndBdj-qEHp!#I]LDWP=,")`.
This tests the methods [`Localization.getCountryCode`](https://github.com/TeamNewPipe/NewPipeExtractor/blob/23fa31a1ec572cfb45f0f2fe88cc3844b151aa73/extractor/src/main/java/org/schabi/newpipe/extractor/localization/Localization.java#L71) and [`Localization.getLanguageCode`](https://github.com/TeamNewPipe/NewPipeExtractor/blob/23fa31a1ec572cfb45f0f2fe88cc3844b151aa73/extractor/src/main/java/org/schabi/newpipe/extractor/localization/Localization.java#L66).
This test is based on the test [`testAllServicesHaveDifferentId`](https://github.com/TeamNewPipe/NewPipeExtractor/blob/23fa31a1ec572cfb45f0f2fe88cc3844b151aa73/extractor/src/test/java/org/schabi/newpipe/extractor/NewPipeTest.java#L19).

Curious to hear what you think!

(I wrote this test as part of a research study at TU Delft. [Find out more](https://github.com/lacinoire/lacinoire/blob/main/README.md))

### DIFF
--- a/extractor/src/test/java/org/schabi/newpipe/extractor/NewPipeTest.java
+++ b/extractor/src/test/java/org/schabi/newpipe/extractor/NewPipeTest.java
@@ -1,6 +1,8 @@
 package org.schabi.newpipe.extractor;
 
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
+import org.schabi.newpipe.extractor.localization.Localization;
 
 import java.util.HashSet;
 
@@ -24,6 +26,21 @@ public class NewPipeTest {
                             + " (current service > " + streamingService.getServiceInfo().getName() + ")";
 
             assertTrue(servicesId.add(streamingService.getServiceId()), errorMsg);
+        }
+    }
+
+    @Test
+    public void testGetCountryCodeAndGetLanguageCode() {
+        try {
+            HashSet<Integer> servicesId = new HashSet<>();
+            for (StreamingService streamingService : NewPipe.getServices()) {
+                Localization Localization1 = new Localization("W4JN-$nw<}7EGpwmm(EQ", "ndBdj-qEHp!#I]LDWP=,");
+                String errorMsg = ((("There are services with the same id = " + streamingService.getServiceId()) + " (current service > ") + streamingService.getServiceInfo().getName()) + ")";
+                streamingService.getTimeAgoParser(new Localization("W4JN-$nw<}7EGpwmm(EQ", "ndBdj-qEHp!#I]LDWP=,"));
+            }
+            Assertions.fail("testAllServicesHaveDifferentId_mg38 should have thrown IllegalArgumentException");
+        } catch (IllegalArgumentException expected) {
+            Assertions.assertEquals("Localization is not supported (\"Localization[W4JN-$nw<}7EGpwmm(EQ-ndBdj-qEHp!#I]LDWP=,]\")", expected.getMessage());
         }
     }
 


extractor/src/test/java/org/schabi/newpipe/extractor/NewPipeTest.java
```diff
@@ -27,6 +29,21 @@ public void testAllServicesHaveDifferentId() throws Exception {
         }
     }
 
+    @Test
+    public void testGetCountryCodeAndGetLanguageCode() {
+        try {
+            HashSet<Integer> servicesId = new HashSet<>();
+            for (StreamingService streamingService : NewPipe.getServices()) {
```
litetex, line 36: I'm not sure why all services are tested.
The implementation is always identical:
https://github.com/lacinoire/NewPipeExtractor/blob/23fa31a1ec572cfb45f0f2fe88cc3844b151aa73/extractor/src/main/java/org/schabi/newpipe/extractor/StreamingService.java#L373-L402

extractor/src/test/java/org/schabi/newpipe/extractor/NewPipeTest.java
```diff
@@ -27,6 +29,21 @@ public void testAllServicesHaveDifferentId() throws Exception {
...
+                Localization Localization1 = new Localization("W4JN-$nw<}7EGpwmm(EQ", "ndBdj-qEHp!#I]LDWP=,");
```
litetex, line 37: This is unused...

Hint in general:
**Always** review your PR yourself before creating/giving it to someone for a review. Maybe also wait some minutes so your brain resets itself a bit.
Helps the later reviewers immensely and you don't embarrass yourself with trivial errors 😉

extractor/src/test/java/org/schabi/newpipe/extractor/NewPipeTest.java
```diff
@@ -27,6 +29,21 @@ public void testAllServicesHaveDifferentId() throws Exception {
...
+                String errorMsg = ((("There are services with the same id = " + streamingService.getServiceId()) + " (current service > ") + streamingService.getServiceInfo().getName()) + ")";
+                streamingService.getTimeAgoParser(new Localization("W4JN-$nw<}7EGpwmm(EQ", "ndBdj-qEHp!#I]LDWP=,"));
+            }
+            Assertions.fail("testAllServicesHaveDifferentId_mg38 should have thrown IllegalArgumentException");
+        } catch (IllegalArgumentException expected) {
+            Assertions.assertEquals("Localization is not supported (\"Localization[W4JN-$nw<}7EGpwmm(EQ-ndBdj-qEHp!#I]LDWP=,]\")", expected.getMessage());
+        }
+    }
```
litetex, line 45: Also keep in mind that this test will always be executed and that it has to be maintained by someone

extractor/src/test/java/org/schabi/newpipe/extractor/NewPipeTest.java
```diff
@@ -27,6 +29,21 @@ public void testAllServicesHaveDifferentId() throws Exception {
         }
     }
 
+    @Test
+    public void testGetCountryCodeAndGetLanguageCode() {
+        try {
+            HashSet<Integer> servicesId = new HashSet<>();
+            for (StreamingService streamingService : NewPipe.getServices()) {
+                Localization Localization1 = new Localization("W4JN-$nw<}7EGpwmm(EQ", "ndBdj-qEHp!#I]LDWP=,");
+                String errorMsg = ((("There are services with the same id = " + streamingService.getServiceId()) + " (current service > ") + streamingService.getServiceInfo().getName()) + ")";
```
litetex, line 38: These variable could be final (if used)

litetex (changes requested):
Thank you for the PR.

I think I have to disappoint you here a bit as I don't think we need a test here in the first place
I don't see a benefit of the test in the current form:
  * The method itself isn't quite good written IMHO as it could be a static one and should probably be somewhere else
  * When we do tests we usually want to test as many cases as possible - in particular corner cases - however you just add a single testcase
  * I think the method isn't complex enough for it to require a test in the first place. It's just 1 or 2 calls to ``TimeAgoPatternsManager.getTimeAgoParserFor`` and an Exception otherwise.

I think it would be better if  you checkout other issues and see if you can work on fixing these.
...Oh and maybe ask if you could work on them before you start 😉 

However here have some recommendations for your future coding:

lacinoire:
Hey @litetex !

Thanks a lot for the detailed feedback on the test 🙏🏼😊
You're raising a lot of good points to improve. Especially because of your pointer of testing as many (corner) cases as possible, I think this would need major rework (not just fixing the code quality issues). → I will close the PR 🙂

PR #859, opened 2022-06-16
Fix extraction of some properties in ItagItems for YouTube livestreams and post-live streams and remove completely SoundCloud HLS workaround

AudricV:
- [x] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API. -> See TeamNewPipe/NewPipe#8153

This PR fixes an issue, when adding `ItagItem`'s properties on livestreams and post livestreams, where the `fps`, `audioSampleRate` and `audioChannels` values were not set for these stream types (and so only set for video streams, thanks to a `else` introduced by @Stypox in c8a77da2abd419b9ca2b581722a8a19a36010610).

This caused exceptions when creating DASH manifests of audio streams (because a valid number of audio channels is required in the `AudioChannelConfiguration` element of DASH manifests). To prevent it again, if this value is unknown, a fallback to `2` has been added (because most YouTube audio streams have 2 channels).

This PR also removes completely the SoundCloud HLS workaround which is being removed progressively by SoundCloud for MP3 streams too. Unfortunately, it means for UMG tracks (which have now a faster loading time because no HLS manifest is now fetched) that no progressive stream will be available for download in NewPipe.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
@@ -1358,13 +1358,20 @@ public class YoutubeStreamExtractor extends StreamExtractor {
 
         if (streamType == StreamType.LIVE_STREAM || streamType == StreamType.POST_LIVE_STREAM) {
             itagItem.setTargetDurationSec(formatData.getInt("targetDurationSec"));
-        } else if (itagType == ItagItem.ItagType.VIDEO
-                || itagType == ItagItem.ItagType.VIDEO_ONLY) {
+        }
+
+        if (itagType == ItagItem.ItagType.VIDEO || itagType == ItagItem.ItagType.VIDEO_ONLY) {
             itagItem.setFps(formatData.getInt("fps"));
         } else if (itagType == ItagItem.ItagType.AUDIO) {
             // YouTube return the audio sample rate as a string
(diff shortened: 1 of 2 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -1358,13 +1358,20 @@ private ItagInfo buildAndAddItagInfoToList(
 
         if (streamType == StreamType.LIVE_STREAM || streamType == StreamType.POST_LIVE_STREAM) {
             itagItem.setTargetDurationSec(formatData.getInt("targetDurationSec"));
-        } else if (itagType == ItagItem.ItagType.VIDEO
-                || itagType == ItagItem.ItagType.VIDEO_ONLY) {
+        }
+
+        if (itagType == ItagItem.ItagType.VIDEO || itagType == ItagItem.ItagType.VIDEO_ONLY) {
```
Stypox, line 1363: Ooops, I didn't notice the `if` above was referring to `streamType` instead of `itagItem`, that's why I accidentally added an `else`

Stypox (approved):
Thank you! Simple fix, and sorry for the mistake in https://github.com/TeamNewPipe/NewPipeExtractor/pull/810/commits/b3c620f0d8b19ddc532fe90dec7b813aea2caa29#diff-e07bd9a3dfb657e0ac68c2847533825c0c25d9cc8486c9a5a0dab1d413d54d42R1361

PR #862, opened 2022-06-18
Delivery methods without design flaws

### LINKED ISSUE #858: Improve handling of HLS/DASH/Progressive HTTP streaming data
<details><summary>Followup issue of <a href="https://github.com/TeamNewPipe/NewPipeExtractor/pull/810#pullrequestreview-980869669">#810</a></summary>

> Currently ``Stream`` is dependent on service specific stuff (like YT's itag). This is a big problem because sooner or later this class will blow up. It's already confusing enough and this PR makes it worse by adding stuff for the delivery/transmission-method (currently the Stream class can only contain progressive HTTP streaming data).
> 
> It's also currently hard to say which service uses which type of streaming and it's not possible to use different delivery methods for the same stream (not sure if this is possible/useful in the first place...).
> 
> Also when looking at the PR for NewPipe: We can't just create a specific playbackresolver/mediasource foreach stream type or service, that will be unmaintainable in the long run.
> 
> My ideas how to resolve these problems:
> #### What should be fixed anyway
> 1. isVideoOnly inside VideoStream should be removed. It's either a ``VideoStream`` or a ``AudioStream`` or a ``VideoAudioStream`` which contains the data of both.
> 2. Move the YT specific stuff out of the stream classes. Example resolution could be a ``YouTubeVideoStream`` which extends ``VideoStream``. However this may be solved (see below) in an even better way.
> 3. I'm not sure how SubtitleStream is supposed to fit into the below constructs, there may be changes required. As far as I have seen subtitles can be delivered e.g. via DASH or in other ways.
> 
> #### Idea 1
> Move everything that is delivery specific from ``Stream``-like classes to a separate ``StreamDeliveryData``-like classes:
> Example proposal:
> NPStreamingArch2 drawio
> 
> 
> #### Idea 2
> If - due to currently unknown reasons to me - it's not possible to do Idea 1 we could - instead of nesting (as done in Idea 1) ``StreamDeliveryData`` into  ``Stream`` - simply create them side by side and derive the corresponding implementations from them.
> Example proposal:
> NPStreamingArch3 drawio

</details>

Overview how the delivery methods work:
DeliveryMethods drawio

litetex:

Fixes #858 

Status: Work in progress


Stypox:
Took a brief look: I like the structure. I usually tend to avoid having this many abstract things, but in this case there is no way to avoid it because of the complexity of the data. With this structure a user of the extractor would need to use `instanceof` to decide how to play a `Stream` he obtained, doesn't he?

Also, what about removing completely the concept of `getVideoStreams`, `getAudioStreams` and `getAudioVideoStreams` and just create a `getStreams` that returns all of them, and then the library user can just call `.filter(VideoStream.class::isInstance).map(VideoStream.class::cast)` to filter out what he needs (we could provide this as a utility method)? This way we would avoid duplicate code for extracting the same thing in 3 places, and we would stop having to store temporarily extracted streaming data in fields that use RAM needlessly. Anyway, this is a suggestion for a future PR.

litetex:
@Stypox 
Sound's interesting but as you said I think this is best in a followup.

I'm currently already struggling with NewPipe itself because nearly everything that is player related has to be reworked (which with the fact that I currently don't have much spare time gradually slows down the progress quite a bit).

Btw there will be further unfinished changes inbound here that are mostly download-related e.g. marking TorrentData as not downloadable or how the playback should be resolved in the best way, e.g. preferring the hlsMasterPlaylistUrl or something else for certain streams.

Stypox:
@litetex any news? :-)

litetex:
@Stypox 
Current state: I'm still working on it (but I'm currently busy doing other stuff so progress is kind of slow). The extractor side should be nearly finished (did a few redesigns regarding downloads) but the client side still needs major changes.
For now I rebased the branch and fixed conflicts from the last few months.

PR #813, opened 2022-03-14
Fixed tests (+ Youtube shorts in channels)

litetex:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).

The following tests are currently failing when I run them locally:
| Problematic test | Fixed? | Note |
| --- | --- | --- |
| ``MediaCCCRecentListExtractorTest.testStreamList()`` | ✔️ | Likely an external problem → opened https://github.com/voc/voctoweb/issues/609<br/>No feedback... Applying workaround: Ignore faulty items |
| ``PeertubeCommentsExtractorTest$Default.testGetComments()``<br/> and ``.testGetCommentsFromCommentsInfo()`` | ✔️ | The old comments/videos got deleted → used new ones |
| ``PeertubeStreamExtractorTest$WhatIsPeertube.testLikeCount()`` | ✔️ | PeerTube video somehow has less likes now |
| ``YoutubeMixPlaylistExtractorTest$Mix.getName()``<br/> ``YoutubeMixPlaylistExtractorTest$MixWithIndex.getName()`` | ✔️ | Video got deleted → used a new one |
| ``YoutubeMusicSearchExtractorTest$Suggestion.testSearchCorrected()``<br/> ``YoutubeMusicSearchExtractorTest$Suggestion.testSearchSuggestion()``<br/> ``YoutubeSearchExtractorTest$Suggestion.testSearchCorrected()`` | ✔️ |  Looks like YT currently has some problems in their backend: <br/> Constantly switching between "Did you mean" and "Showing results for ..." occurs:<br/> YTSearchSuggestionsBroken<br/> ⚠️ Disabled test for now - TODO: Create followup issue |
| ``YoutubeStreamExtractorControversialTest.testSubscriberCount()`` | ✔️ | Channel lost a ton of subscriptions (a few thousands) → decremented our expected number accordingly |
| ``YoutubeChannelExtractorTest$VSauce.testRelatedItems()`` | ✔️ | https://github.com/TeamNewPipe/NewPipe/issues/8034 - Workaround: Return duration = 0 when a YT shorts video is encountered/the duration could not be processed.<br/>⚠️ TODO: Create a followup issue |

PR #863, opened 2022-06-19
Add Content-Type header to all POST requests without an empty body

AudricV:

This PR adds the `Content-Type` header to all POST requests without an empty body made by the extractor. They are made only in YouTube and Bandcamp services, according to the usages of the `post` methods in the `Downloader` class.

Currently, in NewPipe and in tests, this header is partially guessed and added by the network library we use, OkHttp. But this is not the case of all network libraries (e.g. Cronet), where 400 responses would be encountered on YouTube because the HTTP specifications for POST requests are not required.

I also improved related code in the changed files, and fixed some issues on the fly.


Stypox:
Uhm, instead of all of this duplicate code, shouldn't we change the `Downloader.post()` methods so that they take a `String contentType` parameter? Then the implementers of Downloader will take care of setting `Content-Length` (which can be derived from `body.size`) and `Content-Type` (which is the provided value). Since, as you said, having those two headers is a requirement for `post` requests, imo it makes sense to change the `post()` signature, so that we will surely not forget to put the content type anymore as the compiler would complain, while the length is set always.

FireMasterK:
I'm not sure if `Content-Length`, should be set since for example, what if `gzip/brotli` is being used by the downloader?

AudricV:
> I'm not sure if `Content-Length`, should be set since for example, what if `gzip/brotli` is being used by the downloader?

Thank you for the catch! I removed the addition of this header, and kept only the `Content-Type` one.

Stypox (changes requested):
Just create a utility method in the downloader, something like this:
```
postWithContentType(url, headers, body, contentType, localization) {
    final Map<String, List<String>> actualHeaders = new HashMap<>(headers);
    actualHeaders.put("Content-Type", Collections.singletonList(contentType));
    post(url, actualHeaders, body, localization);
}
```
and maybe even this as a shorthand for json requests 
```
postWithContentTypeJson(url, headers, body, localization) {
    postWithContentType(url, headers, body, "application/json", localization);
}
```

The current strategy contains too much duplicated code.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMixPlaylistExtractor.java
```diff
@@ -211,14 +217,19 @@ public InfoItemsPage<StreamInfoItem> getPage(final Page page) throws IOException
         }
 
         final StreamInfoItemsCollector collector = new StreamInfoItemsCollector(getServiceId());
+
         final Map<String, List<String>> headers = new HashMap<>();
         addClientInfoHeaders(headers);
+        //noinspection ArraysAsListWithZeroOrOneArgument
+        headers.put("Cookie", Arrays.asList(page.getCookies().get(COOKIE_NAME)));
```
Stypox, line 224: Why not `Collections.singletonList()`? If there is a reason, please add a comment

AudricV: I will add a comment (that's to allow adding cookies by extractor clients such as the reCAPTCHA one).

AudricV:
Requested changes have been applied.

Note: I reverted update of `YoutubeLiveStreamExtractorTest` and removal of `YoutubeSearchExtractorTest.Suggestion`, as both tests will be fixed in #882. That's the reason why some tests will fail because of missing mocks.

AudricV:
@Stypox Could you please review again this pull request? Thank you in advance :)

Stypox (approved):
@AudricV I rebased and updated the YouTube mocks. The following tests fail, but I don't think they are related. You can merge this PR if you think it is still ok!

image

AudricV:
@Stypox Indeed, you are almost right:

- the first one may be related to handles introduction: the channel ID is not provided in the [`yt:channelId` global element of the RSS feed](https://github.com/TeamNewPipe/NewPipeExtractor/blob/c953e2341493a554fd464c2c93f60f7c54a6c49d/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeFeedExtractor.java#L60) anymore (nothing is provided, the channel ID can be extracted in other ways);
- the second one is due to #976;
- the third one is due to the to the fact that not all [the "learn more" button](https://github.com/TeamNewPipe/NewPipeExtractor/blob/c953e2341493a554fd464c2c93f60f7c54a6c49d/extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/search/YoutubeSearchExtractorTest.java#L274) is uppercase anymore, that's only the case for the first letter (fixed with 2ec296e6746884c13962311fe2dee7c62dc1c65b):
  YoutubeSearchExtractorTest.MetaInfoTest learn more text difference
- the fourth one is related to due to time extraction failures, which are not breaking changes, (when the duration of streams is unknown) it may be related to the fact that duration can be hidden by creators on search results as far I remember, and so inaccessible for us.
  
  I ran again the test to get mocks with only results which have a known duration (in 34d79bd2679d1b851dea67b4e05acb52b09bc8ed);
- the last one is country dependent, at least I could not reproduce this. Your new mocks doesn't contain the suggestion, so I created new ones (in 34d79bd2679d1b851dea67b4e05acb52b09bc8ed). See the conversation and the changes of #813 and #862 for related details.
  
  By the way, [the related code](https://github.com/TeamNewPipe/NewPipeExtractor/blob/c953e2341493a554fd464c2c93f60f7c54a6c49d/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSearchExtractor.java#L107-L115) should be not index array based I think.

opusforlife2:
> fixed some issues on the fly

Anything worth mentioning in the blog post?

PR #869, opened 2022-07-03
Use minimum required permissions for GitHub workflows

mhmdanas:
This reduces the attack surface if the workflows are ever compromised.


litetex (approved):
LGTM

TheAssassin (approved):

Stypox (approved):
Now that we have manually specified permissions for each workflow, I made the default permissions read-only. Let's see if this works correctly...
image

PR #877, opened 2022-07-27
Use Objects.requireNonNull().

Isira-Seneviratne:


litetex (approved):
LGTM

Stypox (approved):

Stypox:
This actually broke three tests, since now `NullPointerException` is (correctly) thrown instead of `IllegalArgumentException`. I'm fixing them in #882, but please take a look at tests when making PRs here.

Isira-Seneviratne:
> This actually broke three tests, since now `NullPointerException` is (correctly) thrown instead of `IllegalArgumentException`. I'm fixing them in #882, but please take a look at tests when making PRs here.

My bad 😅

PR #881, opened 2022-07-29
Use String.join() and Collectors.joining().

Isira-Seneviratne:

### DIFF
--- a/extractor/src/test/java/org/schabi/newpipe/extractor/utils/UtilsTest.java
+++ b/extractor/src/test/java/org/schabi/newpipe/extractor/utils/UtilsTest.java
@@ -3,9 +3,6 @@ package org.schabi.newpipe.extractor.utils;
 import org.junit.jupiter.api.Test;
 import org.schabi.newpipe.extractor.exceptions.ParsingException;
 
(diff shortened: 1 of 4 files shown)


extractor/src/test/java/org/schabi/newpipe/extractor/utils/UtilsTest.java
```diff
@@ -18,12 +18,6 @@ void testMixedNumberWordToLong() throws ParsingException {
         assertEquals(1.5e9, Utils.mixedNumberWordToLong("1,5B"), 0.0);
     }
 
-    @Test
-    void testJoin() {
-        assertEquals("some,random,stuff", Utils.join(",", Arrays.asList("some", "random", "stuff")));
-        assertEquals("some,random,not-null,stuff", Utils.nonEmptyAndNullJoin(",", new String[]{"some", "null", "random", "", "not-null", null, "stuff"}));
```
Stypox, line 24: The second test should not be removed

Stypox (changes requested):

Stypox (approved):
LGTM, tests pass

Issue #876, opened 2022-07-26
Specific playlist broken in EU


FireMasterK:
Similar to #820, but its a completely different error message

URL: https://www.youtube.com/playlist?list=RDCLAK5uy_mqS2vPtJPBzKaS_TeSc5mDZIaMRFo-_pc

## Exception
* __User Action:__ requested playlist
* __Request:__ Start loading: https://piped.kavin.rocks/playlist?list=RDCLAK5uy_mqS2vPtJPBzKaS_TeSc5mDZIaMRFo-_pc
* __Content Country:__ US
* __Content Language:__ en
* __App Language:__ en_US
* __Service:__ YouTube
* __Version:__ 0.23.1
* __OS:__ Linux Android 12 - 32
<details><summary><b>Crash log </b></summary><p>

```
org.schabi.newpipe.extractor.exceptions.ContentNotAvailableException: Got error: "The playlist does not exist."
	at org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper.defaultAlertsCheck(YoutubeParsingHelper.java:1403)
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubePlaylistExtractor.onFetchPage(YoutubePlaylistExtractor.java:76)
	at org.schabi.newpipe.extractor.Extractor.fetchPage(Extractor.java:60)
	at org.schabi.newpipe.extractor.playlist.PlaylistInfo.getInfo(PlaylistInfo.java:68)
	at org.schabi.newpipe.util.ExtractorHelper.lambda$getPlaylistInfo$9(ExtractorHelper.java:176)
	at org.schabi.newpipe.util.ExtractorHelper.$r8$lambda$RLOrwix0kCDfcc_WpFBc8DTJw0Q(Unknown Source:0)
	at org.schabi.newpipe.util.ExtractorHelper$$ExternalSyntheticLambda5.call(Unknown Source:4)
	at io.reactivex.rxjava3.internal.operators.single.SingleFromCallable.subscribeActual(SingleFromCallable.java:43)
	at io.reactivex.rxjava3.core.Single.subscribe(Single.java:4813)
	at io.reactivex.rxjava3.internal.operators.single.SingleDoOnSuccess.subscribeActual(SingleDoOnSuccess.java:35)
	at io.reactivex.rxjava3.core.Single.subscribe(Single.java:4813)
[...]

PR #871, opened 2022-07-08
Fixing up tests and the bugs they exposed

vludax:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).
- [x] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API.

Attempting to fix CI issues. Tested a local build with NewPipe. Let me know if you believe this should rather be done in a series of separate PRs, or there are better solutions :)

PR #882, opened 2022-07-30
Fix all tests

### LINKED ISSUE #820: Mix-Playlist extraction is broken in EU
Currently the Mix-Playlist extraction is broken (looks like YT changed something).
Test returns:
```
Could not get playlistData
org.schabi.newpipe.extractor.exceptions.ExtractionException: Could not get playlistData
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeMixPlaylistExtractor.onFetchPage(YoutubeMixPlaylistExtractor.java:100)
	at org.schabi.newpipe.extractor.Extractor.fetchPage(Extractor.java:60)
	at org.schabi.newpipe.extractor.services.youtube.YoutubeMixPlaylistExtractorTest$ChannelMix.setUp(YoutubeMixPlaylistExtractorTest.java:356)
	...
```

litetex:

This PR fixes all tests.

It also introduces an option to change the YouTube `consent` cookie value, to allow fixing mixes extraction in EU.

Supersedes #871, fixes #820.


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
```diff
@@ -73,8 +66,13 @@
 import java.util.Random;
 import java.util.regex.Pattern;
 
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
+import static org.schabi.newpipe.extractor.NewPipe.getDownloader;
```
Stypox: Why does your editor keep moving these around?

extractor/src/main/java/org/schabi/newpipe/extractor/exceptions/ConsentRequiredException.java
```diff
@@ -0,0 +1,12 @@
+package org.schabi.newpipe.extractor.exceptions;
+
+public class ConsentRequiredException extends ParsingException {
```
Stypox, line 3: Maybe it should be a `ContentNotAvailableException`: it is not the extractor's fault if the page can't be loaded, but rather YouTube not providing mixes without consent, so it is a content no available

Stypox (dismissed):
I rebased and added a commit for https://github.com/TeamNewPipe/NewPipeExtractor/pull/877#issuecomment-1204925022. Now all tests succeed on my PC. The code you changed also looks good; now we should handle `ContentRequiredException`s in NewPipe and allow users to grant consent.

AudricV:
The consent is only required for some playlists (#876) and mixes (#820) in EU.

That's why I'm strongly against allowing tracking in all YouTube requests we made. We should only do so where it is needed.

Remember that one of the main goals of the NewPipe projects is to enhance privacy.

Stypox:
Yes. This PR does not allow any new consent, as it is `false` by default.

opusforlife2:
What I see is that the consent variable has been wrapped in a setter/getter. I assume this is so that other NPE consumers can easily show a consent prompt, e.g. Piped?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
litetex: Looks like IntelliJ uses a different code style than Android studio (despite being based on the same IDE):
grafik


I will reconfigure my IntelliJ so that it uses Android Studio's settings.

Maybe we should also track these settings in Git so that they are consistent between developers.

litetex: For now I used Android Studios code format, since it's already used in the files.

litetex:
> What I see is that the consent variable has been wrapped in a setter/getter. I assume this is so that other NPE consumers can easily show a consent prompt, e.g. Piped?

1. Yes, in theory this should be possible now
2. The variable is static so it's technically no getter or setter
3. The static variable might change in the future for now it's primarily designed as a workaround so that the tests work again

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
Stypox: > Maybe we should also track these settings in Git so that they are consistent between developers.

Yeah, good idea. Could you do that, since you already looked into how it works? Thanks

opusforlife2:
> The variable is static so it's technically no getter or setter

I didn't get this bit.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
litetex: Not in this PR but I created an issue: https://github.com/TeamNewPipe/NewPipeExtractor/issues/892

litetex:
@Stypox I'm done with my changes you can review again if you have time.

litetex:
Update: So there have been a bunch of merge conflicts due to the recent hotfix.

These should be fixed now.

While fixing the conflicts I also fixed 2 additional problems:
1. YouTube Mixes no longer work for the specified (kurzgesagt) video
2. The suggestion/"Did you mean..." now longer works for the specified search term

AudricV (changes requested):

AudricV:
For https://github.com/TeamNewPipe/NewPipeExtractor/pull/882#issuecomment-1205007648, it seems you don't have to change something, as `addYoutubeHeaders()` is never called except by `YoutubeMixPlaylistExtractor` (this method should be called everywhere, **don't touch anything here** because in order to respect what I said, additional changes would be needed<sup>1</sup>), so the consent cookie is only sent on mixes.

Note that the current code as it is would not fix #876, as this consent is not applied on regular playlists.

<sup>1</sup>Here is my idea:

<details>

- Change `generateConsentCookie()` to  `generateConsentCookie(boolean)`, where the boolean parameter would be something like `useYesValue`. If `useYesValue` is true, return `YES+`; otherwise return `PENDING=` and the three random digits;
- Change `addCookieHeader(@Nonnull final Map<String, List<String>> headers)` to `addCookieHeader(@Nonnull final Map<String, List<String>> headers, final boolean consentValue)`;
- Use the boolean value provided to call `generateConsentCookie(boolean)`;
- Change `addYoutubeHeaders(@Nonnull final Map<String, List<String>> headers)` to `addYoutubeHeaders(@Nonnull final Map<String, List<String>> headers, final boolean consentValue)`;
- Call `addCookieHeader(headers, consentValue)` in this method;
- Use `addYoutubeHeaders(headers, YoutubeParsingHelper.isConsentAccepted())` in playlists and mixes (`addYoutubeHeaders` should be used anywhere, but that's not related to your PR at all);
- Note: singleton lists should not be used on `Cookie` header, to allow putting a reCAPTCHA one.
</details>

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
```diff
@@ -1391,8 +1399,9 @@ public static void addCookieHeader(@Nonnull final Map<String, List<String>> head
 
     @Nonnull
     public static String generateConsentCookie() {
-        final int statusCode = 100 + numberGenerator.nextInt(900);
-        return CONSENT_COOKIE + statusCode;
+        return "CONSENT="
+                + (isConsentAccepted() ? CONSENT_COOKIE_YES_VALUE : CONSENT_COOKIE_PENDING_VALUE)
+                + (100 + numberGenerator.nextInt(900));
```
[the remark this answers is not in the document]
litetex: Hmm pretty sure that it was required the last time when I run it.
Anyway seems to work now.
Will modify it accordingly :)

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
```diff
@@ -1692,4 +1691,12 @@ public static boolean isAndroidStreamingUrl(@Nonnull final String url) {
...
         return Parser.isMatch(C_IOS_PATTERN, url);
     }
+
+    public static void setConsentAccepted(final boolean accepted) {
+        consentAccepted = accepted;
+    }
+
+    public static boolean isConsentAccepted() {
+        return consentAccepted;
+    }
```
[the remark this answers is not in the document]
litetex: Added a link to the field.

litetex:
> For [#882 (comment)](https://github.com/TeamNewPipe/NewPipeExtractor/pull/882#issuecomment-1205007648), it seems you don't have to change something, as `addYoutubeHeaders()` is never called except by `YoutubeMixPlaylistExtractor` (this method should be called everywhere, **don't touch anything here** because in order to respect what I said, additional changes would be needed1), so the consent cookie is only sent on mixes.
> 
> Note that the current code as it is would not fix #876, as this consent is not applied on regular playlists.
> 
> 1Here is my idea:

Should I react to this?

If yes: Your solution doesn't make much sense to me, you can simply call ``addCookieHeader`` or not ;)
It also adds a lot of overhead for doing technically the same.

Also with foresighted to https://github.com/TeamNewPipe/NewPipeExtractor/pull/833: 
This would be an instance specific property and I think it can be handled in the settings of an instance with e.g. a toggle.

litetex:
Note: I also fixed https://github.com/TeamNewPipe/NewPipeExtractor/issues/902 by 
1. doing a check if the extracted JS is valid JavaScript and if not try the fallback (Regex)
2. Removing a ``}`` from the regex so that it matches correctly

AudricV (approved):

opusforlife2:
@litetex Could you update the PR description to mention all the changes this makes, such as https://github.com/TeamNewPipe/NewPipeExtractor/pull/882#issuecomment-1221596544?

PR #886, opened 2022-08-04
Make improvements to methods using toArray().

Isira-Seneviratne:

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/kiosk/KioskList.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/kiosk/KioskList.java
@@ -1,5 +1,7 @@
 package org.schabi.newpipe.extractor.kiosk;
 
+import static org.schabi.newpipe.extractor.utils.Utils.isNullOrEmpty;
+
 import org.schabi.newpipe.extractor.NewPipe;
 import org.schabi.newpipe.extractor.Page;
 import org.schabi.newpipe.extractor.StreamingService;
@@ -8,13 +10,12 @@ import org.schabi.newpipe.extractor.linkhandler.ListLinkHandlerFactory;
 import org.schabi.newpipe.extractor.localization.ContentCountry;
 import org.schabi.newpipe.extractor.localization.Localization;
 
-import javax.annotation.Nullable;
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 
-import static org.schabi.newpipe.extractor.utils.Utils.isNullOrEmpty;
+import javax.annotation.Nullable;
 
 public class KioskList {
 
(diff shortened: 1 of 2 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/kiosk/KioskList.java
```diff
@@ -80,8 +81,8 @@ public KioskExtractor getDefaultKioskExtractor(final Page nextPage,
         } else {
             if (!kioskList.isEmpty()) {
                 // if not set get any entry
-                final Object[] keySet = kioskList.keySet().toArray();
-                return getExtractorById(keySet[0].toString(), nextPage, localization);
+                final String[] keySet = kioskList.keySet().toArray(new String[0]);
+                return getExtractorById(keySet[0], nextPage, localization);
```
Stypox, line 85: I think this can be improved, in order to avoid uselessly converting the list to an array (my code was not checked, improve it if there are warnings)
```suggestion
                final String kioskId = kioskList.keySet().stream().findAny().get();
                return getExtractorById(kioskId, nextPage, localization);
```

extractor/src/main/java/org/schabi/newpipe/extractor/utils/Parser.java
```diff
@@ -124,9 +125,7 @@ public static String[] getLinksFromString(final String txt) throws ParsingExcept
                 links.add(txt.substring(ls.getBeginIndex(), ls.getEndIndex()));
             }
 
-            String[] linksarray = new String[links.size()];
-            linksarray = links.toArray(linksarray);
-            return linksarray;
```
Stypox, line 129: I think this method is just unused... Should we remove it or keep it in case it becomes handy later? In the second case, just change the function signature and avoid converting `links` to an array

Stypox (changes requested):

extractor/src/main/java/org/schabi/newpipe/extractor/kiosk/KioskList.java
```diff
@@ -78,10 +79,10 @@ public KioskExtractor getDefaultKioskExtractor(final Page nextPage,
         if (!isNullOrEmpty(defaultKiosk)) {
             return getExtractorById(defaultKiosk, nextPage, localization);
         } else {
-            if (!kioskList.isEmpty()) {
+            final String first = kioskList.keySet().stream().findFirst().orElse(null);
```
Stypox, line 82: Isn't `findAny` more appropriate, since you are accessing an unordered set?

Isira-Seneviratne: Sure.

Stypox (approved):
LGTM

PR #893, opened 2022-08-13
Use immutable sets in YoutubeParsingHelper.

Isira-Seneviratne:
----
* Bump the required Java version to 11.
* Use the immutable `Set` factory methods in `YoutubeParsingHelper`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,8 +5,8 @@ allprojects {
     compileJava.options.encoding = 'UTF-8'
     compileTestJava.options.encoding = 'UTF-8'
 
-    sourceCompatibility = 1.8
-    targetCompatibility = 1.8
+    sourceCompatibility = JavaVersion.VERSION_11
+    targetCompatibility = JavaVersion.VERSION_11
 
     version 'v0.22.1'
     group 'com.github.TeamNewPipe'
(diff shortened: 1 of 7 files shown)


build.gradle
```diff
@@ -5,8 +5,8 @@ allprojects {
     compileJava.options.encoding = 'UTF-8'
     compileTestJava.options.encoding = 'UTF-8'
 
-    sourceCompatibility = 1.8
-    targetCompatibility = 1.8
+    sourceCompatibility = 9
+    targetCompatibility = 9
```
AudricV, line 9: You should have precised that your changes require Java 9!

opusforlife2: ~~precised~~ specified

Isira-Seneviratne: Sorry about that. I updated the description.

Isira-Seneviratne: I didn't use Java 11 methods in this PR, but I guess Java 11 could be used instead.

FireMasterK (approved):
Looks mostly good, any particular reason why we're using `Locale.ROOT` and not `Locale.US` or `Locale.ENGLISH`? Since these are URLs, I don't think it matters much, does it?

Isira-Seneviratne:
> Looks mostly good, any particular reason why we're using `Locale.ROOT` and not `Locale.US` or `Locale.ENGLISH`? Since these are URLs, I don't think it matters much, does it? 

I used the suggestions provided in Android Studio.

PR #905, opened 2022-08-20
[YouTube] Add JavaScript lexer to parse completely throttling decryption function

### LINKED ISSUE #902: [YouTube] Throttling parameter decryption is broken, decrypt function is not again fully extracted
With player `1f7d5369`, the decryption of the throttling parameter fails because the function is not again fully extracted:

function_n_parameter_not_extracted_fully

Left: what is extracted by the extractor; right: the real function

The extractor still works, because this time an exception catch is properly made.

Theta-Dev:
- no API changes

fixes #902

I added a proper javascript lexer to extract the nsig function. The old naive implementation broke because of regexes containing curly braces.

I copied the tokenizer from Rhino because it is private and removed unneeded function. On top of that I had to add logic to determine whether a slash constitutes a regex delimiter or a division operator. I have adapted that from the [RESS](https://github.com/rusty-ecma/RESS/blob/master/src/lib.rs) library.

I also found an issue with the Rhino JS interpreter. Version 1.7.14 uses javax.lang.model.SourceVersion, which is not available on android. This causes the app to load indefinitely when opening a video. If you have any idea how to fix this without downgrading, please help me. I have no idea why this error did not occur before.
Checked with the debugger, the error occurs when executing the (correctly extracted) javascript.
Perhaps the new deobfuscation function uses a feature of the interpreter which causes it to load the class with the failing import.
https://github.com/mozilla/rhino/issues/1149

For now I downgraded Rhino to v1.7.13.


extractor/src/test/java/org/schabi/newpipe/extractor/utils/JavaScriptExtractorTest.java
```diff
@@ -0,0 +1,55 @@
...
+import java.nio.file.Files;
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.schabi.newpipe.FileUtils.resolveTestResource;
+
+public class JavaScriptExtractorTest
+{
+    @Test
+    public void testJsExtractor() throws ParsingException {
```
AudricV, line 20: Remove this unneeded `public` modifier. See [this Sonar rule](https://rules.sonarsource.com/java/RSPEC-5786) for more details.

extractor/src/test/java/org/schabi/newpipe/extractor/utils/JavaScriptExtractorTest.java
```diff
@@ -0,0 +1,55 @@
...
+        final String src = "Wka=function(d){var x = [/,,/,913,/(,)}/g,\"abcdef}\\\"\",];var y = 10/2/1;return x[1][y];}//some={}random-padding+;";
+        final String result = JavaScriptExtractor.matchToClosingBrace(src, "Wka=function");
+        assertEquals("(d){var x = [/,,/,913,/(,)}/g,\"abcdef}\\\"\",];var y = 10/2/1;return x[1][y];}", result);
+    }
+
+    @Test
+    public void testEverythingJs() throws ParsingException, IOException {
```
AudricV, line 27: Same as above.

AudricV (changes requested):
Thank you for taking care of this fix! Could you please rebase your PR (and so drop the Rhino version downgrade change that you made yourself)?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingDecrypter.java
```diff
@@ -117,12 +118,8 @@ private static String parseDecodeFunctionName(final String playerJsCode)
 
     @Nonnull
     private static String parseDecodeFunction(final String playerJsCode, final String functionName)
-            throws Parser.RegexException {
-        try {
-            return parseWithParenthesisMatching(playerJsCode, functionName);
-        } catch (final Exception e) {
-            return parseWithRegex(playerJsCode, functionName);
-        }
```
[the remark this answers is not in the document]
Theta-Dev: I'll keep the regex one. As simple as it is, it seems to be stable.

AudricV (approved):
Thank you again for all the efforts you put to fix this issue!

Your changes look good now to me.

Stypox (approved):
I think this approach looks good, and if we find any better way to do this in the future we can still undo this pretty easily. Code also looks good (I obviously didn't review the copied-over code). Thank you @Theta-Dev!

PR #907, opened 2022-08-22
Remove EMPTY_STRING.

### LINKED ISSUE #842: Why does EMPTY_STRING exist?
This is a followup to my joke PR #825. The reason I made that PR was to emphasize the fact that storing `""` in a constant variable and then using that variable throughout the code is imo completely pointless.
- If we are doing this for *optimization purposes* -> we are not optimizing anything. JVM already does this for us: https://www.tutorialspoint.com/how-and-where-does-string-literals-in-java-stored-in-the-memory.
- If we are doing this for *clarity* -> I personally find `""` clearer, more standard and more recognizable than `EMPTY_STRING` that sometimes could also appear as `Utils.EMPTY_STRING` and from a first look looks like a non-empty constant string. `""` can be considered a constant for which no name is needed, just like `0` (while it would be a good idea to give a name to the various `-1` scattered about in the extractor), and anyway the name `EMPTY_STRING` doesn't give any more information than just `""`.
- Importing `EMPTY_STRING` in all files increases the bytecode size (though this would really be just a minor problem)

I can only find downsides for `EMPTY_STRING` to exist and I would therefore remove it completely.

Isira-Seneviratne:
----
Closes #842.


extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampSearchExtractor.java
```diff
@@ -2,29 +2,30 @@
...
-import org.schabi.newpipe.extractor.utils.Utils;
 
-import javax.annotation.Nonnull;
 import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
 
+import javax.annotation.Nonnull;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
```
AudricV, line 27: Could you remove on the fly this annotation and replace it by the JavaX one? This file seems to contain the last usage of this annotation in the extractor code (line 36).

AudricV (commented):
Thank you!

AudricV (approved):
Thank you!

Stypox (approved):

PR #910, opened 2022-08-24
Add compat Locale.forLanguageTag() implementation.

Isira-Seneviratne:
----
Add a simple `Locale.forLanguageTag()` implementation, as the actual method is not available on API levels below 21.


mhmdanas:
> NewPipe now requires API level 21.

Does the extractor also require API level 21 too?

Isira-Seneviratne:
> > NewPipe now requires API level 21.
> 
> Does the extractor also require API level 21 too?

At the moment it requires API level 19 (due to using the `StandardCharsets` class).

mhmdanas:
Then this PR would raise that requirement to 21, right? Do we want that?

AudricV:
I don't think so, as some apps use the extractor on lower APIs, such as SkyTube.

The question is: does this method would be available with the desugar library based on Java 11?

Isira-Seneviratne:
> Then this PR would raise that requirement to 21, right? 

Yeah.

Do we want that?

> At this point, there aren't many devices running Android versions below 5 (API level 21), so it seems like a worthwhile change to me. What does anyone else think?

Isira-Seneviratne:
> I don't think so, as some apps use the extractor on lower APIs, such as SkyTube.

Oh, right. In that case, changes will need to be made as `StandardCharsets` requires API level 19 (though it seems like it'll be supported in desugar.)

> The question is: does this method would be available with the desugar library based on Java 11?

Probably? I checked using the Android Studio preview and it didn't seem to be available.

Stypox:
> Then this PR would raise that requirement to 21, right? Do we want that?

If I understand correctly, the purpose of this PR is actually to prevent raising the requirement to 21, while being able to use `LocaleCompat.forLanguageTag()`

extractor/src/main/java/org/schabi/newpipe/extractor/stream/SubtitlesStream.java
```diff
@@ -1,7 +1,10 @@
 package org.schabi.newpipe.extractor.stream;
 
+import static org.schabi.newpipe.extractor.utils.Utils.EMPTY_STRING;
+
```
Stypox, line 4: Unused import, or am I missing something?

Stypox (changes requested):
Looks almost good to me, thanks! @AudricV also merge if you want (feel free to ask me to stop pinging you on each extractor PR I approve ;-) )

extractor/src/main/java/org/schabi/newpipe/extractor/utils/LocaleCompat.java
```diff
@@ -0,0 +1,36 @@
+package org.schabi.newpipe.extractor.utils;
+
+import java.util.Locale;
+
+public class LocaleCompat {
```
AudricV, line 5: Could you add some documentation that this class is made to keep compatibility with Android old APIs? Also maybe precise that if the missing method is made later available with desugarging, this way should be preferred.

extractor/src/main/java/org/schabi/newpipe/extractor/utils/LocaleCompat.java
```diff
@@ -0,0 +1,36 @@
...
+    private LocaleCompat() {
+    }
+
+    // Source: LocaleListCompat's private forLanguageTagCompat() method.
+    // Use Locale.forLanguageTag() on API level >= 21 instead.
```
AudricV, line 10: Precise the comment to make sure that we understand this is the Android API level, not the Java one.

extractor/src/main/java/org/schabi/newpipe/extractor/stream/SubtitlesStream.java
Isira-Seneviratne: > > Then this PR would raise that requirement to 21, right? Do we want that?
> 
> If I understand correctly, the purpose of this PR is actually to prevent raising the requirement to 21, while being able to use `LocaleCompat.forLanguageTag()`

The idea was to eventually switch to `Locale.forLanguageTag()` if/when it becomes available via desugaring.

Isira-Seneviratne:
> > Then this PR would raise that requirement to 21, right? Do we want that?
> 
> If I understand correctly, the purpose of this PR is actually to prevent raising the requirement to 21, while being able to use `LocaleCompat.forLanguageTag()`

The idea was to eventually switch to Locale.forLanguageTag() if/when it becomes available via desugaring.

Stypox:
I think it's better if you avoid reflection

extractor/src/main/java/org/schabi/newpipe/extractor/utils/LocaleCompat.java
```diff
@@ -0,0 +1,43 @@
...
+            } else if (args.length > 1) {
+                return new Locale(args[0], args[1]);
+            } else if (args.length == 1) {
+                return new Locale(args[0]);
+            }
+        } else {
+            return new Locale(str);
+        }
+
+        throw new IllegalArgumentException("Can not parse language tag: [" + str + "]");
```
Stypox, line 41: Note: this is unreachable since `args.length` > 0

Isira-Seneviratne: Yeah, I copied over the method implementation from AndroidX's [`LocaleListCompat`](https://developer.android.com/reference/androidx/core/os/LocaleListCompat) class, as noted in the comment above the method.

AudricV (approved):
LGTM now!

Stypox (approved):

PR #912, opened 2022-08-25
Downgrade Rhino since it uses a class not available on Android

Stypox:
Reverts #774
Fixes TeamNewPipe/NewPipe#8876
See https://github.com/mozilla/rhino/issues/1149 for why this is needed

### DIFF
--- a/extractor/build.gradle
+++ b/extractor/build.gradle
@@ -27,9 +27,12 @@ dependencies {
 
     implementation "com.github.TeamNewPipe:nanojson:$nanojsonVersion"
     implementation 'org.jsoup:jsoup:1.15.3'
-    implementation 'org.mozilla:rhino:1.7.14'
     implementation "com.github.spotbugs:spotbugs-annotations:$spotbugsVersion"
     implementation 'org.nibor.autolink:autolink:0.10.0'
+
+    // do not upgrade to 1.7.14, since in 1.7.14 Rhino uses the `SourceVersion` class, which is not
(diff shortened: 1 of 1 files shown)


AudricV (approved):

Stypox:
I tested and it actually works

leslie-alldridge:
Legend @Stypox
To get this fix do we check for updates in app or is there a dev build we should download? Cheers

Theta-Dev:
I have already made a debug build which includes thia fix as well as a JS extractor change

https://github.com/TeamNewPipe/NewPipe/files/9387795/app-debug.zip

nontekkie:
I am not a developer (far from it), but I came here because my NewPipe app keeps breaking. Is the link above (https://github.com/TeamNewPipe/NewPipe/files/9387795/app-debug.zip) the fix I need if videos load infinitely? If so,  once I download it, do I have to perform more steps, or will the download automatically modify my app? Thanks

nontekkie:
Thanks. I've been waiting for an update from my app store, but it hasn't been released yet. (I use a Murena Teracube running a non-Android OS forked from Lineage. These phones have an app suite with open source apps and also Google apps.) I tried to download the latest update from the NewPipe website, but that was unsuccessful. Any other suggestions?

Sent from ProtonMail mobile

-------- Original Message --------

leslie-alldridge:
Lol your setup sounds overly complicated. Don't bother with an app store just download directly from their website. If that doesn't work you need to allow apps from unknown sources. If that fails, I doubt there's a terracube custom os specialist so you're probably on ya own

PR #920, opened 2022-08-28
[YouTube] Fix extraction of the Trending kiosk with the new data model returned

### LINKED ISSUE #899: [YouTube] Trending is empty sometimes
On YouTube, NewPipeExtractor returns an empty array of `relatedItems`.

This affects both Piped and even NewPipe.

Screenshot_20220817-195413_NewPipe

AudricV:
  I didn't test the changes on NewPipe but tested multiple times the changes using the `YoutubeTrendingKioskInfoTest` test class

This pull request adds support of the new data model returned for trends, which is A/B tested or deployed by YouTube.

This new structure allow us to filter easily `Trending shorts` and `Recently trending` sections.

On the previous one, this `Recently trending` section is now filtered (the `Trending shorts` one uses a different structure, which so was already not extracted), by checking whether sections have a title, which isn't the case for normal trends contrary to the other ones.

This makes that the extractor returns now only the real 50 `Now` YouTube trends, with both data models returned by YouTube.

Elements inside arrays are now extracted dynamically instead of only the ones of the first index, using Java 8's Stream API.

**Note**: The `getInitialPage()` method of `YoutubeTrendingExtractor` can now throw a `ParsingException` if no selected tab (corresponding to the one of the trends type extracted) has been found.

Finally, the licence header of this class has been moved to the top of the file and updated.

Fixes #899, fixes TeamNewPipe/NewPipe#8835.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeTrendingExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeTrendingExtractor.java
@@ -1,26 +1,25 @@
-package org.schabi.newpipe.extractor.services.youtube.extractors;
-
 /*
  * Created by Christian Schabesberger on 12.08.17.
  *
  * Copyright (C) Christian Schabesberger 2018 <chris.schabesberger@mailbox.org>
- * YoutubeTrendingExtractor.java is part of NewPipe.
+ * YoutubeTrendingExtractor.java is part of NewPipe Extractor.
  *
- * NewPipe is free software: you can redistribute it and/or modify
+ * NewPipe Extractor is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
  * (at your option) any later version.
  *
- * NewPipe is distributed in the hope that it will be useful,
+ * NewPipe Extractor is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeTrendingExtractor.java
```diff
@@ -92,25 +91,64 @@ public String getName() throws ParsingException {
...
-        for (final Object itemSectionRenderer : itemSectionRenderers) {
-            final JsonObject expandedShelfContentsRenderer = ((JsonObject) itemSectionRenderer)
-                    .getObject("itemSectionRenderer").getArray("contents").getObject(0)
-                    .getObject("shelfRenderer").getObject("content")
-                    .getObject("expandedShelfContentsRenderer");
-            for (final Object ul : expandedShelfContentsRenderer.getArray("items")) {
-                final JsonObject videoInfo = ((JsonObject) ul).getObject("videoRenderer");
-                collector.commit(new YoutubeStreamInfoItemExtractor(videoInfo, timeAgoParser));
-            }
+        final JsonObject tabContent = getTrendingTabRenderer().getObject("content");
```
Stypox, line 97: Maybe move the `.getObject("content")` to `getTrendingTabRenderer()`, adding `has("content")` to the filtering there?

Stypox (changes requested):
Lgtm other than that small detail, thank you :-)

Stypox (approved):

PR #925, opened 2022-09-09
[YouTube] Fix throttling parameter decryption on Android

Theta-Dev:
- no API changes

I finally found out why the ThrottlingDecryptor keeps failing on NewPipe, while all extractor tests are passing.
Android apparently uses a different regex engine than regular java (see this SO thread https://stackoverflow.com/questions/45074813/regex-pattern-error-in-android-when-matching-closing-brace).

That's why the regex used to extract the nsig function fails to compile with a syntax error on android, unless you escape the curly brace with a slash.


AudricV (approved):
I force-pushed to your branch to suppress the IDE warning about the redundant escape with a comment explaining why this is done, based on what you said in the PR body.

Thank you for finding this!

Stypox (approved):
Great, lgtm :-)

PR #926, opened 2022-09-10
[YouTube] Fix extraction of like count with the new data model

AudricV:
  I didn't test the changes on NewPipe but I ran multiple times YouTube's StreamExtractor tests and the like counts' ones always passed.

In the new data model currently A/B tested or deployed, the like count button data is the same, but its path has been changed.

The extraction of the like count has been also improved, by using the multiple accessibility data available instead of the only one which was used before. This accessibility data has been also deprioritized, because it is language dependent when there is no view.

Like count is always known, even for age-restricted videos, so returning -1 if the like count extraction fails on this type of videos is not needed and hides an extraction error: that's the reason why this return has been removed and the exception will always be thrown, even if a video is age-restricted.

Fixes TeamNewPipe/NewPipe#8892.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
@@ -417,28 +417,88 @@ public class YoutubeStreamExtractor extends StreamExtractor {
     @Override
     public long getLikeCount() throws ParsingException {
         assertPageFetched();
+
+        // If ratings are not allowed, there is no like count available
+        if (!playerResponse.getObject("videoDetails").getBoolean("allowRatings")) {
+            return -1;
+        }
+
         String likesString = "";
+
         try {
-            likesString = getVideoPrimaryInfoRenderer()
+            final JsonArray topLevelButtons = getVideoPrimaryInfoRenderer()
                     .getObject("videoActions")
                     .getObject("menuRenderer")
-                    .getArray("topLevelButtons")
-                    .getObject(0)
-                    .getObject("toggleButtonRenderer")
-                    .getObject("defaultText")
-                    .getObject("accessibility")
+                    .getArray("topLevelButtons");
+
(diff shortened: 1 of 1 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -417,28 +417,88 @@ public long getViewCount() throws ParsingException {
...
+                                || toggleButtonRenderer.getObject("defaultIcon")
+                                .getString("iconType")
+                                .equalsIgnoreCase("LIKE")
+                                || toggleButtonRenderer.getObject("toggleButtonSupportedData")
+                                .getObject("toggleButtonIdData")
+                                .getString("id")
+                                .equalsIgnoreCase("TOGGLE_BUTTON_ID_TYPE_LIKE"))
+                        .findFirst()
+                        .orElseThrow(() -> new ParsingException(
+                                "Ratings are enabled even though the like button is missing"));
```
Stypox, line 471: This is the wrong way around:
```suggestion
                                "The like button is missing even though ratings are enabled"));
```

Stypox (changes requested):
Other than that lgtm

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
AudricV: It was already as it is before my changes, but I will change this.

<em>Why I am blamed for code that I didn't wrote :/</em>

opusforlife2: No one is blaming you. xD We're just striking the iron while it's hot.

opusforlife2 (approved):
Here you go, since it was just a one line language change.

Stypox (approved):

PR #927, opened 2022-09-13
[YouTube] Return mqdefault thumbnails in fast feed

Stypox:

The `hqdefault` thumbnails, which are returned by default by the fast feed extractor, have black bars at the top and at the bottom, and are thus not optimal. So this PR just replaces `hqdefault` with `mqdefault`: it seems to work well without issues after some testing, and unlike `highresdefault` the `mqdefault` thumbnail always exists according to [this SO](https://stackoverflow.com/a/20542029/9481500). This fixes the problem with black bands in NewPipe: https://github.com/TeamNewPipe/NewPipe/issues/8890#issuecomment-1236708859

Debug apk: [app-debug.zip](https://github.com/TeamNewPipe/NewPipeExtractor/files/9557474/app-debug.zip)
Both me (Stypox) and [a user](https://github.com/TeamNewPipe/NewPipe/issues/8890#issuecomment-1245467976) tested the apk and it works. All thumbnails load.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeFeedInfoItemExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeFeedInfoItemExtractor.java
@@ -91,6 +91,10 @@ public class YoutubeFeedInfoItemExtractor implements StreamInfoItemExtractor {
 
     @Override
     public String getThumbnailUrl() {
-        return entryElement.getElementsByTag("media:thumbnail").first().attr("url");
+        // The hqdefault thumbnail has some black bars at the top and at the bottom, while the
+        // mqdefault doesn't, so return the mqdefault one. It should always exist, according to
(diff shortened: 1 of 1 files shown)


opusforlife2:
"mq" sounds like "medium quality". Is that the case? If so, do we want to lower the resolution like that? Also, will this still be needed after Audric's image PR?

Stypox:
See the discussion on IRC. The answers are: yes, yes and yes. It is better to have a medium quality thumbnail without black bars than a high quality one with black bars. Also, this is a change just to the fast feed extractor, for which it is expected that some information might be missing, so we should try our best to get the most out of the information we can (the black bars look very ugly). This will still be needed after AudricV's PR since this PR does not extract a new type of image URL, but rather generates such URL (which is something AudricV's PR doesn't do with youtube (and it is the right thing to do, generally)).

Stypox:
I'm merging this for now. If there are better solutions in the future, it's not difficult to revert this small PR.

PR #928, opened 2022-09-17
Parse YouTube comments as HTML

FireMasterK:

This change is to add support for clickable timestamps and URLs.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsInfoItemExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsInfoItemExtractor.java
@@ -183,7 +183,7 @@ public class YoutubeCommentsInfoItemExtractor implements CommentsInfoItemExtract
                 // https://github.com/TeamNewPipe/NewPipeExtractor/issues/380#issuecomment-668808584
                 return "";
             }
-            final String commentText = getTextFromObject(contentText);
+            final String commentText = getTextFromObject(contentText, true);
             // YouTube adds U+FEFF in some comments.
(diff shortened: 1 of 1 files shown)


Stypox (approved):
Looks good to me, thanks!
@AudricV feel free to merge these small PRs without waiting for my or anyone's approval ;-)

AudricV:
@Stypox I was not agree to merge this. `getCommentText` should return a plain text, not HTML (see PeerTube's `CommentsInfoItemExtractor` implementation).

Also, how NewPipe would handle this? It will probably display the comment with HTML tags.

If we want to return HTML, we should generalize the `Description` class, like already suggested, and return this class instead of a `String`.

Stypox:
Well then write a comment if you already thought about this... When I looked at this PR I didn't find any comment saying it should not be merged...

Just like descriptions can come in many formats, can't comments do so too?

FireMasterK:
> @Stypox I was not agree to merge this. `getCommentText` should return a plain text, not HTML (see PeerTube's `CommentsInfoItemExtractor` implementation).

Express that! Even if its just 1 line it helps.

> Also, how NewPipe would handle this? It will probably display the comment with HTML tags.

Descriptions currently are also HTML? Currently, I think it may do that. I think it should be easy to fix that?

> If we want to return HTML, we should generalize the `Description` class, like already suggested, and return this class instead of a `String`.

That could be done indeed. Would you like a PR for this?

AudricV:
> Express that! Even if its just 1 line it helps.

I didn't have the time to do so and then I forgot a bit this PR, sorry!

> That could be done indeed. Would you like a PR for this?

Yes, it would be great!

Stypox:
> I didn't have the time to do so and then I forgot a bit this PR, sorry!

That's fine, commits can be reverted anyway in case the Description PR takes some time ;-)

PR #936, opened 2022-09-25
[YouTube] Add comment reply count support

### LINKED ISSUE #935: Add comment reply count support
Show how many comment replies just as youtube "2 REPLIES"

xz-dev:

Add comment reply count support, just like youtube.

Fix #935

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfoItemExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfoItemExtractor.java
@@ -114,10 +114,11 @@ public interface CommentsInfoItemExtractor extends InfoItemExtractor {
     /**
      * The count of comment replies.
      *
-     * @return the count of the replies, or -1 if replies are not supported
+     * @return the count of the replies
+     * or {@link CommentsInfoItem#UNKNOWN_REPLY_COUNT} if replies are not supported
      */
     default int getReplyCount() throws ParsingException {
-        return -1;
+        return CommentsInfoItem.UNKNOWN_REPLY_COUNT;
     }
 
     /**
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfoItemExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfoItemExtractor.java
@@ -16,14 +16,14 @@ public interface CommentsInfoItemExtractor extends InfoItemExtractor {
      * or {@link CommentsInfoItem#NO_LIKE_COUNT} if it is unavailable.
      *
      * <br>
-     *
+     * <p>
      * NOTE: Currently only implemented for YT {@link
      * YoutubeCommentsInfoItemExtractor#getLikeCount()}
      * with limitations (only approximate like count is returned)
      *
-     * @see StreamExtractor#getLikeCount()
      * @return the comment's like count
      * or {@link CommentsInfoItem#NO_LIKE_COUNT} if it is unavailable
+     * @see StreamExtractor#getLikeCount()
      */
     default int getLikeCount() throws ParsingException {
         return CommentsInfoItem.NO_LIKE_COUNT;
@@ -103,14 +103,26 @@ public interface CommentsInfoItemExtractor extends InfoItemExtractor {
 
     /**
      * The playback position of the stream to which this comment belongs.
+     *
      * @see CommentsInfoItem#getStreamPosition()
      */
     default int getStreamPosition() throws ParsingException {
         return CommentsInfoItem.NO_STREAM_POSITION;
     }
 
+
+    /**
+     * The count of comment replies.
+     *
+     * @return the count of the replies, or -1 if replies are not supported
+     */
+    default int getReplyCount() throws ParsingException {
+        return -1;
+    }
+
     /**
      * The continuation page which is used to get comment replies from.
+     *
      * @return the continuation Page for the replies, or null if replies are not supported
      */
     @Nullable
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfoItemExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfoItemExtractor.java
@@ -110,7 +110,6 @@ public interface CommentsInfoItemExtractor extends InfoItemExtractor {
         return CommentsInfoItem.NO_STREAM_POSITION;
     }
 
-
     /**
      * The count of comment replies.
      *
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsInfoItemExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsInfoItemExtractor.java
@@ -251,9 +251,9 @@ public class YoutubeCommentsInfoItemExtractor implements CommentsInfoItemExtract
 
     @Override
     public int getReplyCount() throws ParsingException {
-        final JsonObject cr = getCommentRenderer();
-        if (cr.has("replyCount")) {
-            return cr.getInt("replyCount");
+        final JsonObject commentRenderer = getCommentRenderer();
+        if (commentRenderer.has("replyCount")) {
+            return commentRenderer.getInt("replyCount");
         }
         return UNKNOWN_REPLY_COUNT;
     }
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsInfoItemExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsInfoItemExtractor.java
@@ -249,6 +249,15 @@ public class YoutubeCommentsInfoItemExtractor implements CommentsInfoItemExtract
     }
 
     @Override
+    public int getReplyCount() throws ParsingException {
+        try {
+            return JsonUtils.getNumber(json, "comment.commentRenderer.replyCount").intValue();
+        } catch (final Exception e) {
+            return -1;
+        }
+    }
+
+    @Override
     public Page getReplies() throws ParsingException {
         try {
             final String id = JsonUtils.getString(
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsInfoItemExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsInfoItemExtractor.java
@@ -1,5 +1,6 @@
 package org.schabi.newpipe.extractor.services.youtube.extractors;
 
+import static org.schabi.newpipe.extractor.comments.CommentsInfoItem.UNKNOWN_REPLY_COUNT;
 import static org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper.getTextFromObject;
 
 import com.grack.nanojson.JsonArray;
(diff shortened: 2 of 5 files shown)


TobiGr (changes requested):
Thank you for that addition.

Please implement the [tests for YouTube](https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeCommentsExtractorTest.java).

The reply count should also be extracted for the other services (e.g. PeerTube).

extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfoItemExtractor.java
```diff
@@ -103,14 +103,26 @@ default boolean isUploaderVerified() throws ParsingException {
...
     }
 
+
+    /**
+     * The count of comment replies.
+     *
+     * @return the count of the replies, or -1 if replies are not supported
+     */
+    default int getReplyCount() throws ParsingException {
+        return -1;
```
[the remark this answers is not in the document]
xz-dev: Constant, I forget it.

AudricV (changes requested):
Additional changes that should be made.

Also, please remove the unneeded JavaDoc changes, which were probably made by your IDE.

extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfoItemExtractor.java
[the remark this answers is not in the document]
xz-dev: Done.

xz-dev:
> Additional changes that should be made.
> 
> Also, please remove the unneeded JavaDoc changes, which were probably made by your IDE.

Restored

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsInfoItemExtractor.java
```diff
@@ -248,6 +248,15 @@ public String getUploaderUrl() throws ParsingException {
         }
     }
 
+    @Override
+    public int getReplyCount() throws ParsingException {
+        try {
+            return JsonUtils.getNumber(json, "comment.commentRenderer.replyCount").intValue();
+        } catch (final Exception e) {
+            return -1;
```
TobiGr, line 256: Please use `UNKNOWN_REPLY_COUNT`

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeCommentsExtractorTest.java
```diff
@@ -328,5 +329,17 @@ public void testGetCommentsFirstReplies() throws IOException, ExtractionExceptio
...
+
+        @Test
+        public void testGetCommentsReplyCount() throws IOException, ExtractionException {
+            final InfoItemsPage<CommentsInfoItem> comments = extractor.getInitialPage();
+
+            DefaultTests.defaultTestListOfItems(YouTube, comments.getItems(), comments.getErrors());
+
+            CommentsInfoItem firstComment = comments.getItems().get(0);
+
+            assertNotEquals(-1, firstComment.getReplyCount(), "First reply comment can't get count");
```
TobiGr, line 341: Use `UNKNOWN_REPLY_COUNT`

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeCommentsExtractorTest.java
```diff
@@ -328,5 +329,17 @@ public void testGetCommentsFirstReplies() throws IOException, ExtractionExceptio
...
+            assertNotEquals(0, firstComment.getReplyCount(), "First reply comment count is zero");
```
TobiGr, line 342:
```suggestion
            assertGreater(300, firstComment.getReplyCount(), "First reply comment count is smaller than the expected 300 comments");

```

TobiGr (commented):
Just some small things from my end

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeCommentsExtractorTest.java
xz-dev: Done. https://github.com/TeamNewPipe/NewPipeExtractor/pull/936/commits/4b4c7e869a22d22fc7a52ad76cf2ed36c8667a17

TobiGr:
> The reply count should also be extracted for the other services (e.g. PeerTube).

Implementing this for PeerTube should be the same effort as you had with YouTube - less than ten lines. Can you please take care of it?

xz-dev:
> > The reply count should also be extracted for the other services (e.g. PeerTube).
> 
> Implementing this for PeerTube should be the same effort as you had with YouTube - less than ten lines. Can you please take care of it?

We even not support get the PeerTube comment reply.
So, I think it's that's another job what we need to do.

xz-dev:
I think the pull request is ready been merge, is there any questions?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsInfoItemExtractor.java
```diff
@@ -248,6 +249,15 @@ public String getUploaderUrl() throws ParsingException {
...
 
+    @Override
+    public int getReplyCount() throws ParsingException {
+        try {
+            return JsonUtils.getNumber(json, "comment.commentRenderer.replyCount").intValue();
+        } catch (final Exception e) {
+            return UNKNOWN_REPLY_COUNT;
+        }
+    }
+
```
TobiGr, line 260: @Stypox @AudricV Is there any reason, why these have a try & catch? Shouldn't we throw an exception to get informed about the problem?

xz-dev: I think you are right, but we need handle error in UI.

xz-dev: But if continue to think about it, if Youtube change API, we need adapt it as quick as possible but not hide it.
So, I think we can throw the error and in Newpipe client, we check reply page and count, if we have one of these, that means there have comment replies.

xz-dev: That means we need change [this line](https://github.com/TeamNewPipe/NewPipeExtractor/blob/4b4c7e869a22d22fc7a52ad76cf2ed36c8667a17/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsInfoItemExtractor.java#L262) also.

TobiGr: All non-critical errors thrown here, are added to an error list in the collectors. They are  displayed via a snackbar or notification in NewPipe automatically.
I was asking @Stypox and @AudricV regarding the other code, because some of these thrown errors might just mean, that the info is not present due to some settings by the content creator.
The comment count however, should always be available and thus exceptions thrown.

xz-dev: It's done https://github.com/TeamNewPipe/NewPipeExtractor/pull/936/commits/c4f6a69dbcbdad86207a25b1cb4e6ba8a4558237

TobiGr:
I added a commit to fix exceptions when comments do not have replies and thus no `replyCount`

TobiGr (approved):
Thank you. LGTM now

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsInfoItemExtractor.java
```diff
@@ -248,6 +249,15 @@ public String getUploaderUrl() throws ParsingException {
         }
     }
 
+    @Override
+    public int getReplyCount() throws ParsingException {
+        final JsonObject cr = getCommentRenderer();
```
AudricV, line 254: Maybe use a better name here, like `commentRenderer`?

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeCommentsExtractorTest.java
```diff
@@ -328,5 +331,17 @@ public void testGetCommentsFirstReplies() throws IOException, ExtractionExceptio
...
                     "First reply comment did not match");
         }
+
+        @Test
+        public void testGetCommentsReplyCount() throws IOException, ExtractionException {
+            final InfoItemsPage<CommentsInfoItem> comments = extractor.getInitialPage();
+
+            DefaultTests.defaultTestListOfItems(YouTube, comments.getItems(), comments.getErrors());
+
+            CommentsInfoItem firstComment = comments.getItems().get(0);
```
AudricV, line 341: Use `final` here, like it is required in the normal code.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeCommentsExtractorTest.java
```diff
@@ -328,5 +331,17 @@ public void testGetCommentsFirstReplies() throws IOException, ExtractionExceptio
...
+
+            assertNotEquals(UNKNOWN_REPLY_COUNT, firstComment.getReplyCount(), "First reply comment can't get count");
```
AudricV, line 343:
```suggestion
            assertNotEquals(UNKNOWN_REPLY_COUNT, firstComment.getReplyCount(), "Could not get the reply count of the first comment");
```

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeCommentsExtractorTest.java
```diff
@@ -328,5 +331,17 @@ public void testGetCommentsFirstReplies() throws IOException, ExtractionExceptio
...
+            assertGreater(300, firstComment.getReplyCount(), "First reply comment count is smaller than the expected 300 comments");
```
AudricV, line 344: I think you should simply use [`assertGreater(long, long)`](https://github.com/TeamNewPipe/NewPipeExtractor/blob/197d8876fcbc8126fb74d1c45563f1fe3bef0bcd/extractor/src/test/java/org/schabi/newpipe/extractor/ExtractorAsserts.java#L77), which will show the message `ACTUAL_COMMENT_REPLY_COUNT  is not > EXPECTED_COMMENT_REPLY_COUNT`.

```suggestion
            assertGreater(300, firstComment.getReplyCount());
```

This has the advantage to know the real count extracted, which is useful to update the expected count if some users remove their comment reply(ies), and so make that the comment reply count is under the expected value.

AudricV (changes requested):
A few changes remaining and this can be merged in my opinion.

Thank you for this feature!

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeCommentsExtractorTest.java
xz-dev: I had copy from [here](https://github.com/TeamNewPipe/NewPipeExtractor/blob/197d8876fcbc8126fb74d1c45563f1fe3bef0bcd/extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeCommentsExtractorTest.java#L325).

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeCommentsExtractorTest.java
xz-dev: Done. https://github.com/TeamNewPipe/NewPipeExtractor/commit/f88e235089f4dc8d2b260192f1ebd01dd4dee20d

AudricV (approved):
LGTM, thank you very much!

Please avoid in the future writing usernames with `@` in commits, as if they contain these usernames in this form, you would probably send a GitHub notification each time the commits are rebased and/or pushed (that's what happened for me with your latest commits), including in repositories forks if I am not wrong.

This PR will be merged using the `Squash and merge` command.

xz-dev:
Ok, I'll, usernames with @ in commits is github auto add it here, I don't know why github do that, so I keep it here.

PR #937, opened 2022-09-26
Fix tests

TobiGr:


TobiGr:
PeerTube now returns different (smaller) thumbnails and uploader avatars!

AudricV:
We see often test failures due to PeerTube thumbnail or avatar URL changes. I would recommend changing what is tested in these tests, in order to not have to update them frequently.

TobiGr:
What about creating a test that is checking whether requesting an URL returns a 200 status code? Sould be easy to implement using `Downloader.head(url)`. However, we might need to check whether we penetrate PeerTube servers to much with those requests. Maybe those tests should only be run once during the nightly test cycle and not with every test suit for e.g. PRs and commits.

AudricV:
I would check instead if the URL we get contains something.

Issue #942, opened 2022-10-10
Add ability to identify short-form stream items 


chowder:
Content from TikTok, YouTube Shorts, and Instagram Reels are known as "short-form content". 

There should be a way to distinguish between regular media and short-form media, e.g. for filtering in upstream apps. This would be especially useful in situations where a single extractor extracts both regular and short-form content (e.g. YouTube). 

I can see this being supported by either: 
1. Implementing a `isShortForm` method on `StreamInfo` (and related) classes 
2. Extending the enum list of `StreamType` to support `SHORT_FORM_VIDEO`/`SHORT_FORM_AUDIO` 

With the second option I suppose we'd be teetering into something more suitable as a bit flag.

I'm happy to have a go at implementing this - appreciate any advice/suggestions.

chowder:
Implemented in https://github.com/TeamNewPipe/NewPipeExtractor/pull/946

PR #946, opened 2022-10-17
Add ability to identify short-form `StreamInfoItem`s

chowder:
Addresses https://github.com/TeamNewPipe/NewPipeExtractor/issues/942 


TobiGr:
thank you. I think providing the same information for `StreamInfo` would be nice. Can you do this as well?

chowder:
I'll have a look to see if I can do the same for `StreamInfo`

Also, I noticed that a number of tests are failing in the CI action, but these failures are unrelated to my changes (the test that I introduced is passing), do I have to fix them?

TobiGr:
You don't need to fix the unrelated tests, I created a PR to fix them.
However, [this test](https://github.com/TeamNewPipe/NewPipeExtractor/actions/runs/3276083975/jobs/5393324731#step:5:246) should be fixed by you.

chowder:
Ah thanks, I've regenerated the test mocks now. 

I've also added the same `isShortFormContent` methods for `StreamInfo` and `StreamExtractor`, but haven't added any actual implementations; for YouTube in particular I couldn't find any fields that distinguish between regular and short videos on the `/watch?v=` links (using [this video](https://www.youtube.com/watch?v=R1Ehb3JA-cM) as reference.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
```diff
@@ -324,4 +324,16 @@ public String getShortDescription() throws ParsingException {
...
+
+    @Override
+    public boolean isShortFormContent() throws ParsingException {
+        try {
+            final String webPageType = videoInfo.getObject("navigationEndpoint")
+                    .getObject("commandMetadata").getObject("webCommandMetadata")
+                    .getString("webPageType");
+            return !isNullOrEmpty(webPageType) && webPageType.equals("WEB_PAGE_TYPE_SHORTS");
+        } catch (final Exception e) {
+            throw new ParsingException("Could not determine if short-form content", e);
```
AudricV, line 336: This seems not understandable as an exception message: I suggest using something like this (given by @opusforlife2 to me):

```suggestion
            throw new ParsingException("Could not determine if this is short-form content", e);
```

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamExtractor.java
```diff
@@ -554,6 +554,16 @@ public List<MetaInfo> getMetaInfo() throws ParsingException {
         return Collections.emptyList();
     }
 
+    /**
+     * Whether the stream is a short-form content. These are content in the style of TikTok,
+     * YouTube Shorts, or Instagram Reels videos.
```
AudricV, line 559:
```suggestion
     * Whether the stream is a short-form content.
     *
     * <p>
     * Short-form contents are contents in the style of TikTok, YouTube Shorts, or Instagram Reels videos.
     * </p>
```

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamInfoItem.java
```diff
@@ -130,6 +131,15 @@ public void setUploaderVerified(final boolean uploaderVerified) {
...
 
+    public boolean isShortFormContent() {
+        return shortFormContent;
+    }
+
+    public void setShortFormContent(final boolean shortFormContent) {
+        this.shortFormContent = shortFormContent;
+    }
+
+
```
AudricV, line 142: Remove this extra blank line.

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamInfoItemExtractor.java
```diff
@@ -127,4 +127,14 @@ public interface StreamInfoItemExtractor extends InfoItemExtractor {
     default String getShortDescription() throws ParsingException {
         return null;
     }
+
+    /**
+     * Check if the stream is a short-form content
+     *
+     * @return {@code true} if the stream is a short-form content
+     * @throws ParsingException thrown if there is an error in the extraction
+     */
```
AudricV, line 136: You should use the first description everywhere (with my changes, I think), which give us more details about what is this type of contents.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/search/YoutubeSearchExtractorTest.java
```diff
@@ -379,4 +380,49 @@ public void testVideoDescription() throws IOException, ExtractionException {
...
+            final List<InfoItem> items = extractor.getInitialPage().getItems();
+            boolean hasShortFormContent = false;
+            for (InfoItem item : items) {
+                if (((StreamInfoItem) item).isShortFormContent()) {
+                    hasShortFormContent = true;
+                    break;
+                }
+            }
+
+            assertTrue(hasShortFormContent);
```
AudricV, line 425: Remove the unneeded `public` test method modifier here. Also, I recommend using for the test Java 8's `Stream` API instead:

```suggestion
        void testShortFormContent() throws IOException, ExtractionException {
            assertTrue(extractor.getInitialPage()
                    .getItems()
                    .stream()
                    .filter(StreamInfoItem.class::isInstance)
                    .map(StreamInfoItem.class::cast)
                    .anyMatch(StreamInfoItem::isShortFormContent));
```

AudricV (changes requested):
Thank you very much! Here are some changes you need to do in this PR.

Could you also rebase it?

FireMasterK:
I've addressed the comments and rebased this PR at https://github.com/FireMasterK/NewPipeExtractor/tree/shorts

chowder:
Thank you so much @FireMasterK, I've cherry-picked your commit addressing the remaining comments.

FireMasterK (approved):

chowder:
> Except that I am not sure if we should use setIsShortFormContent(boolean) instead of setShortFormContent(boolean) for setters

I agree, but I was trying to be consistent with other boolean setters - e.g. [`isUploaderVerified`/`setUploaderVerified`](https://github.com/TeamNewPipe/NewPipeExtractor/blob/b1a899fd47ca5db9bc7cf65ac70a5f20178fb47d/extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamInfo.java#L537) 😅

FireMasterK:
Intellij's setter generator would generate them as `setShortFormContent(boolean)`, I would keep it the way it is. (and to keep things consistent)

PR #950, opened 2022-10-23
Add search link type in the extractor

Sola0404:

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/StreamingService.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/StreamingService.java
@@ -86,7 +86,8 @@ public abstract class StreamingService {
         NONE,
         STREAM,
         CHANNEL,
-        PLAYLIST
+        PLAYLIST,
+        SEARCH,
     }
 
     private final int serviceId;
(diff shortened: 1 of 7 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/StreamingService.java
```diff
@@ -282,6 +290,27 @@ public CommentsExtractor getCommentsExtractor(final String url) throws Extractio
     // Utils
     //////////////////////////////////////////////////////////////////////////*/
 
+    public final String getIdByUrl(final String url) throws ParsingException{
```
Stypox, line 293: Instead of adding this function just to use it for search URLs in the client, I think a better option is what I suggested in the client review

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeSearchLinkHandlerFactory.java
```diff
@@ -0,0 +1,63 @@
...
+    public String getId(String url) throws ParsingException {
+        try {
+            final URL urlObj = Utils.stringToURL(url);
+
+            if (!Utils.isHTTP(urlObj) || !(YoutubeParsingHelper.isYoutubeURL(urlObj)
+                    || YoutubeParsingHelper.isInvidioURL(urlObj))) {
+                throw new ParsingException("the url given is not a YouTube-URL");
+            }
+
+            final String listID = Utils.getQueryValue(urlObj, "search_query");
```
Stypox, line 40: Change `listId` to `searchQuery` and update the exception message accordingly

Stypox (changes requested):
Looks almost good to me. @FireMasterK could you take a look, too? Also, do we need tests for the link handler?

FireMasterK:
> Looks almost good to me. @FireMasterK could you take a look, too? Also, do we need tests for the link handler?

Honestly, I don't know how this part of the extractor works at all, so I would be the wrong candidate.

We likely do need tests, since I did check the Channel link handler, which has tests.

Stypox:
@Sola0404 do you plan to keep working on this? Also, you should also implement this for all other services, and not just for youtube, if possible (shouldn't be too difficult though).

PR #953, opened 2022-10-24
[YouTube] Add support for attributed text description

Theta-Dev:
- no API changes

YouTube is currently A/B testing a new video description format. The new format includes chips with the YouTube icon for internal links, as shown below:

image

Internally the new format does not use a list of text components but a string with the entire plaintext content along with a list of `commandRuns` which include the link data. I included a sample file for testing (`attributed_description.json`).

Because NewPipeExtractor falls back to the description extracted from the player endpoint, being unable to extract the new description format does not cause any issues when using NewPipe. However this should fix the description unit tests.


extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelperTest.java
```diff
@@ -48,4 +54,12 @@ public void testConvertFromGoogleCacheUrl() {
...
+    }
```
FireMasterK, line 64: Why not just update the mocks instead?

FireMasterK (approved):

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelperTest.java
```diff
@@ -1,8 +1,5 @@
 package org.schabi.newpipe.extractor.services.youtube;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-
```
AudricV, line 5: Could you revert this?

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelperTest.java
```diff
@@ -12,6 +9,9 @@
 
 import java.io.IOException;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
```
AudricV, line 14: Same as above.

AudricV (changes requested):
Thanks, please revert the test changes which are unrelated to your PR (Bandcamp + PeerTube + Learning playlist) and address the comments above.

AudricV:
I resolved the conflicts, rebased the PR, removed the test changes (please open a separate PR for that), improved a bit the documentation of the method you added and squashed all the commits, as they were redundant.

Thanks for your hard work!

AudricV (approved):

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
```diff
@@ -1009,6 +1009,90 @@ public static String getTextFromObject(final JsonObject textObject, final boolea
...
+    @Nullable
+    public static String getAttributedDescription(
+            @Nullable final JsonObject attributedDescription) {
+        if (isNullOrEmpty(attributedDescription)) {
+            return null;
+        }
+
+        final String content = attributedDescription.getString("content");
+        final JsonArray commandRuns = attributedDescription.getArray("commandRuns");
+        if (content == null || commandRuns.isEmpty()) {
```
Stypox, line 1032: Isn't it ok for commandRuns to be empty?

Stypox (changes requested):

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
FireMasterK: We should just return `content` in such cases imo

Theta-Dev: Yes, this check can be omitted as the remaining text is always appended, even if threre are no command runs

PR #958, opened 2022-10-29
[YouTube] Support new metadata format of playlists

AudricV:
- [ ] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe). As the issue is pretty rare and both formats can be returned at the same time, I don't know when I can get only the new one. Here is a link to a debug APK built by the CI with the changes: https://github.com/AudricV/NewPipe/actions/runs/3352327380
- [ ] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API. No API changes, but an extractor update will be required in it

This PR adds support of the new playlist layout which is currently A/B tested. This new layout doesn't provide the uploader avatar.

This fixes the extraction of the videos count and the uploader name when this interface is got by the extractor.

The code of learning playlists video count check has been also removed, as it seems to be not relevant anymore (the video count seems to be returned for these playlists with both layouts).

Finally, unneeded overrides of subchannel methods, which don't apply to the YouTube service, have been removed.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubePlaylistExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubePlaylistExtractor.java
@@ -44,16 +44,19 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 public class YoutubePlaylistExtractor extends PlaylistExtractor {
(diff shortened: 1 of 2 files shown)


FireMasterK (approved):

TobiGr (approved):

Stypox:
LGTM

PR #959, opened 2022-10-29
Add uploaderUrl and uploaderVerified to PlaylistInfoItem.

FireMasterK:

Note: There are no tests for this anywhere, so I haven't added any

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/playlist/PlaylistInfoItem.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/playlist/PlaylistInfoItem.java
@@ -2,6 +2,8 @@ package org.schabi.newpipe.extractor.playlist;
 
 import org.schabi.newpipe.extractor.InfoItem;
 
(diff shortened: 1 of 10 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/playlist/PlaylistInfoItem.java
```diff
@@ -23,6 +25,22 @@ public void setUploaderName(final String uploaderName) {
         this.uploaderName = uploaderName;
     }
 
+    public String getUploaderUrl() {
```
TobiGr, line 28: Please annotate this method as well as the parameter in the setter as nullable

TobiGr (commented):
looks good. just a minor note

FireMasterK:
> What about the other services?

I don't have the time or patience to implement them, sorry. I implemented this for Piped (where this PR is already deployed), where YouTube is the only used service.

TobiGr (approved):

PR #971, opened 2022-11-09
[YouTube] Improve duration parsing

lrusso96:

(No API change tho)

I have simplified ad optimized the code to parse the duration of YouTube videos.
In the previous version, the parsing method `parseDurationString` was a bit redundant, making some extra (useless) method calls for short videos.
In particular, there were calls to "sanitize" and check the "0" strings, added within the method.
This code is not only more compact, but only makes the necessary calls to `convertDurationToInt`: e.g., for short videos (<1h), this approach seems twice faster on my device(s).

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
@@ -318,38 +318,16 @@ public final class YoutubeParsingHelper {
                 ? input.split(":")
                 : input.split("\\.");
 
-        String days = "0";
-        String hours = "0";
-        String minutes = "0";
-        final String seconds;
-
-        switch (splitInput.length) {
-            case 4:
-                days = splitInput[0];
-                hours = splitInput[1];
-                minutes = splitInput[2];
-                seconds = splitInput[3];
-                break;
-            case 3:
-                hours = splitInput[0];
-                minutes = splitInput[1];
-                seconds = splitInput[2];
-                break;
-            case 2:
-                minutes = splitInput[0];
-                seconds = splitInput[1];
-                break;
-            case 1:
-                seconds = splitInput[0];
-                break;
-            default:
-                throw new ParsingException("Error duration string with unknown format: " + input);
+        final int[] units = {24, 60, 60, 1};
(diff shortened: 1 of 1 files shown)


Redirion (approved):
lgtm. I like it. Like a small puzzle that you have solved with the offset.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
```diff
@@ -318,38 +318,16 @@ public static int parseDurationString(@Nonnull final String input)
...
-                throw new ParsingException("Error duration string with unknown format: " + input);
-        }
-
-        return ((convertDurationToInt(days) * 24
-                + convertDurationToInt(hours)) * 60
-                + convertDurationToInt(minutes)) * 60
-                + convertDurationToInt(seconds);
+        final int[] units = {24, 60, 60, 1};
+        final int offset = units.length - splitInput.length;
+        if (offset < 0) {
```
acrodemocide, line 323: In the case statement being replaced, the valid values were 1, 2, 3, and 4. Consider changing this if statement to:

if (offset < 1 || offset > 4)

This should preserve the same logic that existed in the if statement.

acrodemocide (commented):
Consider changing the if statement commented on to preserve the logic being replaced for the case statement.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
lrusso96: > In the case statement being replaced, the valid values were 1, 2, 3, and 4. Consider changing this if statement to:
> 
> if (offset < 1 || offset > 4)
> 
> This should preserve the same logic that existed in the if statement.

@acrodemocide `splitInput.length` is a number strictly greater than 0, while `units.length` is equal to (the magic number) 4.
Then `offset` cannot be a number strictly greater than 4.

acrodemocide: Thank you for clarifying.

acrodemocide (approved):
Comment resolved

PR #972, opened 2022-11-13
Use OkHttp's HttpUrl.

Isira-Seneviratne:
----
OkHttp's [`HttpUrl`](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-http-url/) class offers greater functionality compared to the standard [`URL`](https://docs.oracle.com/javase/8/docs/api/java/net/URL.html) class, such as a builder class for creating request URLs.

I created a relatively small changeset in this PR to get feedback on if the idea was sound, with more uses of `HttpUrl` to be added in future PRs.

### DIFF
--- a/extractor/build.gradle
+++ b/extractor/build.gradle
@@ -34,6 +34,9 @@ dependencies {
     // available on Android (even when using desugaring), and `NoClassDefFoundError` is thrown
     implementation 'org.mozilla:rhino:1.7.13'
 
+    //noinspection GradleDependency --> do not update OkHttp to keep supporting Android 4.x users
+    implementation 'com.squareup.okhttp3:okhttp:3.12.13'
+
     checkstyle "com.puppycrawl.tools:checkstyle:$checkstyleVersion"
 
     testImplementation platform("org.junit:junit-bom:$junitVersion")
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/linkHandler/BandcampSearchQueryHandlerFactory.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/linkHandler/BandcampSearchQueryHandlerFactory.java
@@ -6,20 +6,20 @@ import static org.schabi.newpipe.extractor.services.bandcamp.extractors.Bandcamp
 
 import org.schabi.newpipe.extractor.exceptions.ParsingException;
 import org.schabi.newpipe.extractor.linkhandler.SearchQueryHandlerFactory;
-import org.schabi.newpipe.extractor.utils.Utils;
 
-import java.io.UnsupportedEncodingException;
 import java.util.List;
+
+import okhttp3.HttpUrl;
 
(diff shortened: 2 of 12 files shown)


FireMasterK:
In my opinion, we should stick to `java.net.URL` for public methods (so that we don't break people's existing code). But, we can use HttpUrl for places where it adds benefits like query string parsing, like how you did using `queryParameter`.

Isira-Seneviratne:
> In my opinion, we should stick to `java.net.URL` for public methods (so that we don't break people's existing code). 

I updated the existing methods to convert the `HttpUrl` objects to `URL` ones. I could add some additional methods to the code to avoid issues.

FireMasterK:
> I updated the existing methods to convert the `HttpUrl` objects to `URL` ones. I could add some additional methods to the code to avoid issues.

Yes, currently for the methods/constructors you have changed, it should be fine. It's just for any future changes you may make.

acrodemocide (commented):
Cool concept with the new URL library. How does it feel with the malformed URL exceptions? I saw catch statements replaced to catch an argument exception instead of a malformed URL exceptions, so I wasn't quite sure how those are handled with the new library.

The biggest risk I see is that if we start adopting it, is it something that we'll replace the entire app with soon? Or will there be enough pushback back on other pull requests so that we end up with some places using the old URL class and others using the new class? If we get buy in by the owners of the project and can get it replaced relatively quickly and easily, it could be a good addition to how URLs are managed.

Stypox:
@Isira-Seneviratne from the IRC channel:
> I don't think we will merge neither Retrofit nor OkHttp in the extractor. Those are client side-libraries.

Isira-Seneviratne:
> @Isira-Seneviratne from the IRC channel:
> > I don't think we will merge neither Retrofit nor OkHttp in the extractor. Those are client side-libraries.

Thanks for the heads up.

Isira-Seneviratne:
> Cool concept with the new URL library. How does it feel with the malformed URL exceptions? I saw catch statements replaced to catch an argument exception instead of a malformed URL exceptions, so I wasn't quite sure how those are handled with the new library.

It rethrows the MalformedURLException as an IllegalArgumentException.

> 
> The biggest risk I see is that if we start adopting it, is it something that we'll replace the entire app with soon? Or will there be enough pushback back on other pull requests so that we end up with some places using the old URL class and others using the new class? If we get buy in by the owners of the project and can get it replaced relatively quickly and easily, it could be a good addition to how URLs are managed.

FireMasterK:
Maybe we should use an alternative URL parsing library? (Or, maybe we can create our own helper class)

I agree, okhttp is quite bloated if we just use it for parsing URLs

Isira-Seneviratne:
> Maybe we should use an alternative URL parsing library? (Or, maybe we can create our own helper class)
> 
> I agree, okhttp is quite bloated if we just use it for parsing URLs

I added a few changes involving the construction of the URLs called in the extractor.

PR #987, opened 2022-11-28
Use Description object for comments text.

FireMasterK:

See https://github.com/TeamNewPipe/NewPipeExtractor/pull/928#issuecomment-1328313586

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfoItemExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfoItemExtractor.java
@@ -5,6 +5,7 @@ import org.schabi.newpipe.extractor.Page;
 import org.schabi.newpipe.extractor.exceptions.ParsingException;
 import org.schabi.newpipe.extractor.localization.DateWrapper;
 import org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeCommentsInfoItemExtractor;
+import org.schabi.newpipe.extractor.stream.Description;
 import org.schabi.newpipe.extractor.stream.StreamExtractor;
 
 import javax.annotation.Nullable;
@@ -41,8 +42,8 @@ public interface CommentsInfoItemExtractor extends InfoItemExtractor {
     /**
      * The text of the comment
      */
-    default String getCommentText() throws ParsingException {
-        return "";
+    default Description getCommentText() throws ParsingException {
+        return Description.EMPTY_DESCRIPTION;
     }
 
     /**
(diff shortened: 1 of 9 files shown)


TobiGr (approved):
lgtm

extractor/src/main/java/org/schabi/newpipe/extractor/comments/CommentsInfoItemExtractor.java
```diff
@@ -41,8 +42,8 @@ default String getTextualLikeCount() throws ParsingException {
     /**
      * The text of the comment
      */
-    default String getCommentText() throws ParsingException {
-        return "";
+    default Description getCommentText() throws ParsingException {
+        return new Description("", Description.PLAIN_TEXT);
```
Stypox, line 46: Use `EMPTY_DESCRIPTION` here

extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubeCommentsInfoItemExtractor.java
```diff
@@ -59,13 +60,15 @@ public DateWrapper getUploadDate() throws ParsingException {
     }
 
     @Override
-    public String getCommentText() throws ParsingException {
+    public Description getCommentText() throws ParsingException {
         final String htmlText = JsonUtils.getString(item, "text");
```
Stypox, line 64: Since here the comment is returned as HTML, can't you just return `Description(htmlText, Description.HTML)`?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsInfoItemExtractor.java
```diff
@@ -176,18 +176,20 @@ public String getTextualLikeCount() throws ParsingException {
...
     @Override
-    public String getCommentText() throws ParsingException {
+    public Description getCommentText() throws ParsingException {
         try {
             final JsonObject contentText = JsonUtils.getObject(getCommentRenderer(), "contentText");
             if (contentText.isEmpty()) {
                 // completely empty comments as described in
                 // https://github.com/TeamNewPipe/NewPipeExtractor/issues/380#issuecomment-668808584
-                return "";
+                return new Description("", Description.PLAIN_TEXT);
```
Stypox, line 185: `EMPTY_DESCRIPTION` here, too

Stypox (changes requested):
Thanks!

extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubeCommentsInfoItemExtractor.java
FireMasterK: No, because it contains some junk like css classes, etc.

Stypox (approved):

PR #989, opened 2022-11-28
Block wrong nullable/nonnull imports in checkstyle

Stypox:

Same as https://github.com/TeamNewPipe/NewPipe/pull/9474, but here the Android annotation is forbidden and instead the `javax` is allowed. @TobiGr


AudricV:
Could you forbid the edu ones too or not?

Stypox:
Which ones are those? Android Studio wouldn't suggest them to me.

PR #990, opened 2022-11-28
[YouTube] Implement bold/italic/strike-through support

FireMasterK:

Todo:
- [x] Tests


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
```diff
@@ -940,16 +940,49 @@ public static String getTextFromObject(final JsonObject textObject, final boolea
...
+                if (italic) {
+                    textBuilder.append("</i>");
+                }
+                if (bold) {
+                    textBuilder.append("</b>");
+                }
+
+                continue;
             }
             textBuilder.append(text);
```
AudricV, line 987: Can't you use a `else` here instead of a `continue`?

AudricV (commented):
Except my question below, LGTM!

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
FireMasterK: That is true indeed!

Stypox (approved):

FireMasterK:
> This does not seem to work well for video descriptions. Calling `getDescription().getContent()` for [this stream](https://www.youtube.com/watch?v=XHegE9h0ZvA) returns this description (I omitted most of it in `[...]`):
> 
> ```
> [...] a daily message from me: &lt;a href="https://www.youtube.com/c/yogawithadriene/community"&gt;https://www.youtube.com/c/yogawithadr...&lt;/a&gt;<br><br>SIGN UP at &lt;a href="https://do.yogawithadriene.com/center"&gt;https://do.yogawithadriene.com/center&lt;/a&gt; to receive [...]
> ```
> 
> As you can see the `<br>` tags are correct, while the `<a>` tags have `&lt;` and `&gt;` instead of `<` and `>`. So HTML renderers interpret those as escaped characters instead of as tag delimiters.

I think this might be a side effect of #1005 instead :thinking:

TobiGr:
Escaping `text` right after `if (html) {` should do the job, i guess.

FireMasterK:
> Escaping `text` right after `if (html) {` should do the job, i guess.

We already have such a check, yet this happens. Will need to investigate further.

TobiGr:
that's not what i meant. see #1006

PR #992, opened 2022-11-30
Use String.isBlank().

Isira-Seneviratne:
----
Use `String`'s `isBlank()` method, introduced in Java 11.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
@@ -306,23 +306,8 @@ public final class Utils {
         return map == null || map.isEmpty();
     }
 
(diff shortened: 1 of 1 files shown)


FireMasterK (approved):

AudricV:
I think we merged this PR too fast... This method doesn't exist on Android until API 33 and is not available with desugarging!

Isira-Seneviratne:
> I think we merged this PR too fast... This method doesn't exist on Android until API 33 and is not available with desugarging!

I checked it on my end first, and the method gets desugared.

AudricV:
I didn't find anything about desugarging of this method. Do you have any source?

Isira-Seneviratne:
> I didn't find anything about desugarging of this method. Do you have any source?

I built and decompiled the app, and there was a separate class containing the method backport.

PR #941, opened 2022-10-09
[PeerTube]  Support comment replies

TobiGr:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).
- [x] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API.

Based on / Requires #936 
Related TeamNewPipe/NewPipe#9020 + TeamNewPipe/NewPipe#9410

Implement `getReplyCount()` and `getReplies()` for PeerTube.

PR #993, opened 2022-12-03
[SoundCloud] Add support for comment replies

TobiGr:
- [x] ~I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API.~  
Not needed. Is done in https://github.com/TeamNewPipe/NewPipe/pull/9410

Follow up on #936 and #941 

To Do:
- [ ] self review


extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudCommentsInfoItemExtractor.java
```diff
@@ -77,6 +143,54 @@ public String getUrl() {
...
+                    } else if (foundReply) {
+                        // Only the comments directly after the original comment
+                        // having the same timestamp are replies to the original comment.
+                        // The first comment not having the same timestamp
+                        // is the next top-level comment.
+                        break;
+                    }
+                }
+            }
+            replyCount = replies.size();
```
Stypox, line 178: Are you creating the `replies` array just to keep track of the count? Can't you just set `replyCount` to 0 at the beginning and also use it to replace the `replies` array and the `foundReply` boolean?

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudCommentsInfoItemExtractor.java
```diff
@@ -77,6 +143,54 @@ public String getUrl() {
...
+            if (collector.getItems().isEmpty()) {
```
Stypox, line 179: I guess you are collecting items here to make sure not to show that there are replies when actually all replies give an error when collected.
- Since you are already collecting replies here, can't you set the items as the `repliesPage` content directly instead of passing the `json` again? Or maybe make a `Serializable` structure of some sort. To avoid recollecting them again later.
- Or even better, I would skip this check completely and avoid collecting items beforehand. If one of the comments yields an error while being collected, then the error would be shown in the frontend and it is expected that in case of error on an item the list of replies might be empty.

Stypox (changes requested):
Looks mostly good to me, but it looks like at SoundCloud they overcomplicated comment replies :sweat_smile:

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudCommentsInfoItemExtractor.java
TobiGr: Good catch. I used the `replies` List to retrieve other info previously. It's not needed anymore.

PR #996, opened 2022-12-04
[PeerTube] Support searching for playlists and channels

TobiGr:

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/PeertubeParsingHelper.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/PeertubeParsingHelper.java
@@ -111,14 +111,12 @@ public final class PeertubeParsingHelper {
                 final InfoItemExtractor extractor;
                 if (sepia) {
                     extractor = new PeertubeSepiaStreamInfoItemExtractor(item, baseUrl);
+                } else if (isPlaylistInfoItem) {
+                    extractor = new PeertubePlaylistInfoItemExtractor(item, baseUrl);
+                } else if (isChannelInfoItem) {
+                    extractor = new PeertubeChannelInfoItemExtractor(item, baseUrl);
                 } else {
-                    if (isPlaylistInfoItem) {
-                        extractor = new PeertubePlaylistInfoItemExtractor(item, baseUrl);
-                    } else if (isChannelInfoItem) {
-                        extractor = new PeertubeChannelInfoItemExtractor(item, baseUrl);
-                    } else {
-                        extractor = new PeertubeStreamInfoItemExtractor(item, baseUrl);
-                    }
+                    extractor = new PeertubeStreamInfoItemExtractor(item, baseUrl);
                 }
                 collector.commit(extractor);
             }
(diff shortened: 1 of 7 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/PeertubeParsingHelper.java
```diff
@@ -101,12 +105,20 @@ public static void collectStreamsFrom(final InfoItemsCollector collector,
...
                 } else {
-                    extractor = new PeertubeStreamInfoItemExtractor(item, baseUrl);
+                    if (isPlaylistInfoItem) {
+                        extractor = new PeertubePlaylistInfoItemExtractor(item, baseUrl);
+                    } else if (isChannelInfoItem) {
+                        extractor = new PeertubeChannelInfoItemExtractor(item, baseUrl);
+                    } else {
+                        extractor = new PeertubeStreamInfoItemExtractor(item, baseUrl);
+                    }
                 }
```
Stypox, line 122: No need to nest `if`s again, just do `else if (isPlaylistInfoItem)` directly.

extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubeChannelInfoItemExtractor.java
```diff
@@ -0,0 +1,70 @@
...
+import com.grack.nanojson.JsonArray;
+import com.grack.nanojson.JsonObject;
+
+import org.schabi.newpipe.extractor.channel.ChannelExtractor;
+import org.schabi.newpipe.extractor.channel.ChannelInfoItemExtractor;
+import org.schabi.newpipe.extractor.exceptions.ParsingException;
+
+import javax.annotation.Nonnull;
+
+public class PeertubeChannelInfoItemExtractor implements ChannelInfoItemExtractor {
```
Stypox, line 12: ~~Isn't it possible that PeerTube channels have parent channels?~~ Yes but they are not shown in search results even by Peertube websites, so ignore this comment.

extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubeChannelInfoItemExtractor.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+        int highestRes = -1;
+        JsonObject avatar = null;
+        for (final Object a: avatars) {
+            if (((JsonObject) a).getInt("width") > highestRes) {
+                avatar = (JsonObject) a;
+                highestRes = avatar.getInt("width");
+            }
+        }
+        return baseUrl + avatar.getString("path");
```
Stypox, line 48:
```suggestion
        return item.getArray("avatars").stream()
                .filter(object -> object instanceof JsonObject)
                .map(object -> (JsonObject) object)
                .max(Comparator.comparingInt(avatar -> avatar.getInt("width")))
                .map(avatar -> baseUrl + avatar.getString("path"))
                .orElse(null);
```

Stypox (changes requested):
Looks almost good to me, thanks :-)

extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubeChannelInfoItemExtractor.java
TobiGr: That looks way easier. For some reason, I don't think of using streams...

Stypox (approved):
Thank you! The CI fail is unrelated

PR #997, opened 2022-12-05
[YouTube] Add support for extracting auto-translated captions

### LINKED ISSUE #977: Extract automatically translated subtitles
It's some time since YouTube began supporting automatic translation for subtitles into variety of languages. It works for every video that has automatically generated subtitles.

Currently NewPipe only supports automatically generated subtitles in English and of course manually embedded subtitles, but nothing more AFAIK and so I presume it's extraction is not supported by the NewPipe Extractor.
I think it would be cool and could help a lot of users if all these subtitles would be extracted by NewPipe Extractor and would be made available in NewPipe.

The only issue I am thinking about is if these automatically translated subtitles are hosted in the same format on YouTube servers as the original ones or if they are translated in real-time. If that's the case it would make implementation probably much harder.

[Here](https://youtu.be/lADBHDg-JtA)'s an example of YouTube video that has automatically translated subtitles. (You will first need to enable English subtitles for the automatic translation option to show up.)

TobiGr:

Extract auto-translated captions for YouTube videos.

### `SubtitlesStream`
This adds `isAutoTranslated()` next to `isAutoGenerated()` to distinguish between auto-generated subtitles which use speech2text and auto-translated captions based on Google translator.
Additionally, `getBaseLocale()`, `getDisplayBaseLanguageName()` and `getBaseLanguageTag()` were added to access info on the language which was used for auto-translations.

## Issues closed by this PR

Closes TeamNewPipe/NewPipeExtractor#977
Based on and adresses TeamNewPipe/NewPipe#8023


extractor/src/main/java/org/schabi/newpipe/extractor/stream/SubtitlesStream.java
```diff
@@ -150,6 +152,18 @@ public Builder setAutoGenerated(final boolean autoGenerated) {
             return this;
         }
 
+        /**
+         * Set whether the subtitles have been automatically
+         * (i.e. by a machine like Google Translator) translated by the streaming service.
```
AudricV, line 157:
```suggestion
         * Set whether the subtitles have been automatically translated
         * (i.e. by a machine like Google Translator) by the streaming service.
         *
```

extractor/src/main/java/org/schabi/newpipe/extractor/stream/SubtitlesStream.java
```diff
@@ -217,6 +237,7 @@ public SubtitlesStream build() {
      * @param deliveryMethod the {@link DeliveryMethod} of the stream
      * @param languageCode   the language code of the stream
      * @param autoGenerated  whether the subtitles are auto-generated by the streaming service
+     * @param autoTranslated
```
AudricV, line 240: Parameter description is missing.

extractor/src/main/java/org/schabi/newpipe/extractor/stream/SubtitlesStream.java
```diff
@@ -276,6 +299,19 @@ public boolean isAutoGenerated() {
...
     }
 
+    /**
+     * Whether the subtitles are translated automatically by a machine.
+     * <p>
+     * Some streaming services provide automatically translated subtitles.
+     * YouTube, for example, uses Google translator to generate translated subtitles.
+     * Automatically translated subtitles might not coincide completely with the original text.
+     * </p>
+     * @return {code true} if the subtitles are auto-translated, {@link false} otherwise
```
AudricV, line 309:
```suggestion
     *
     * <p>
     * Some streaming services provide automatically translated subtitles.
     * YouTube, for example, uses Google Translator to generate translated subtitles.
     * Automatically translated subtitles might not coincide completely with the original text.
     * </p>
     *
     * @return {code true} if the subtitles are auto-translated, {@link false} otherwise
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -717,13 +716,32 @@ public List<SubtitlesStream> getSubtitles(final MediaFormat format) throws Parsi
                         .setMediaFormat(format)
                         .setLanguageCode(languageCode)
                         .setAutoGenerated(isAutoGenerated)
+                        .setAutoTranslated(false)
                         .build());
+                if (i == 0 && caption.getBoolean("isTranslatable")
```
AudricV, line 721: I would not base the extraction on the index, but rather on whether the subtitles are auto-generated:

```suggestion
                if (isAutoGenerated && caption.getBoolean("isTranslatable")
```

Also, this PR doesn't add support of subtitles translation for uploaded subtitles. For instance, see https://www.youtube.com/watch?v=_cMxraX_5RE: you can translate from German to French and from English to French, and the translations are different.

We may need another property in `SubtitlesStream` for this.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -717,13 +716,32 @@ public List<SubtitlesStream> getSubtitles(final MediaFormat format) throws Parsi
...
+                                .setContent(cleanUrl + "&fmt=" + format.getSuffix()
+                                        + "&tlang=" + tLanguageCode, true)
+                                .setMediaFormat(format)
+                                .setLanguageCode(tLanguageCode)
+                                .setAutoGenerated(isAutoGenerated)
+                                .setAutoTranslated(true)
+                                .build());
+                    }
+                }
+
```
AudricV, line 737:
```suggestion
```

AudricV (changes requested):

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
TobiGr: I don't understand why we should use `isAutoGenerated` here. For better quality, it should be `!isAutoGenerated`. Manually added captions should be exact.
I was also wondering whether we should provide the auto-translated captions by default. Extracting the data for and generating ~100 SubtitleStreams takes some time. I'd definitely not recommend to do this for all available languages by default. On the other hand, we could provide a method which does this when needed.

Facni:
What happened to this?

Stypox:
Is something blocking this PR from getting merged (except the needed rebase)? @AudricV

(some user asked via email about this feature)

PR #1000, opened 2022-12-08
[YouTube] Improve YoutubeStreamInfoItemExtractor

AudricV:
- [x] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe). Debug APK to test (contains an app change to return the comment text as a string, not as a `Description`, in order to avoid the corresponding build error): [app-debug.zip](https://github.com/TeamNewPipe/NewPipeExtractor/files/10185647/app-debug.zip)

This PR improves some code in `YoutubeStreamInfoItemExtractor` and:
- makes duration of video premieres available
- returns view count and upload date of streams in playlists
- uses another non-localized method to determine whether a stream is a running livestream (as a last resort)
- removes the YT Shorts workaround code, as it was only useful on channels, and Shorts have been moved into a separated channel tab.

As a side effect, if duration of YouTube stream items cannot be extracted, `-1` is always returned.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
@@ -10,6 +10,7 @@ import org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeStreamLi
 import org.schabi.newpipe.extractor.stream.StreamInfoItemExtractor;
 import org.schabi.newpipe.extractor.stream.StreamType;
 import org.schabi.newpipe.extractor.utils.JsonUtils;
+import org.schabi.newpipe.extractor.utils.Parser;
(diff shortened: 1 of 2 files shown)


opusforlife2:
1000! 🎉

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
```diff
@@ -116,30 +126,44 @@ public String getName() throws ParsingException {
...
+                            timeOverlay.getObject("thumbnailOverlayTimeStatusRenderer")
+                                    .getObject("text"));
                 }
             }
 
             if (isNullOrEmpty(duration)) {
+                if (isPremiere()) {
+                    // Premieres can be livestreams, so the duration is not available in this
+                    // case
+                    return -1;
```
TobiGr, line 160: We really need to introduce more constants for those default values in the `Info` classes. But that's something for another PR, maybe a _good first issue_.

PR #1001, opened 2022-12-10
Implement game channels

john-peterson:
They are useful as a twitch alternative. that gigantic data hog. that still hasnt fixed background play sufficiently. tihs recent tab even show a few low view count videos. much appreciated as i cant stand the constantn yapping or engagenemnt spin machines. even live styreams are still there. still mourn the loss of the gaming app

btw i have disabled all content downloads so its possible that some of them are broken if you show all horrendous thumbs and everything uncesseseray


FireMasterK:
I feel this could be better implemented in #951 since your change essentially adds support for a "recent" tab, which could be added there instead (along with support for other tabs if necessary).

Stypox:
@john-peterson yes I agree with FireMasterK. It's better to wait for #951 to be merged and then take care of this. Thanks!

PR #1013, opened 2023-01-15
[YouTube] Now music mixes can be treated as normal mixes

Stypox:

The fix was simpler than expected. A while ago, apparently, YouTube treated music playlists as just special playlists, and so a normal `YouTubePlaylistExtractor` was used to fetch them. Now, instead, when using that method you get "This playlist type is unviewable.". The solution was just to remove the distinction between other mixes and music mixes: now they are all treated the same way using `YoutubeMixPlaylistExtractor`. By taking a look at the browser's network tab I could confirm this is the right approach.

The first commit just changes 4 lines of code. The second one adds a mix test along with mocks.

Fixes TeamNewPipe/NewPipe#9536

Testing APK: [app-debug.zip](https://github.com/TeamNewPipe/NewPipeExtractor/files/10421431/app-debug.zip)


TobiGr (approved):
LGTM

PR #1015, opened 2023-01-17
[YouTube] Fix channel ID extraction of YouTube channels RSS feeds

AudricV:

In RSS channels feeds, the `yt:channelId` element doesn't provide the channel ID anymore (it has no value), like the `id` element (only the `yt:channel:` prefix is provided). This currently makes `getId()` method of `YoutubeFeedExtractor` returning `null`, and so fails the method `testId` of `YoutubeFeedExtractorTest.Kurzgesagt` test class.

We need now to extract it from the channel URL provided in two elements: `author` -> `uri` and `feed` -> `link`.

This PR also avoids `NullPointerException`s which could happen in `getUrl` and `getName` methods of `YoutubeFeedExtractor`.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeFeedExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeFeedExtractor.java
@@ -22,6 +22,8 @@ import java.io.IOException;
 import javax.annotation.Nonnull;
 
 public class YoutubeFeedExtractor extends FeedExtractor {
+    private static final String WEBSITE_CHANNEL_BASE_URL = "https://www.youtube.com/channel/";
+
     public YoutubeFeedExtractor(final StreamingService service, final ListLinkHandler linkHandler) {
         super(service, linkHandler);
     }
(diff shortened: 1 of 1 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeFeedExtractor.java
```diff
@@ -57,19 +61,44 @@ public ListExtractor.InfoItemsPage<StreamInfoItem> getInitialPage() {
...
+        final String idFromAuthorUriElement = getUrl().replace(WEBSITE_CHANNEL_BASE_URL, "");
+        if (isNullOrEmpty(idFromAuthorUriElement)) {
+            final Element linkElement = document.select("feed > link[rel*=alternate]")
+                    .first();
+            if (linkElement != null) {
+                final String linkElementChannelId = linkElement.attr("href")
+                        .replace(WEBSITE_CHANNEL_BASE_URL, "");
+                if (!isNullOrEmpty(linkElementChannelId)) {
+                    return linkElementChannelId;
+                }
```
Stypox, line 73: This will have the same effect I think, since if linkElementChannelId were null or empty then control would reach the other return which would return null anyway.
```suggestion
                return linkElement.attr("href")
.replace(WEBSITE_CHANNEL_BASE_URL, "");
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeFeedExtractor.java
```diff
@@ -57,19 +61,44 @@ public ListExtractor.InfoItemsPage<StreamInfoItem> getInitialPage() {
     @Nonnull
```
Stypox, line 61: Mmmh, this method is marked as Nonnull, but are nulls turned into empty strings?

Stypox (approved):
Lgtm other than the two small details. Thank you!

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeFeedExtractor.java
AudricV: I just noticed that this could be used as a fallback for the channel URL, so I am moving this part into the `getUrl` method.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeFeedExtractor.java
AudricV: If we couldn't extract the URL, an empty string is provided, nulls are just handled to avoid `NullPointerException`s if there is no element found in the path we provide to jsoup (the [`first` method of the `Elements` class returns `null`](https://jsoup.org/apidocs/org/jsoup/select/Elements.html#first()) in this case).

Stypox (approved):

PR #1021, opened 2023-01-24
[YouTube] Fix getting the comment text if the comment contains a hashtag

### LINKED ISSUE #1019: [YouTube] ParsingException for hashtags links in descriptions and comments
See https://github.com/TeamPiped/Piped/issues/1967

`org.schabi.newpipe.extractor.exceptions.ParsingException: canonicalBaseUrl is null and browseId is not a channel ("{browseId=FEhashtag, params=6gUJCgdidW5uaWVz}")`

TobiGr:

This fix needs further investigation.
The current approach is very strict and throws a ParsingException if the `webCommandMetadata` is present, but no URL field is given. I was not able to find any other occurrences of that field, but that does not mean there aren't any (I don't know much about the `navigationEndpoint`, maybe somebody can help us out here). A possibly better solution is to put the condition before the `if () else if ...` block as shown below to prevent any unintentional exceptions.
``` Java
@Nullable
public static String getUrlFromNavigationEndpoint(@Nonnull final JsonObject navigationEndpoint)
        throws ParsingException {
   if (navigationEndpoint.has("webCommandMetadata")) {
        // this case needs to be handled before the browseEndpoint,
        // e.g. for hashtags in comments
        final JsonObject metadata = navigationEndpoint.getObject("webCommandMetadata");
        if (metadata.has("url")) {
            return "https://www.youtube.com" + metadata.getString("url");
        }
    }
    if (navigationEndpoint.has("urlEndpoint")) {
        // ...
    } else if (navigationEndpoint.has("browseEndpoint")) {
    // ...
```
Fixes #1019

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
@@ -845,6 +845,14 @@ public final class YoutubeParsingHelper {
     @Nullable
     public static String getUrlFromNavigationEndpoint(@Nonnull final JsonObject navigationEndpoint)
             throws ParsingException {
+        if (navigationEndpoint.has("webCommandMetadata")) {
+            // this case needs to be handled before the browseEndpoint,
+            // e.g. for hashtags in comments
+            final JsonObject metadata = navigationEndpoint.getObject("webCommandMetadata");
+            if (metadata.has("url")) {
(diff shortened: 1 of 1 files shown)


Stypox:
I think the solution you propose in the PR body is better. In the end, I don't think anything could go wrong if `webCommandMetadata` is actually supposed to appear when the other two aren't there. But if this wasn't actually the case, we would be more future-proof.

Stypox (approved):

PR #1022, opened 2023-01-28
[YouTube] Support live URLs

AudricV:
- [ ] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe). I didn't tested in the app but added tests in the extractor. As `LinkHandler` tests still pass with my non-breaking API changes, everything should work fine in the app.
- [x] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API. This pull request will only require support for opening YouTube's `/live/` links from external apps, as the extractor update will be done in a separate PR.

This pull request adds support for YouTube's live URLs, which are in the form `https://www.youtube.com/live/LIVE_ID`, where `LIVE_ID` is the video ID of the livestream. This behavior applies to past, planned and running livestreams.

Note that these URLs are then transformed to `watch` URLs in the address bar with the desktop client once the watch page has been loaded.

I also added a test for these live URLs and did minor improvements to the files changed (see the commits for details).

This pull request is the extractor part of TeamNewPipe/NewPipe#9546.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeStreamLinkHandlerFactory.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeStreamLinkHandlerFactory.java
@@ -1,3 +1,23 @@
+/*
+ * Created by Christian Schabesberger on 02.02.16.
+ *
+ * Copyright (C) Christian Schabesberger 2018 <chris.schabesberger@mailbox.org>
+ * YoutubeStreamLinkHandlerFactory.java is part of NewPipe Extractor.
+ *
+ * NewPipe Extractor is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * NewPipe Extractor is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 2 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeStreamLinkHandlerFactory.java
```diff
@@ -15,41 +35,21 @@
...
 public final class YoutubeStreamLinkHandlerFactory extends LinkHandlerFactory {
 
     private static final Pattern YOUTUBE_VIDEO_ID_REGEX_PATTERN
             = Pattern.compile("^([a-zA-Z0-9_-]{11})");
     private static final YoutubeStreamLinkHandlerFactory INSTANCE
             = new YoutubeStreamLinkHandlerFactory();
-    private static final List<String> SUBPATHS
-            = Arrays.asList("embed/", "shorts/", "watch/", "v/", "w/");
+    private static final Set<String> SUBPATHS
+            = Set.of("embed/", "live/", "shorts/", "watch/", "v/", "w/");
```
TobiGr, line 52: You can also create unmodifiable lists using the same method in `List`. What the reasoning for switching to a set?

TobiGr (commented):
LGTM, just a small question

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeStreamLinkHandlerFactory.java
AudricV: I thought that using sets would be better, but it doesn't seem to be the case for our usage. I changed `SUBPATHS` type back to `List`.

TobiGr (approved):

PR #1026, opened 2023-02-12
Add descriptive and locale properties to audio streams

AudricV:
- [ ] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API. No breaking changes.

This PR adds support to extract descriptive audio support property in `AudioStream` (only implemented on YouTube, with a test added), which are currently not distinguished with main audio tracks.

It also adds the ability to get a `Locale` for audio languages, which should make easier for clients to get a language object, and so avoid them to parse the audio track id (which may be not reliable in some services, such as MediaCCC where there is no audio track ID but only a language tag) to get a `Locale`. Support and tests have been added for YouTube (streams and DASH manifests creators) and MediaCCC (other services doesn't seem to support multiple audio tracks, as far I know).

A few documentation improvements and fixes (in the code and in the documentation) around the changes done have been also made.

This PR is required to add an optimal multiple audio tracks support in NewPipe.


Stypox (changes requested):
Great! Code looks good to me.

Could you add some javadoc explanation about what descriptive audios are?

AudricV:
> Could you add some javadoc explanation about what descriptive audios are?

Done

extractor/src/main/java/org/schabi/newpipe/extractor/stream/AudioStream.java
```diff
@@ -421,15 +485,36 @@ public String getAudioTrackId() {
...
+     *
+     * @return the {@link Locale} of the audio or {@code null}
+     */
+    @Nullable
+    public Locale getAudioLocale() {
+        return audioLocale;
+    }
+
+    /**
+     * Returns whether this stream is a descriptive audio.
```
Stypox, line 510: Since this is the API accessible from library users, I think it's best to document it here, too.
(also, no need to write "Returns ..." and then "@return ..." twice)

```suggestion
     * A descriptive audio is an audio in which descriptions of visual elements of a video are
     * added in the original audio, with the goal to make a video more accessible to blind and
     * visually impaired people.
     *
```

Stypox (changes requested):

extractor/src/main/java/org/schabi/newpipe/extractor/stream/AudioStream.java
AudricV: > also, no need to write "Returns ..." and then `@return ...` twice

I agree if there is only the "return" text. Now that I've added the description in the Javadoc for the getter, I think starting the doc with the description would be strange, so I kept it.

I also changed the `return` Javadoc description to give an explicit return value.

Stypox (approved):

PR #1029, opened 2023-02-21
[YouTube] Fix partial non-extraction of "No views" string in stream items

AudricV:

As the "No views" string is returned in the case there is no view on a YouTube video, a number cannot be parsed in this case, so -1 was returned, as the detection of this string was only made for the first value (`viewCountText`).

This string is now detected in all methods to get the view count of a stream item and the corresponding code has been refactored to reduce code duplication as much as possible.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
@@ -18,6 +18,7 @@ import java.time.Instant;
 import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
+import java.util.regex.Pattern;
 
 import static org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper.getTextFromObject;
 import static org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper.getThumbnailUrlFromInfoItem;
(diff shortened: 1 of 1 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
```diff
@@ -284,67 +292,86 @@ public long getViewCount() throws ParsingException {
...
 
         // No view count extracted: return -1, as the view count can be hidden by creators on videos
         return -1;
     }
 
+    private long getViewCountFromViewCountText(@Nonnull final String viewCountText,
+                                               final boolean isMixedNumber)
+            throws NumberFormatException, ParsingException {
+        // These approaches are language dependent
+        if (viewCountText.toLowerCase().contains("no views")) {
```
Stypox, line 347: Why didn't you use `ACCESSIBILITY_DATA_NO_VIEWS_REGEX_LOWERCASE` here, too? Or why didn't you use `.contains("no views")` below, too?

AudricV: I didn't use the regex here because the view count text should only contain the `No views` string and `contains` checks should be faster than using a regex.

For the title accessibility data parsing, we can't use this approach, because if a video title contains `no views`, the extractor will think that there is no views on this video, while it may be not the case.

Stypox: Can't you just do `.endsWith("no views")`?

Stypox (approved):

PR #1030, opened 2023-02-23
[Bandcamp] Support loading additional comments

### LINKED ISSUE #621: Load additional Bandcamp comments
Follow-up to #586 and #620.

Currently, we are only loading the first batch of comments (max. 3).

Example of an album where multiple pages of comments are available: https://alfamist.bandcamp.com/album/bring-backs

petlyh:
No NewPipe changes required.

The Bandcamp service now supports loading multiple pages of comments.
Closes #621 

Initial comments are now extracted from the `data-blob` attribute of the `#collectors-data` element, with subsequent pages of comments being fetched from the API. The way initial comments are being extracted was changed to:
1. have the first page include _all_ of the initial comments, not just the ones shown.
2. be able to use the same `CommentsInfoItemExtractor` for both the initial page and subsequent pages.

Note: I was unable to find a track with more than 2 pages of comments to test.

Before:

After:

https://user-images.githubusercontent.com/88139840/220970548-b3070865-d036-4312-a436-96a46d3faaf2.mp4

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampCommentsExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampCommentsExtractor.java
@@ -1,9 +1,13 @@
 package org.schabi.newpipe.extractor.services.bandcamp.extractors;
 
+import static org.schabi.newpipe.extractor.services.bandcamp.extractors.BandcampExtractorHelper.BASE_API_URL;
+
+import com.grack.nanojson.JsonArray;
+import com.grack.nanojson.JsonObject;
+import com.grack.nanojson.JsonWriter;
+
 import org.jsoup.Jsoup;
 import org.jsoup.nodes.Document;
-import org.jsoup.nodes.Element;
-import org.jsoup.select.Elements;
 import org.schabi.newpipe.extractor.Page;
 import org.schabi.newpipe.extractor.StreamingService;
 import org.schabi.newpipe.extractor.comments.CommentsExtractor;
@@ -11,12 +15,20 @@ import org.schabi.newpipe.extractor.comments.CommentsInfoItem;
 import org.schabi.newpipe.extractor.comments.CommentsInfoItemsCollector;
 import org.schabi.newpipe.extractor.downloader.Downloader;
 import org.schabi.newpipe.extractor.exceptions.ExtractionException;
+import org.schabi.newpipe.extractor.exceptions.ParsingException;
+import org.schabi.newpipe.extractor.exceptions.ReCaptchaException;
 import org.schabi.newpipe.extractor.linkhandler.ListLinkHandler;
+import org.schabi.newpipe.extractor.utils.JsonUtils;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.Collections;
 
 import javax.annotation.Nonnull;
-import java.io.IOException;
 
 public class BandcampCommentsExtractor extends CommentsExtractor {
+
+    private static final String REVIEWS_API_URL = BASE_API_URL + "/tralbumcollectors/2/reviews";
 
     private Document document;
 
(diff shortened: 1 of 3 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampCommentsExtractor.java
```diff
@@ -39,18 +51,77 @@ public InfoItemsPage<CommentsInfoItem> getInitialPage()
...
         }
 
-        return new InfoItemsPage<>(collector, null);
+        if (!collectorsData.getBoolean("more_reviews_available")) {
+            return new InfoItemsPage<>(collector, null);
+        }
+
+        final String trackId = getTrackId();
+        final String token = getNextPageToken(reviews);
+        return new InfoItemsPage<>(collector, new Page(trackId, token));
```
AudricV, line 69: You should use the constructor with a `List` containing `trackId` and `token` values: `trackId` is not a URL, but an ID.

```suggestion
        return new InfoItemsPage<>(collector, new Page(List.of(trackId, token)));
```

(Don't forget to import the `List` class)

extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampCommentsExtractor.java
```diff
@@ -39,18 +51,77 @@ public InfoItemsPage<CommentsInfoItem> getInitialPage()
...
 
     @Override
     public InfoItemsPage<CommentsInfoItem> getPage(final Page page)
             throws IOException, ExtractionException {
-        return null;
+
+        final CommentsInfoItemsCollector collector = new CommentsInfoItemsCollector(getServiceId());
+
+        final String trackId = page.getUrl();
+        final String token = page.getId();
```
AudricV, line 79:
```suggestion
        final List<String> pageIds = page.getIds();
        final String trackId = pageIds.get(0);
        final String token = pageIds.get(1);
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampCommentsExtractor.java
```diff
@@ -39,18 +51,77 @@ public InfoItemsPage<CommentsInfoItem> getInitialPage()
...
+            collector.commit(
+                    new BandcampCommentsInfoItemExtractor((JsonObject) review, getUrl()));
+        }
+
+        if (!reviewsData.getBoolean("more_available")) {
+            return new InfoItemsPage<>(collector, null);
+        }
+
+        final String nextToken = getNextPageToken(reviews);
+        return new InfoItemsPage<>(collector, new Page(trackId, nextToken));
```
AudricV, line 93: No need to use an intermediate variable here:

```suggestion
        return new InfoItemsPage<>(collector,
                new Page(trackId, getNextPageToken(reviews)));
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampCommentsExtractor.java
```diff
@@ -39,18 +51,77 @@ public InfoItemsPage<CommentsInfoItem> getInitialPage()
...
+                            .value("count", 7)
+                            .array("exclude_fan_ids").end()
+                            .end().done().getBytes(StandardCharsets.UTF_8)).responseBody());
+        } catch (final IOException | ReCaptchaException e) {
+            throw new ParsingException("Could not fetch reviews", e);
+        }
+    }
+
+    private String getNextPageToken(final JsonArray reviews) throws ParsingException {
+        return reviews.stream().map(JsonObject.class::cast)
```
AudricV, line 115: You should ensure the type of the object before casting to avoid `ClassCastException`s (even if it should not happen in this case):

```suggestion
        return reviews.stream()
                .filter(JsonObject.class::isInstance)
                .map(JsonObject.class::cast)
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampCommentsExtractor.java
```diff
@@ -39,18 +51,77 @@ public InfoItemsPage<CommentsInfoItem> getInitialPage()
...
+                .map(o -> o.getString("token"))
+                .reduce((a, b) -> b)
```
AudricV, line 117: Please use explicit variable names, `o` should be `review`, but I don't know what is `a` and `b`.

AudricV (changes requested):
Thank you very much for this PR!

Please apply the following changes and add `@Nonnull` and `@Nullable` annotations where that's relevant in the code you added and changed.

You will also have to wait a review from @fynngodau.

The Bandcamp tests failures from the CI doesn't seem to be related to your changes (timeouts).

fynngodau:
Hello @petlyh, many thanks for opening this PR. I'm happy to see that you have taken a look at this old Bandcamp issue!

Finding a _track_, not an album, with many comments is indeed quite tricky! I found this one, with three pages (one initial + two): https://laurashigihara.bandcamp.com/track/cube-land

extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampCommentsExtractor.java
fynngodau: Since this line gets the last item, `a` and `b` are probably just supposed to be `first` and `second` like in https://stackoverflow.com/a/21441634/20044811?

Stypox: I think it's fine to have `a` and `b`, but a comment should be added to explain that `.reduce((a, b) -> b)` is just keeping the last element.
```suggestion
                .map(review -> review.getString("token"))
                .reduce((a, b) -> b) // keep only the last element
```

Stypox (changes requested):
Code looks good to me, except for the things already pointed out. I don't know anything about Bandcamp though.

Stypox (approved):

AudricV (approved):

fynngodau (approved):
Thabk you for applying the review comments, and many thanks from my side yet again!

PR #1032, opened 2023-02-24
[YouTube] Fix hashtags links extraction and escape HTML links

### LINKED ISSUE #1019: [YouTube] ParsingException for hashtags links in descriptions and comments

AudricV:

`webCommandMetadata` object is contained inside a `commandMetadata` one, so it is not accessible from the root of the `navigationEndpoint` object.

The corresponding statement has been moved at the bottom of the specific endpoints parsing, as the `webCommandMetadata` object is present almost everywhere, otherwise URLs of some endpoints would have be changed, such as uploader URLs (from channel IDs to handles).

As no `ParsingException` is now thrown by `getUrlFromNavigationEndpoint`, and so by `getTextFromObject`, `getUrlFromObject` and `getTextAtKey`, the methods which were catching `ParsingExceptions` thrown by these methods had to be updated.

URLs got in the HTML version of `getTextFromObject` are now escaped properly to provide valid HTML to clients. This has been also done for attribute descriptions, with the description text for this type of descriptions.

As YouTube descriptions are in HTML format (except for the fallback on the JSON player response, which is plain text and only happens when there is no visual metadata or a breaking change), URLs returned are escaped, so tests which are testing presence of URLs with escaped characters had to be updated (it was only the case for `YoutubeStreamExtractorDefaultTest.DescriptionTestUnboxing`).

I've also updated the mocks of two tests classes of `YoutubeCommentsExtractorTest` which were missing, in order to test completely my changes: `RepliesTest` and `FormattingTest`.

Fixes #1019 (for real this time)
Related issue: TeamNewPipe/NewPipe#9774


Stypox (approved):
Thank you for the thorough PR description. Code looks good to me. Thanks!

PR #1033, opened 2023-02-28
[Bandcamp] Handle paywalled tracks

petlyh:
No NewPipe changes required.

Tracks that cannot be accessed without being purchased now show an appropriate error.

Fixes TeamNewPipe/NewPipe#8634

|Before|After|
|-|-|
|||

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampStreamExtractor.java
@@ -15,6 +15,7 @@ import org.schabi.newpipe.extractor.MediaFormat;
 import org.schabi.newpipe.extractor.StreamingService;
 import org.schabi.newpipe.extractor.downloader.Downloader;
 import org.schabi.newpipe.extractor.exceptions.ExtractionException;
+import org.schabi.newpipe.extractor.exceptions.PaidContentException;
 import org.schabi.newpipe.extractor.exceptions.ParsingException;
 import org.schabi.newpipe.extractor.linkhandler.LinkHandler;
 import org.schabi.newpipe.extractor.localization.DateWrapper;
@@ -56,6 +57,10 @@ public class BandcampStreamExtractor extends StreamExtractor {
         if (albumJson.getArray("trackinfo").size() > 1) {
             // In this case, we are actually viewing an album page!
             throw new ExtractionException("Page is actually an album, not a track");
+        }
+
+        if (albumJson.getArray("trackinfo").getObject(0).isNull("file")) {
+            throw new PaidContentException("This track is not available without being purchased");
         }
     }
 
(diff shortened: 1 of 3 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampStreamExtractor.java
```diff
@@ -57,6 +58,10 @@ public void onFetchPage(@Nonnull final Downloader downloader)
             // In this case, we are actually viewing an album page!
             throw new ExtractionException("Page is actually an album, not a track");
         }
+
+        if (albumJson.getArray("trackinfo").getObject(0).isNull("file")) {
```
Stypox, line 62: Are you sure that all tracks that have no streamable track are paid tracks? Can't the reason be different in some cases?

fynngodau: The other possible reason is that it is not available yet because the album is currently up for pre-order. For example, this album is right now: https://tamarmitchell.bandcamp.com/album/samos-return

However it is impossible to see whether all of the tracks can be played for free after the album is released, or whether the tracks will require purchase to play. Therefore I think it is an okay option to throw this exception in either case.

fynngodau: I guess otherwise we could test for `unreleased_track` and throw a different exception if that is true. Is there an appropriate exception for this case? Otherwise I would stick with `PaidContentException`.

fynngodau:
Hi again @petlyh, thanks for opening another Bandcamp PR!

One suggestion: because I didn't know about the `PaidContentException` (or possibly it didn't exist at the time), an incorrect `ContentNotAvailableException` is thrown in the album code if an album needs to be paid. Could you change that to also throw `PaidContentException`?

Besides this, I'm very happy to see another Bandcamp fix from you. If you make another Bandcamp PR in the future, feel free to ping me directly so that I can have a look quicker :)

extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/extractors/BandcampStreamExtractor.java
petlyh: Personally, I think `ContentNotAvailableException` is more appropriate, since the track is currently unavailable until the album is released, whether it's purchased or not.

However, it seems like the tracks that can't be played due to the album being in pre-order do not have their own pages yet. Therefore, I don't think it's a case that can occur in the `StreamExtractor`. Going to the album in NewPipe and clicking on one of those tracks causes an error because the URL for them gets set to `https://tamarmitchell.bandcamp.com/null`, so maybe it should be handled differently in the `PlaylistExtractor`?

fynngodau: I see. I think I noticed the concatenation with `null` before but was unsure about the best way to fix it.

fynngodau (approved):
Thanks (and sorry for slow feedback loop, I didn't notice you pushing a commit somehow :/ )!

PR #1040, opened 2023-03-14
[YouTube] Fix ParsingException when comments are unavailable in a video

### LINKED ISSUE #969: [YouTube] ParsingException thrown when comments are unavailable in a video
Example video: https://piped.kavin.rocks/watch?v=92d75BMvevs
Test with PipedAPI: https://pipedapi.kavin.rocks/comments/92d75BMvevs

```
org.schabi.newpipe.extractor.exceptions.ParsingException: Unable to get contents.twoColumnWatchNextResults.results.results.contents
    at org.schabi.newpipe.extractor.utils.JsonUtils.getValue(JsonUtils.java:36)
    at org.schabi.newpipe.extractor.utils.JsonUtils.getInstanceOf(JsonUtils.java:44)
    at org.schabi.newpipe.extractor.utils.JsonUtils.getArray(JsonUtils.java:80)
    at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeCommentsExtractor.findInitialCommentsToken(YoutubeCommentsExtractor.java:83)
    at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeCommentsExtractor.isCommentsDisabled(YoutubeCommentsExtractor.java:272)
    at org.schabi.newpipe.extractor.comments.CommentsInfo.getInfo(CommentsInfo.java:49)
    at org.schabi.newpipe.extractor.comments.CommentsInfo.getInfo(CommentsInfo.java:29)
    at org.schabi.newpipe.extractor.comments.CommentsInfo.getInfo(CommentsInfo.java:24)
```

bjoernls:
Issue #969. Parsing exception is handled when the JSON path does not exist in the case of disabled comments.

Fixes #969

The output is now
`{"comments":[],"nextpage":null,"disabled":true}`

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
@@ -68,7 +68,14 @@ public class YoutubeCommentsExtractor extends CommentsExtractor {
     @Nullable
     private String findInitialCommentsToken(final JsonObject nextResponse)
             throws ExtractionException {
-        final String token = getJsonContents(nextResponse)
+        final JsonArray contents = getJsonContents(nextResponse);
+
+        // For videos where comments are unavailable, this would be null
+        if (contents == null) {
+            return null;
+        }
+
+        final String token = contents
                 .stream()
                 // Only use JsonObjects
                 .filter(JsonObject.class::isInstance)
@@ -104,13 +111,13 @@ public class YoutubeCommentsExtractor extends CommentsExtractor {
         return token;
     }
 
-    @Nonnull
+    @Nullable
     private JsonArray getJsonContents(final JsonObject nextResponse) {
         try {
             return JsonUtils.getArray(nextResponse,
                     "contents.twoColumnWatchNextResults.results.results.contents");
         } catch (final ParsingException e) {
-            return new JsonArray(Collections.emptyList());
+            return null;
         }
     }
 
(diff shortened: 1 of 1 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
```diff
@@ -105,6 +104,16 @@ private String findInitialCommentsToken(final JsonObject nextResponse)
         return token;
     }
 
+    @Nonnull
+    private JsonArray getJsonContents(final JsonObject nextResponse) {
+        try {
+            return JsonUtils.getArray(nextResponse,
+                    "contents.twoColumnWatchNextResults.results.results.contents");
+        } catch (final ParsingException pe) {
```
AudricV, line 112: You don't need to use `pe`, you could just `e` for the exception in this context:

```suggestion
        } catch (final ParsingException e) {
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
```diff
@@ -105,6 +104,16 @@ private String findInitialCommentsToken(final JsonObject nextResponse)
...
+            return new JsonArray(Collections.emptyList());
```
AudricV, line 113: Why do you want to use the constructor with a `Collection` where you pass an empty list? Can't you just use an empty `JsonArray` instead?

AudricV (changes requested):
Thanks for this fix!

Please use appropriate PR and commit names in the future: do not repeat the issue title but describe what you did for both things (so in this case it would be something like "[YouTube] Fix ParsingException when getting comments on videos without comments").

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
bjoernls: It's a coding convention I've become accustomed to. The line of thinking is that the function returns the content of the JSON, and should be read-only.

bjoernls: Ok :)

AudricV (approved):

PR #1039, opened 2023-03-11
[YouTube] Fix descriptive audio role property value in generated DASH manifests

AudricV:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [ ] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).

The note 2 of the section 5.8.5.5 of [ISO/IEC 23009-1](https://standards.iso.org/ittf/PubliclyAvailableStandards/c083314_ISO_IEC%2023009-1_2022(en).zip) specifies that when using the alternate role value, a description of what the alternate track is should be added.

As `alternate` is a generic role and a `description` role exists, it should be used instead of the `alternate` one.

The property value has been of course updated in the corresponding test.

Thanks to @absidue for flagging this on the Invidious Matrix channel.

PR #1041, opened 2023-03-19
Add track types to audio streams

Theta-Dev:
-> https://github.com/TeamNewPipe/NewPipe/pull/9937

I replaced the `descriptiveAudio` property of streams added by #1026 with the enum property `audioTrackType`.

The reason for this is that YouTube supports multiple types of audio streams. The ones that I have encountered are:
- Original audio (`acont=original`)
- Dubbed (`acont=dubbed`)
- Descriptive audio (`acont=descriptive`)

The new property allows creating unique, localized names for every audio track without relying on the `trackName` which is always fetched in English.


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/dashmanifestcreators/YoutubeDashManifestCreatorsUtils.java
```diff
@@ -298,8 +299,10 @@ public static void generateRoleElement(@Nonnull final Document doc,
             final Element roleElement = doc.createElement(ROLE);
 
             setAttribute(roleElement, doc, "schemeIdUri", "urn:mpeg:DASH:role:2011");
-            setAttribute(roleElement, doc, "value", itagItem.isDescriptiveAudio()
-                    ? "alternate" : "main");
+            setAttribute(roleElement, doc, "value",
+                    itagItem.getAudioTrackType() == null
+                            || itagItem.getAudioTrackType() == AudioTrackType.ORIGINAL
+                            ? "main" : "alternate");
```
AudricV, line 305: Could you apply the changes of #1039 here and use the correct property for dubbed tracks (see the standard in the PR I linked)?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -1484,16 +1487,25 @@ private ItagInfo buildAndAddItagInfoToList(
...
+                    switch (atype) {
+                        case "original":
+                        itagItem.setAudioTrackType(AudioTrackType.ORIGINAL);
+                            break;
+                        case "dubbed":
+                            itagItem.setAudioTrackType(AudioTrackType.DUBBED);
+                            break;
+                        case "descriptive":
+                            itagItem.setAudioTrackType(AudioTrackType.DESCRIPTIVE);
+                            break;
```
AudricV, line 1502: This `break` statement should be not needed.

extractor/src/main/java/org/schabi/newpipe/extractor/stream/AudioTrackType.java
```diff
@@ -0,0 +1,33 @@
...
+     * <p>
+     * A descriptive audio is an audio in which descriptions of visual elements of a video are
+     * added in the original audio, with the goal to make a video more accessible to blind and
+     * visually impaired people.
+     * </p>
+     *
+     * @see <a href="https://en.wikipedia.org/wiki/Audio_description">
+     * https://en.wikipedia.org/wiki/Audio_description</a>
+     */
+    DESCRIPTIVE,
```
AudricV, line 32: Remove this unneeded comma.

```suggestion
    DESCRIPTIVE
```

extractor/src/test/java/org/schabi/newpipe/downloader/DownloaderFactory.java
```diff
@@ -8,7 +8,7 @@ public class DownloaderFactory {
 
     public static final String RESOURCE_PATH = "src/test/resources/org/schabi/newpipe/extractor/";
 
-    private static final DownloaderType DEFAULT_DOWNLOADER = DownloaderType.REAL;
+    private static final DownloaderType DEFAULT_DOWNLOADER = DownloaderType.RECORDING;
```
AudricV, line 11: Please revert this change which you probably didn't intend to commit.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeDashManifestCreatorsTest.java
```diff
@@ -233,7 +234,8 @@ private void assertAdaptationSetElement(@Nonnull final Document document,
     private void assertRoleElement(@Nonnull final Document document,
                                    @Nonnull final ItagItem itagItem) {
         final Element element = assertGetElement(document, ROLE, ADAPTATION_SET);
-        assertAttrEquals(itagItem.isDescriptiveAudio() ? "alternate" : "main", element, "value");
+        assertAttrEquals(itagItem.getAudioTrackType() == null || itagItem.getAudioTrackType() == AudioTrackType.ORIGINAL
+                ? "main" : "alternate", element, "value");
```
AudricV, line 238: Applying the changes from my comment above about audio track type.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -810,6 +811,8 @@ public String getErrorMessage() {
             "\\bc\\s*&&\\s*d\\.set\\([^,]+\\s*,\\s*(:encodeURIComponent\\s*\\()([a-zA-Z0-9$]+)\\("
     };
     private static final String STS_REGEX = "signatureTimestamp[=:](\\d+)";
+    private static final String AUDIO_STREAM_TYPE_REGEX =
```
AudricV, line 814: Due to the number of executions of this code, please compile the regex string as a `Pattern` and pass directly this pattern.

This would increase by a lot the performance of your current code, especially when it will run on servers such as Piped backends.

AudricV (changes requested):
Thanks for your PR!

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/dashmanifestcreators/YoutubeDashManifestCreatorsUtils.java
```diff
@@ -298,15 +301,35 @@ public static void generateRoleElement(@Nonnull final Document doc,
...
+                case ORIGINAL:
+                    return "main";
+                case DUBBED:
+                    return "dub";
+                case DESCRIPTIVE:
+                    return "description";
+            }
+            return "alternate";
+        }
+        return "main";
```
FireMasterK, line 330:
```suggestion
                default:
                    return "alternate";
            }
```

extractor/src/main/java/org/schabi/newpipe/extractor/stream/AudioTrackType.java
```diff
@@ -0,0 +1,33 @@
+package org.schabi.newpipe.extractor.stream;
+
+/**
+ * An enum representing the track type of an {@link AudioStream} extracted by a {@link
+ * StreamExtractor}.
+ */
+public enum AudioTrackType {
+    /**
+     * The original audio track of the video
```
AudricV, line 9: While that's not the case currently, maybe one day audio services we support and/or add could support multiple audio tracks, so I would use `stream` instead of `video` here.

Also, as multiple original audio tracks exist for some services (YouTube for instance), I would use an indefinite article.

```suggestion
     * An original audio track of the stream.
```

For consistency with most of the JavaDocs, please also add a period at the end of the title of the ones you added.

extractor/src/main/java/org/schabi/newpipe/extractor/stream/AudioTrackType.java
```diff
@@ -0,0 +1,33 @@
...
+     */
+    ORIGINAL,
+
+    /**
+     * Audio track with the original voices replaced, typically in a different language
```
AudricV, line 14:
```suggestion
     * An audio track with the original voices replaced, typically in a different language.
```

extractor/src/main/java/org/schabi/newpipe/extractor/stream/AudioStream.java
```diff
@@ -506,21 +500,9 @@ public Locale getAudioLocale() {
...
-     * @return {@code true} this audio stream is a descriptive audio, {@code false} otherwise
-     * @see <a href="https://en.wikipedia.org/wiki/Audio_description">
-     *     https://en.wikipedia.org/wiki/Audio_description</a>
-     */
-    public boolean isDescriptive() {
-        return isDescriptive;
+    @Nullable
+    public AudioTrackType getAudioTrackType() {
+        return audioTrackType;
     }
```
AudricV, line 506: Please add documentation for this getter, this one will be likely the one which will be used the most by clients.

AudricV (changes requested):
Almost good, you only have some small changes remaining (the ones below + the ones of FireMasterK). Thanks again!

AudricV (approved):
I just pushed a commit removing unused imports, public method modifiers and the IP you used to generate stream URLs in `YoutubeParsingHelperTest`, which were the remaining changes that were needed to be done for me.

The commits of this PR will be squashed and merged to avoid unneeded mocks additions and deletions to the Git repository.

Thanks again!

PR #1043, opened 2023-03-25
Simplify time ago calculation.

Isira-Seneviratne:
----
Simplify the time ago calculation method. This removes the workaround used by PrettyTime (a PR to remove the library from NewPipe is [here](https://github.com/TeamNewPipe/NewPipe/pull/9878)).


Stypox:
See https://github.com/TeamNewPipe/NewPipe/pull/9878#issuecomment-1495912563

PR #158, opened 2019-04-25
Implement time ago parser and improve localization handling

mauriciocolli:
Used some of the work that @wojcik-online did as the base, finishing the implementation of the time ago parser.

- Time ago parsing for stream items where the service shows it (search, channel videos list)
- Handle special cases for languages where the number is not shown
    - Hebrew, for example, shows "לפני שעתיים" for "2 hours ago". We cannot assume that it is 1 for those cases (popular among other languages), as it would be wrong in the bigger time units.
- Rework the Downloader base implementation, allowing for more advanced things to be done
    - Get the response headers
    - Response code status and message
    - Possibility of making other type of requests (`OPTIONS`, `HEAD`, `POST`...)
- Separate the localization from the content country (just like YouTube let's the user choose both).
- Parse watching count in live streams items
- Fix parsing of future videos (those with an option to set a reminder)

---

**Edit:** So I did see the PR #196 (related TeamNewPipe/NewPipe#2632), and as a result, no language parsing is needed anymore and english will be forced over others.

However, I think the language handling contained in this PR is still useful for NewPipe as we can now force languages easily and, in the future, help parse sites in other languages.

PR #1045, opened 2023-04-02
[YouTube] Extract trends from A/B tested "Videos" tab and fix extraction of trends name from A/B tested new title design

### LINKED ISSUE #1046: [YouTube] Trends are not returned sometimes
YouTube is A/B testing a new trending UI which completely reworks what data is sent and how it is displayed. This new structure is currently not supported by the extractor.

Here is a screenshot of the new UI in French:

yt_new_trending_ui

With this UI, the `Recently trending` section is removed and the trending videos section has now a title (`Trending videos`), which makes [the current code returning no items](https://github.com/TeamNewPipe/NewPipeExtractor/blob/1b6fe5edd6b479d50d96400bc3c4e7f78fd5caa7/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeTrendingExtractor.java#L124-L126), as the current/previous UI returned no title for trending videos but one for recently trending videos and shorts.

Here is how looks the new videos trending data structure which is now returned (some fields have been omitted):

```json
{
    "itemSectionRenderer": {
        "contents": [
            {
                "shelfRenderer": {
                    "title": {
                        "runs": [
                            {
                                "text": "Trending videos"
                            }
                        ]
                    },
                    "endpoint": {
                        "commandMetadata": {
                            "webCommandMetadata": {
                                "url": "/feed/trending?bp=4gIOGgxtb3N0X3BvcHVsYXI%3D",
                                "webPageType": "WEB_PAGE_TYPE_BROWSE",
                                "rootVe": 6827,
                                "apiUrl": "/youtubei/v1/browse"
                            }
                        },
                        "browseEndpoint": {
                            "browseId": "FEtrending",
                            "params": "4gIOGgxtb3N0X3BvcHVsYXI%3D"
                        }
                    },
                    "content": {
                        "horizontalListRenderer": {
                            "items": [
                                {
                                    "gridVideoRenderer": {
                                        // A gridVideoRenderer without description
                                    }
                                },
                                // Other gridVideoRenderers
                            ]
                        }
                    }
                }
            }
        ]
    }
}
```

In this A/B test, a new trending tab called `Videos` has been added and replaces the `Now` tab. It returns only the trending videos without anything else and is accessible with a `browse` request on the InnerTube API with the `FEtrending` `browseId` and the parameters `4gIOGgxtb3N0X3BvcHVsYXI%3D` (a protobuf object which means as a raw value `44 { 3: "most_popular"}`).

This tab works regardless of the inclusion of the request in the A/B test.

Video items on this tab contains their description like on the current/previous UI and on search results.

Reproduce rate of this A/B test is very low right now (1 or 2% in my region at the time this issue is written).

Theta-Dev:
- [X] No API changes

YouTube is A/B testing a new layout for the trending page, where they moved the list of trending videos into a separate tab.

Issue reported on Reddit: https://www.reddit.com/r/NewPipe/comments/127tyh8/no_v%C3%ADdeos_is_it_a_bug/

Pictures here: https://code.thetadev.de/ThetaDev/rustypipe/src/branch/main/notes/AB_Tests.md#4-video-tab-on-the-trending-page

I have changed the extractor to fetch this new tab. If the new layout is not enabled, YT simple returns the *Now* page like before.

The parser had to be adjusted, because we cannot filter out all shelves with titles any more (the shelves in the *Videos* tab have titles).

Fixes #1046.


Stypox (approved):

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeTrendingExtractor.java
```diff
@@ -41,12 +41,15 @@
...
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.util.stream.Stream;
 
 import javax.annotation.Nonnull;
 
 public class YoutubeTrendingExtractor extends KioskExtractor<StreamInfoItem> {
     private JsonObject initialData;
 
+    private static final String VIDEO_TAB_PARAMS = "4gIOGgxtb3N0X3BvcHVsYXI%3D";
```
AudricV, line 51: Please rename this field to `VIDEOS_TAB_PARAMS`.

AudricV (changes requested):
Except the following comments, your changes look good to me.

However, I cannot reproduce the A/B test anymore. Do you think YouTube reverted their test or stopped it? This may lead to the removal of this parameter, and may break trends extraction instead of fixing it in the future if using these parameters return a bad response (unlikely to happen, as wrong parameters seem to be ignored).

FireMasterK (approved):

Theta-Dev:
I cannot get the A/B test with the videos tab any more, too. Maybe YouTube did scrap this idea, maybe it's just the end of the test run and they are evaluating the results.

I would keep the change for now, if the new layout does not appear again within half a year we could take it out again for the sake of simplicity.

Requesting non-existant tabs will simply return the default tab (it's the same thing for channel tabs), so I dont see any issues.

FireMasterK:
I'm seeing an increased error rate on Piped when fetching the trending page since the 27th of April, maybe there's a new A/B test? Just speculating

image

Theta-Dev:
are you using this PR or the regular version of the Extractor?

AudricV:
> Btw, why do we throw an exception when we fail to extract a text that is always the same?

For localization purposes I think. When support of YouTube extraction in other content languages than English will be added again, an incorrect information would be returned for these non-English languages.

Theta-Dev:
I see. Falling back to "Trending" if the title could not be extracted would not be bad either, so I changed it.

AudricV:
> Falling back to "Trending" if the title could not be extracted would not be bad either, so I changed it.

That's a very bad idea to return an English-only string, due to the reasons explained above: if YouTube changes their Trending title and how it is accessed, the extractor would return a value which would always pass the test. That's why I reverted this change.

> When and why was the extraction of other languages removed, though?

It was removed in #158, due to parsing issues, as several parts of the YouTube extractor relies on English strings (age-restricted and paid content checks, accessibility data parsing, ...). I worked a few months ago on trying to restoring this support and I try to will resume my work after we have finished with major features on the extractor.

AudricV (approved):

PR #1049, opened 2023-04-13
[SoundCloud] Fix certain playlists showing as empty

petlyh:
No NewPipe changes required.

For certain SoundCloud playlists, the initial playlist data does not include the full track info for any of the tracks. Because of this, no items get collected in `getIntialPage` and NewPipe shows the playlist as empty.

This change makes it so that `getPage` is used in initial extraction in those cases to prevent `getInitialPage` from returning no items.

I have tested that other playlists still work, including actually empty ones.

Fixes TeamNewPipe/NewPipe#10012

|Before|After|
|-|-|
|||


Stypox:
This should be changed on NewPipe's side. Does this commit fix the issue? https://github.com/TeamNewPipe/NewPipe/pull/10018/commits/e236d370c85cbc9766fbd64330d6cf5754fee81d

PR #1050, opened 2023-04-16
[YouTube] Add support for CarouselHeaderRenderer

Theta-Dev:
- [X] no API changes

YouTube is using an alternative header design for certain events like the Coachella music festival (I think I have seen something similar for last years football world cup, so I dont think this is a one-off thing).

I made the required changes to the ChannelExtractor to parse this header correctly.

image


extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeChannelExtractorTest.java
```diff
@@ -648,4 +648,94 @@ public void testVerified() throws Exception {
             assertFalse(extractor.isVerified());
         }
     }
+
+    public static class Coachella implements BaseChannelExtractorTest {
```
Stypox, line 652: Call this test "CarouselHeader" instead of "Coachella", so that it is clear what is being tested

Stypox (changes requested):
LGTM except for the small comment

Stypox (approved):

PR #1052, opened 2023-04-16
[PeerTube] Fix multi level comment replies

TobiGr:
- [x] ~~I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API.~~ (not needed)

Nested replies are stored in the original thread response and are now stored inside the Page's body returned by `getReplies()`.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubeCommentsExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubeCommentsExtractor.java
@@ -3,6 +3,7 @@ package org.schabi.newpipe.extractor.services.peertube.extractors;
 import com.grack.nanojson.JsonArray;
 import com.grack.nanojson.JsonObject;
 import com.grack.nanojson.JsonParser;
+import com.grack.nanojson.JsonParserException;
 import org.schabi.newpipe.extractor.Page;
 import org.schabi.newpipe.extractor.StreamingService;
 import org.schabi.newpipe.extractor.comments.CommentsExtractor;
@@ -17,6 +18,7 @@ import org.schabi.newpipe.extractor.services.peertube.PeertubeParsingHelper;
 import org.schabi.newpipe.extractor.utils.Utils;
 
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 
 import static org.schabi.newpipe.extractor.services.peertube.PeertubeParsingHelper.COUNT_KEY;
 import static org.schabi.newpipe.extractor.services.peertube.PeertubeParsingHelper.ITEMS_PER_PAGE;
@@ -26,6 +28,9 @@ import static org.schabi.newpipe.extractor.utils.Utils.isNullOrEmpty;
 import javax.annotation.Nonnull;
 
 public class PeertubeCommentsExtractor extends CommentsExtractor {
+    static final String CHILDREN = "children";
+    private static final String IS_DELETED = "isDeleted";
+    private static final String TOTAL = "total";
 
     /**
      * Use {@link #isReply()} to access this variable.
@@ -49,7 +54,7 @@ public class PeertubeCommentsExtractor extends CommentsExtractor {
         }
     }
 
-    private boolean isReply() throws ParsingException {
+    boolean isReply() throws ParsingException {
         if (isReply == null) {
             if (getOriginalUrl().contains("/videos/watch/")) {
                 isReply = false;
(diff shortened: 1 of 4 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
```diff
@@ -306,6 +307,19 @@ public static <K, V> boolean isNullOrEmpty(final Map<K, V> map) {
...
+     * <br>
+     * The page is empty when the page's URL is null or empty and the page's body is null.
+     *
+     *  @param page the {@link Page} to check
+     * @return whether the page is empty when the page's URL is null or empty
+     * and the page's body is null.
+     */
+    public static boolean isNullOrEmpty(final Page page) {
+        return page == null || (isNullOrEmpty(page.getUrl()) && page.getBody() == null);
+    }
```
Stypox, line 321: We already have `Page.isValid`, so you might want to change that to be
```
return page != null && (
    !isNullOrEmpty(page.getUrl())
    || !isNullOrEmpty(page.getIds())
    || page.getBody() != null);
```

Stypox (changes requested):
Code looks good to me. I tested and it works. Just one small thing to fix :-)

extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
TobiGr: The logic of your posted code does not seem right to me. When the page is `null`, the method will always return `false`;
I'd go with
``` Java
return page == null || (
        isNullOrEmpty(page.getUrl())
                && isNullOrEmpty(page.getIds())
                && page.getBody() == null
);
```

Stypox: Yes, that's because I'm talking about the `Page.isValid(Page page)` function, which already exists and whose logic is inverted. This function should be removed in favour of that one.

TobiGr: Ah ok. Yeah, that makes sense :+1:

TobiGr: Silly me did not check whether the method is still used...
I needed it for a previous implementation. The `isNullOrEmpty(final Page page)` is not used at all. I just removed the method and that should be enough. A page should still have some kind of identifying info in it.

extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubeCommentsExtractor.java
```diff
@@ -95,36 +102,46 @@ public InfoItemsPage<CommentsInfoItem> getPage(final Page page)
...
+                    total = json.getArray(CHILDREN).size();
+                    collectRepliesFrom(collector, json);
+                } else {
+                    total = json.getLong(TOTAL);
+                    collectCommentsFrom(collector, json);
+                }
+            } else {
+                throw new ExtractionException("Unable to get PeerTube kiosk info");
+            }
+        } else {
```
AudricV, line 129: Aren't these `else` statements unneeded?

AudricV (changes requested):

extractor/src/test/java/org/schabi/newpipe/extractor/services/peertube/PeertubeCommentsExtractorTest.java
```diff
@@ -14,10 +14,9 @@
 
 import java.io.IOException;
 import java.util.List;
+import java.util.Optional;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.*;
```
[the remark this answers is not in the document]
Stypox: Do we really care about imports? What we should do is distribute IDE settings to prevent these kind of automatic refactorings. Resolving review comments about imports just wastes time IMO.

extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubeCommentsExtractor.java
TobiGr: Do you want me to add another return above? I could do that, but IMO if/else is better here.

AudricV (approved):

PR #1056, opened 2023-04-30
[YouTube] Switch to new search suggestion domain and improve error handling

AudricV:

- Switch to the new domain used by YouTube for search suggestions, `suggestqueries-clients6.youtube.com`, and add the `xhr` query parameter with the `t` value, to allow getting responses without requiring trim;
- Use the Java 8 Stream API to collect search suggestions and improve invalid response detection by checking whether the content type of the response returned is JSON;
- Improve `YoutubeSuggestionExtractorTest` and update its corresponding mocks.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSuggestionExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSuggestionExtractor.java
@@ -1,6 +1,27 @@
+/*
+ * Created by Christian Schabesberger on 28.09.16.
+ *
+ * Copyright (C) Christian Schabesberger 2015 <chris.schabesberger@mailbox.org>
+ * YoutubeSuggestionExtractor.java is part of NewPipe Extractor.
+ *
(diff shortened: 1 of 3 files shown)


FireMasterK:
Would it make sense to fall back to the old search API if the new one fails? I've observed the new API being rate limited on Piped at times, so I think it could help.

AudricV:
Isn't the same rate limit applied when you switch the search suggestion host? It happens for YouTube manifests.

FireMasterK:
I've never really checked that, will see when I get rate limited again.

FireMasterK:
It looks like the rate limits seem to be different, I tried the new search API on a rate-limited/blocked Piped node, and it worked when the old one failed.

AudricV:
Ok, so what should we do? Pick a search suggestion host randomly?

FireMasterK:
We could do that indeed. I was thinking we could store a static variable to store the last time we were rate limited as a `long`, and if been < 5 minutes, we use the old API. I can help with this if you'd like!

FireMasterK:
It looks like I was using ipv6 on my testing, but my backends run on ipv4 only, the rate limit is also indeed on the new API, so I think this is ready to merge!

PR #1061, opened 2023-05-11
Extract playlists description

ChunkyProgrammer:

This PR adds a getDescription method for playlist extraction + tests for the description. I am not planning on creating a PR in NewPipe to support this (I implemented this to hopefully add this to the Piped API)

### API changes
This PR adds
- `PlaylistExtractor.getDescription()` and the corresponding  `PlaylistInfo.getDescription()`
- `PlaylistInfoItemExtractor.getDescription()` and the corresponding  `PlaylistInfoItem.getDescription()`

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/playlist/PlaylistExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/playlist/PlaylistExtractor.java
@@ -21,6 +21,8 @@ public abstract class PlaylistExtractor extends ListExtractor<StreamInfoItem> {
 
     public abstract long getStreamCount() throws ParsingException;
 
+    public abstract String getDescription() throws ParsingException;
+
     @Nonnull
     public String getThumbnailUrl() throws ParsingException {
         return "";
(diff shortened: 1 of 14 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/playlist/PlaylistExtractor.java
```diff
@@ -21,6 +22,8 @@ public PlaylistExtractor(final StreamingService service, final ListLinkHandler l
 
     public abstract long getStreamCount() throws ParsingException;
 
+    public abstract Description getDescription() throws ParsingException;
```
AudricV, line 25: I think you should provide a default implementation returning `Description.EMPTY_DESCRIPTION`, like done for other methods below, as most of the implementations return this value.

AudricV (changes requested):
Thanks, you need to also implement these changes in `PlaylistInfo`.

TobiGr:
> Thanks, you need to also implement these changes in PlaylistInfo.

I already implemented that locally. There are a few more changes which I am going to push within the next days.

extractor/src/main/java/org/schabi/newpipe/extractor/playlist/PlaylistExtractor.java
ChunkyProgrammer: Would you still want this to be done? TobiGr added support for PeerTube and Soundcloud

TobiGr:
@fynngodau I am unsure which parts of the album description should be used as description. Can you give me an advice?
My current plan is to add `.tralbum-about`, `.tralbum-credits` and `#license`.

FireMasterK (approved):

TobiGr (approved):

fynngodau:
The copyright text is debatable because the individual tracks also always have copyright info, so it is not strictly necessary here and could be removed. The credits otoh should be read out here, as long as there's no separate method `getCredits`.

PR #1063, opened 2023-05-20
[YouTube] Add support for ultralow audio formats

yshalsager:

---

While using yt-dlp I noticed that there are two new audio formats for ultra-low connections (599 m4a and 600 webm) which NewPipe doesn't support yet, so I added them.

```
[info] Available formats for NyLdD8wk2XI:
ID  EXT   RESOLUTION FPS CH │   FILESIZE  TBR PROTO │ VCODEC       VBR ACODEC      ABR ASR MORE INFO
──────────────────────────────────────────────────────────────────────────────────────────────────────────────
599 m4a   audio only      2 │    6.61MiB  31k dash  │ audio only       mp4a.40.5   31k 22k ultralow, m4a_dash
600 webm  audio only      2 │    7.46MiB  35k dash  │ audio only       opus        35k 48k ultralow, webm_dash
139 m4a   audio only      2 │   10.48MiB  49k dash  │ audio only       mp4a.40.5   49k 22k low, m4a_dash
249 webm  audio only      2 │   10.89MiB  51k dash  │ audio only       opus        51k 48k low, webm_dash
250 webm  audio only      2 │   13.89MiB  65k dash  │ audio only       opus        65k 48k low, webm_dash
140 m4a   audio only      2 │   27.80MiB 129k dash  │ audio only       mp4a.40.2  129k 44k medium, m4a_dash
251 webm  audio only      2 │   27.06MiB 126k dash  │ audio only       opus       126k 48k medium, webm_dash
```

Here's a screenshot from the build app after these changes, works as expected.

<details>
<summary>Screenshot</summary>

Screenshot_1684584369

</details>

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/ItagItem.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/ItagItem.java
@@ -51,14 +51,14 @@ public class ItagItem implements Serializable {
             //////////////////////////////////////////////////////////////////
             new ItagItem(171, AUDIO, WEBMA, 128),
             new ItagItem(172, AUDIO, WEBMA, 256),
+            new ItagItem(599, AUDIO, M4A, 32),
             new ItagItem(139, AUDIO, M4A, 48),
             new ItagItem(140, AUDIO, M4A, 128),
             new ItagItem(141, AUDIO, M4A, 256),
+            new ItagItem(600, AUDIO, WEBMA_OPUS, 35),
             new ItagItem(249, AUDIO, WEBMA_OPUS, 50),
             new ItagItem(250, AUDIO, WEBMA_OPUS, 70),
             new ItagItem(251, AUDIO, WEBMA_OPUS, 160),
-            new ItagItem(599, AUDIO, M4A, 32),
-            new ItagItem(600, AUDIO, WEBMA_OPUS, 48),
 
             /// VIDEO ONLY ////////////////////////////////////////////
             //           ID      Type     Format  Resolution  FPS  ////
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/ItagItem.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/ItagItem.java
@@ -57,6 +57,8 @@ public class ItagItem implements Serializable {
             new ItagItem(249, AUDIO, WEBMA_OPUS, 50),
             new ItagItem(250, AUDIO, WEBMA_OPUS, 70),
             new ItagItem(251, AUDIO, WEBMA_OPUS, 160),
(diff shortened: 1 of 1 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/ItagItem.java
```diff
@@ -57,6 +57,8 @@ public class ItagItem implements Serializable {
             new ItagItem(249, AUDIO, WEBMA_OPUS, 50),
             new ItagItem(250, AUDIO, WEBMA_OPUS, 70),
             new ItagItem(251, AUDIO, WEBMA_OPUS, 160),
+            new ItagItem(599, AUDIO, M4A, 32),
+            new ItagItem(600, AUDIO, WEBMA_OPUS, 48),
```
AudricV, line 61: Itag 600 is not a 48k audio, that's the itag 249. Instead, it is a 30-35-40k audio.

AudricV (changes requested):
Thanks, please respect the media format and bitrate orders in the list.

A NewPipe support for the added itags is also needed, as they are now whitelisted: see TeamNewPipe/NewPipe#9748.

By the way, this change will probably make the app download these 30k audios instead of the 50k ones for demuxed video streams, when these audio streams are available. This audio selection bug should be definitively fixed before we release a new NewPipe version.

Also use a better branch name in the future, like described in our contribution guidelines.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/ItagItem.java
yshalsager: @AudricV Thanks for your review! I have updated the PR.
Please let me know if I should submit another PR to the app to whitelist the new formats now or after merging this one.

AudricV (approved):
Thanks, though I am not sure about what average value we should give to the itag 600. Is there someone else which may have more input about that?

This PR will be merged using the squash and merge option.

AudricV:
> I am not sure about what average value we should give to the itag 600. Is there someone else which may have more input about that?

I didn't get an answer on this question, so I will keep the average value as it is.

> This audio selection bug should be definitively fixed before we release a new NewPipe version.

See TeamNewPipe/NewPipe#10446.

> I think you will need to do so

@yshalsager Do you want to do so or should I do this myself? You simply need to update the extractor commit and add 599 and 600 formats.

I am merging the pull request, thanks for your contribution!

yshalsager:
> @yshalsager Do you still want to do so or should I do this myself? You simply need to update the extractor commit and, of course, to add 599 and 600 formats to the allowed itags list of NewPipe.

Sorry, I am on a vacation for a few weeks, would be grateful if you added it.

> I am merging the pull request, thanks for your contribution!

My pleasure, thanks for your review and awesome efforts!

PR #1068, opened 2023-06-01
[YouTube] Fix parsing short relative date formats (English only)

### LINKED ISSUE #1067: [YouTube] New short date format
YouTube is A/B testing a new, shorter format for video upload dates.

The long format is still used for comments and playlist update dates.

### en-US

- Year: `y`
- Month: `mo`
- Week: `w`
- Day: `d`
- Hour: `h`
- Minute: `min`

### en-GB

- Year: `yr`
- Month: `mo`
- Week: `wk`
- Day: `day(s)`
- Hour: `hr`
- Minute: `min`

image

The A/B test currently occurs very rarely (14 out of 1000 requests).

Here is a visitor data cookie for testing: `CgtJeFI1cWVSWVBTNCj6yc6jBg%3D%3D`

Theta-Dev:
- [X] no API changes

I added support for the new, short date format (e.g. 1wk ago).

Fixes #1067

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/localization/TimeAgoParser.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/localization/TimeAgoParser.java
@@ -93,6 +93,7 @@ public class TimeAgoParser {
         final String escapedSeparator = patternsHolder.wordSeparator().equals(" ")
                 // From JDK8 → \h - Treat horizontal spaces as a normal one
                 // (non-breaking space, thin space, etc.)
+                // Also split the string on numbers to be able to parse strings like "2wk"
                 ? "[ \\t\\xA0\\u1680\\u180e\\u2000-\\u200a\\u202f\\u205f\\u3000\\d]"
                 : Pattern.quote(patternsHolder.wordSeparator());
 
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/localization/TimeAgoParser.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/localization/TimeAgoParser.java
@@ -93,7 +93,7 @@ public class TimeAgoParser {
         final String escapedSeparator = patternsHolder.wordSeparator().equals(" ")
                 // From JDK8 → \h - Treat horizontal spaces as a normal one
                 // (non-breaking space, thin space, etc.)
-                ? "[ \\t\\xA0\\u1680\\u180e\\u2000-\\u200a\\u202f\\u205f\\u3000]"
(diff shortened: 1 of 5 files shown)


TobiGr:
LGTM, but what about the other languages? I'd guess we need to update them, too.

extractor/src/main/java/org/schabi/newpipe/extractor/localization/TimeAgoParser.java
```diff
@@ -93,7 +93,7 @@ private boolean textualDateMatches(final String textualDate, final String agoPhr
         final String escapedSeparator = patternsHolder.wordSeparator().equals(" ")
                 // From JDK8 → \h - Treat horizontal spaces as a normal one
                 // (non-breaking space, thin space, etc.)
-                ? "[ \\t\\xA0\\u1680\\u180e\\u2000-\\u200a\\u202f\\u205f\\u3000]"
```
FireMasterK, line 96: Why is `\d` added here? Could you update the above comment to indicate this?

FireMasterK (approved):

extractor/src/main/java/org/schabi/newpipe/extractor/localization/TimeAgoParser.java
Theta-Dev: To be able to parse strings where numbers directly follow digit (eg. `1wk`. I'll add that to the comment

Theta-Dev:
Parsing YouTube data in other languages is currently disabled and it would require more extensive changes to the dictionary and the parser.

I have implemented (and tested) a parser that works with all languages. Here is the dictionary for it. Also note that there are some special cases that have to be handled seperately (e.g. in French `a` is both an article and the short form of "years").

https://code.thetadev.de/ThetaDev/rustypipe/src/branch/main/testfiles/dict/dictionary.json

TobiGr (approved):

AudricV:
> Parsing YouTube data in other languages is currently disabled

Yes, but the timeago parser can be also used for something else than YouTube by clients, even if that's its main goal.

Is the data you provided extracted from YouTube? Do you wish to do other languages support using your similar approach to us? I think we should have a parser separating date units (seconds, hours, days, ...), digits (1, 2, 3, ...) and number units (tens, hundreds, thousands, ...) for each language we want to add support.

> in French a is both an article and the short form of "years"

Did you mean the verb `have` at the third singular person and the short form of `years`? There is not `a` article, but a `à` one.

Theta-Dev:
This is the French term in question (5 years ago):

`il y a 5 a`

The data from the parsing dictionary is a combination of data extracted from YouTube and the CLDR repository.

AudricV:
Merging this PR, as we would need to overhaul the timeago parser system to work with other languages on short time units. Thanks for the fix!

PR #1070, opened 2023-06-02
[YouTube] Check whether player responses are valid for all InnerTube clients used

FireMasterK:

This is in light of the recent YouTube changes.
Also removed unneeded empty string fallback when the `videoId` property in the player responses is missing.


AudricV:
You should change the Javadoc and the parameters of the method used, as it doesn't apply anymore to only additionnal player responses and YouTube return a different video error for the IP rate limit case.

You should also change this behavior for the TV client we use.

AudricV (approved):

PR #1071, opened 2023-06-06
Init services at the correct place

TobiGr:

I saw that flaw while working on something different. This also increases readability


extractor/src/main/java/org/schabi/newpipe/extractor/ServiceList.java
```diff
@@ -34,27 +34,21 @@
...
      */
     private static final List<StreamingService> SERVICES = Collections.unmodifiableList(
-            Arrays.asList(
-                    YouTube = new YoutubeService(0),
-                    SoundCloud = new SoundcloudService(1),
-                    MediaCCC = new MediaCCCService(2),
-                    PeerTube = new PeertubeService(3),
-                    Bandcamp = new BandcampService(4)
-            ));
+            Arrays.asList(YouTube, SoundCloud, MediaCCC, PeerTube, Bandcamp));
```
AudricV, line 51: As you are changing this, I think you should use `List.of` instead on the fly.

AudricV (approved):

PR #1074, opened 2023-06-15
Add more audio media formats and MediaFormat.getAllForMimeType(mimeType)

TobiGr:

Add a few more media formats that can be used by SoundCloud creators to upload their streams.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/MediaFormat.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/MediaFormat.java
@@ -22,8 +22,11 @@ package org.schabi.newpipe.extractor;
  * along with NewPipe.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import javax.annotation.Nonnull;
 import java.util.Arrays;
+import java.util.List;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 /**
  * Static data about various media formats support by NewPipe, eg mime type, extension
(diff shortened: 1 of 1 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/MediaFormat.java
```diff
@@ -121,6 +132,12 @@ public static MediaFormat getFromMimeType(final String mimeType) {
                 .orElse(null);
     }
 
+    public static List<MediaFormat> getAllFromMimeType(final String mimeType) {
+        return Arrays.stream(MediaFormat.values())
+                .filter(mediaFormat -> mediaFormat.mimeType.equals(mimeType))
+                .collect(Collectors.toList());
+    }
```
AudricV, line 139: This method should be documented and annotations should be added.

Also, should we make the list returned unmodifiable?

extractor/src/main/java/org/schabi/newpipe/extractor/MediaFormat.java
```diff
@@ -41,9 +43,18 @@ public enum MediaFormat {
     M4A       (0x100, "m4a",       "m4a",  "audio/mp4"),
     WEBMA     (0x200, "WebM",      "webm", "audio/webm"),
     MP3       (0x300, "MP3",       "mp3",  "audio/mpeg"),
+    MP2       (0x310, "MP2",       "mp2",  "audio/mpeg"),
     OPUS      (0x400, "opus",      "opus", "audio/opus"),
     OGG       (0x500, "ogg",       "ogg",  "audio/ogg"),
     WEBMA_OPUS(0x200, "WebM Opus", "webm", "audio/webm"),
+    AIFF      (0x600, "AIFF",      "aiff", "audio/aiff"),
+    /**
+     * Same as {@link MediaFormat.AIFF}, just with the shorter suffix / file extension
```
AudricV, line 52:
```suggestion
     * Same as {@link MediaFormat.AIFF}, just with the shorter suffix/file extension.
```

AudricV (changes requested):

extractor/src/main/java/org/schabi/newpipe/extractor/MediaFormat.java
TobiGr: I don't think there is any good reason to make the list unmodifiable. The user of the interface might want to use that List, e.g. reduce the entries.

AudricV (approved):

Issue #1077, opened 2023-06-26
[YouTube] Cannot get streams of playlists with only shorts


AudricV:
With the rollout of the sort by oldest feature on the new filtering system, YouTube also added filters to filter playlists, allowing users to filter videos and shorts inside playlists.

When shorts filter is enabled or when there are only shorts in a playlist, YouTube return shorts UI elements instead of standard video ones, the ones that are also used for shorts shelves in searches and suggestions and shorts in the corresponding channel tab.

Short UI elements (`reelItemRenderer`s) don't include any information about shorts' upload date, shorts' uploader name and shorts' uploader URL; their duration needs to be parsed from accessibility data and their view count extracted is not exact.

The data structure has changed, switching from `playlistVideoListRenderer`s and `playlistVideoRenderer`s to `richGridRenderer`s and `richItemRenderer`s + `reelItemRenderer`s respectively), so extraction of playlists in this case returns no streams. `reelItemRenderer`s are also not supported by the extractor.

The shorts playlist UI doesn't give any continuation when there are more than 100 shorts in a playlist. This seems to be a bug on YouTube's side.

I think we have three possibilities to fix or mitigate the issues these changes raised:
- ~~use a mobile client (`MWEB`, `ANDROID` and probably `IOS`), on which filters are available but video UI is returned by default when there are only shorts in a playlist;~~ (edit: shorts UI returned and forced for some playlists)
- use the `TVHTML5` client, where no filters and no precise view count are available (only human ones such as `1k views` are returned);
- keep the desktop variant and support `reelItemRenderer`s, even if we loose a lot of information.

Theta-Dev:
Using the mobile client is a solution at the moment, but since this is a A/B test, I dont think that will work permanently. If YouTube wants their crippled shorts UI, they are going to implement it everywhere.

I guess we need to support reel item renderers for playlists, even though we'll lose a lot of metadata.

AudricV:
I also found that currently, if we use a continuation to request the first page, regular video items and continuations are returned for these playlists! The new playlist header isn't returned in this case though.

Here is the protobuf schema for playlists continuations:

```proto
syntax = "proto3";

message PlaylistContinuation {
  PlaylistContinuationParameters parameters = 80226972;
}

message PlaylistContinuationParameters {
  string browseId = 2;
  string continuationProperties = 3; // A PlaylistContinuationProperties message encoded as a Base64 string
  string playlistId = 35;
}

[...]

Issue #611, opened 2021-04-13
Cannot get video in some Auto-generated channels


westlinkin:
Like: 

Music: https://www.youtube.com/channel/UC-9-kyTW8ZkZNDHQJ6FgpwQ

gaming: https://www.youtube.com/gaming

TiA4f8R:
We need to change something when extracting these channels. It returns 404 because NewPipe adds a `/c/` before `gaming`, which isn't recognized by YouTube.

westlinkin:
> We need to change something when extracting these channels. It returns 404 because NewPipe adds a `/c/` before `gaming`, which isn't recognized by YouTube.

what about the music channel?

TiA4f8R:
For the music channel, this channel has no `Videos` tab. Channel extraction on NewPipe Extractor relies currently on the `Videos` tab for now. Some work for the channel tabs extraction has been already made in #279. However, someone needs to reuse the work already made to create a new PR rebased on the `dev` branch.

PR #1082, opened 2023-07-19
Add support for channel tabs and channel tags

### LINKED ISSUE #227: Playlist in channels
I am currently working on bandcamp support [in my fork](https://github.com/fynngodau/NewPipeExtractor). Even though it was difficult to understand at first, I feel like I now know what I'm doing.

I noticed that `ChannelExtractor.getInitialPage()` must return an `InfoItemsPage` of type `<StreamInfoItem>`. This means that channels can't contain playlist items, unlike search results, where the return type is `InfoItemsPage<InfoItem>`.

For bandcamp support, this is inconvenient, as streams (tracks) don't appear on their channel (artist's profile) themselves when they are part of a playlist (an album).

That's why I'd like to ask this to be changed so that channels can also contain playlists.

### LINKED ISSUE #501: [FR] YouTube Topic Channels
What is actually needed here to support getting content from topic channels? I want to look into it for some time now, but I have no idea where to start.

### LINKED ISSUE #585: [Bandcamp] Support albums in artists' profile
Bandcamp only displays tracks (streams), not albums (playlists), on the artist's profile page (channel) because of #227.

AudricV:
  Already done by someone else: TeamNewPipe/NewPipe#9182, which needs to be updated to the latest changes made in this pull request (see below).

This PR supersedes #951 by reusing the changes from @Theta-Dev and @Stypox and applying the latest requested changes where relevant. Thanks a lot for your original work!

It introduces support of channel tabs and tags. The channel tab changes are breaking changes, as `ChannelExtractor` is now only an `Extractor` abstract implementation instead of a `ListExtractor`one: this means that `ChannelExtractor` do not provide `StreamInfoItem` anymore. They can be extracted from channel tabs instead.

Channel tabs (which are strings to be compliant with the current `LinkHandler` filtering and sorting system) need to be fetched separately from channels. A dynamic list of available tabs is returned by `ChannelInfo` and `ChannelExtractor` with the method `getTabs`. The tab list is a list of `ListLinkHandler`s, which can be used to fetch tabs using the `getChannelTabExtractor` method.
In some cases, tab data is sent with the channel data, so fetching tabs would require an additional and duplicate request. To avoid this issue, a new custom `LinkHandler` can be used to create a `ChannelTabExtractor` instance with the data already fetched in it: `ReadyChannelTabListLinkHandler`.

Alternatively, a specific tab can be fetched using the corresponding `ChannelExtractor` methods taking a specific channel tab name (which should be one of the `ChannelTabs` constants) and returning channel tab extractors. In this case, `ReadyChannelTabListLinkHandler`s cannot be used, so duplicate requests could happen with this process; passing an invalid or unsupported tab name may return an exception indicating that the given tab name is invalid or unsupported.

Services' channel tab extractors, which so have `ChannelTabExtractor` as a base abstract class, are `ListExtractor` implementations allowing any `InfoItem` type. This results in the ability to provide streams, playlists and channels in channel tabs, and would allow to implement a posts item type in the future to support contents of YouTube's community tab for instance.

Channel tabs could be also fetched with their `Info` equivalent, `ChannelTabInfo`.

Note: Contrary to the original PR, channel tabs have their own package, `tabs`, a package of the `channel` one, in which `ChannelTabExtractor`, `ChannelTabInfo` and `ChannelTabs` classes are located. The `ReadyChannelTabListLinkHandler` class, which may need to be renamed, is still in the `linkhandler` package.

The following channel tabs are supported with this PR:

- for YouTube: Videos (tab's parameters have been updated to use the one provided and used by the desktop website), Shorts (by using only the channel tab except for age-restricted channels, see below), Live, Playlists and Channels;
- for SoundCloud: Tracks, Playlists and Albums;
- for MediaCCC: Videos;
- for PeerTube: Videos, Playlists and Channels;
- for Bandcamp: Tracks and Albums.

What is done with #951 but has been not highlighted is the support of channel tags: a list of strings which can be get using the `getTags` method of `ChannelInfo` and `ChannelExtractor`. Only YouTube seem to provide this information, so support has been added for this service only, other ones return an empty list, the default behavior.

I decided to do a little bit more than copying and improving changes of the original PR in this one: I also added support of detecting age-restricted YouTube channels and getting `StreamInfoItem`s of them using YouTube's system playlists. These playlists may contain geo-restricted (and so unavailable) videos. This only works when getting tabs from a channel, due to the usage of a `ReadyChannelTabListLinkHandler` and a `ChannelTabExtractor` implementation wrapping a `PlaylistExtractor` in this case. This support should be added when getting directly tabs, by moving the corresponding code as a fallback for stream items to YouTube's standard `ChannelTabExtractor`.

So, for these channels, the Videos, Shorts and Live tabs will be always returned, even if a tab or multiple tabs do or does not exist (it/they would return no items in this case, even if system playlists don't exist when a type of a content has been not created by a channel).
Note that support of the Shorts tab using system playlists is currently broken, as YouTube return a short UI for playlists with only shorts in most cases; this UI is currently not supported by the extractor. See #1077 for more details.

Tests have been of course updated to reflect the changes made in this PR, and the YouTube channels mocks had to be updated.

Closes #227, closes #501 (only support for topic channels which have regular tabs has been added, so system channels such as Music or Gaming return no tabs, this is tracked by #611), closes #585.

Note for reviewers: I recommend to review the changes commit by commit, as they are huge.

Debug APK built by @Stypox: [app-debug.zip](https://github.com/TeamNewPipe/NewPipeExtractor/files/12244479/app-debug.zip)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeChannelHelper.java
```diff
@@ -0,0 +1,271 @@
...
+
+        if (channelId[0].startsWith("UC")) {
+            return channelId[0];
+        }
+
+        // If the URL is not a /channel URL, we need to use the navigation/resolve_url endpoint of
+        // the InnerTube API to get the channel id.
+        // Otherwise, we couldn't get information about the channel associated with this URL, if
+        // there is one.
+        if (!channelId[0].equals("channel")) {
```
Stypox, line 49: ~~I would change this by putting `if (channelId[0].equals("channel")) { return channelId[1]; }` at the top and therefore de-indenting what now is in the `if`.~~ Oh, I just saw that this `if` can passthrough. Maybe you can just extend this comment to explain that if the `browseId` extraction fails then the unresolved id will be returned.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeChannelHelper.java
```diff
@@ -0,0 +1,271 @@
...
+                id = browseId;
+                level++;
+            } else {
+                ajaxJson = jsonResponse;
+                break;
+            }
+        }
+
+        if (ajaxJson == null) {
+            throw new ExtractionException("Got no channel response");
```
Stypox, line 180:
```suggestion
            throw new ExtractionException("Got no channel response after 3 redirects");
```

Stypox (changes requested):
Partial review; I had some time to look at the first few commits (up until the YouTube one, where I stopped at `YoutubeChannelExtractor`)

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeChannelHelper.java
Stypox: Also, I think there is an extra "such as" in the javadoc of this function

extractor/src/test/java/org/schabi/newpipe/extractor/ExtractorAsserts.java
```diff
@@ -144,6 +148,16 @@ public static void assertContains(
         assertNotNull(shouldBeContained, "shouldBeContained is null");
         assertNotNull(container, "container is null");
         assertTrue(container.contains(shouldBeContained),
-                "'" + shouldBeContained + "' should be contained inside '" + container +"'");
+                "'" + shouldBeContained + "' should be contained inside '" + container + "'");
+    }
+
+    public static void assertTabsContained(@Nonnull final List<ListLinkHandler> tabs,
```
Stypox, line 154: I would call it `assertTabsContain`

extractor/src/test/java/org/schabi/newpipe/extractor/services/peertube/PeertubeChannelTabExtractorTest.java
```diff
@@ -0,0 +1,212 @@
...
+        @Override
+        public void testOriginalUrl() throws Exception {
+            assertEquals("https://framatube.org/accounts/framasoft/video-channels",
+                    extractor.getOriginalUrl());
+        }
+
+        @Test
+        @Override
+        public void testRelatedItems() throws Exception {
+            defaultTestRelatedItems(extractor);
```
Stypox, line 203: Maybe the `defaultTest(Related|More)Items` should take a second parameter that would allow making sure all items are of a specific type (e.g. in this case `ChannelInfoItem`)

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeChannelExtractorTest.java
```diff
@@ -737,5 +724,145 @@ public void testSubscriberCount() throws Exception {
         public void testVerified() throws Exception {
             assertTrue(extractor.isVerified());
         }
+
+        @Test
+        @Override
+        public void testTabs() throws Exception {
+            assertTabsContained(extractor.getTabs(), ChannelTabs.VIDEOS, ChannelTabs.SHORTS,
```
Stypox, line 731: Maybe also assert that the videos tab has already been fetched?

extractor/src/main/java/org/schabi/newpipe/extractor/linkhandler/ReadyChannelTabListLinkHandler.java
```diff
@@ -0,0 +1,55 @@
...
+ * following snippet of code in the service's
+ * {@link StreamingService#getChannelTabExtractor(ListLinkHandler)}:
+ * <pre>
+ * if (linkHandler instanceof ReadyChannelTabListLinkHandler) {
+ *     return ((ReadyChannelTabListLinkHandler) linkHandler).getChannelTabExtractor(this);
+ * }
+ * </pre>
+ * </p>
+ */
+public class ReadyChannelTabListLinkHandler extends ListLinkHandler {
```
Stypox, line 32: Did we decide on a better name or not in the end?

Stypox (changes requested):
There are a few small things to change, but overall everything is fine. If you don't have objections, I can fix everything myself and then merge. Thank you again to you and Theta-Dev for the effort on this PR!

Btw, for the future try to keep the commit subject under 50 characters: https://stackoverflow.com/q/2290016

Stypox (approved):
~~Aha! I took back ownership of commits!~~ I rebased and tested locally, everything works fine. I will open a PR for the small suggestions I proposed above, as they are not blocking.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelTabExtractor.java
```diff
@@ -0,0 +1,486 @@
...
+     *
+     * <p>
+     * A valid {@code visitorData} is required to get continuations of shorts in channels.
+     * </p>
+     *
+     * <p>
+     * It should be not used when it is not needed, in order to reduce YouTube's tracking.
+     * </p>
+     */
+    private final boolean useVisitorData;
```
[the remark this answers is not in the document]
Stypox: Nevermind, scrap this one

PR #1086, opened 2023-08-01
[YouTube] Don't throw an exception when there is no banner available on a channel

AudricV:
      I tested my changes by using a temporary unit test with the channel from which the issue has been found and another one with which the issue was reproducible (see below).

On YouTube, channels may not have a banner, so no exception should be thrown if no banner is found. This PR makes `YoutubeChannelExtractor.getBannerUrl` returning `null` when there is no banner instead of throwing a `ParsingException`, like it was the case before (the current behavior has been introduced in #1050).

As `carouselHeaderRender`s do not contain a banner, I tested this behavior in the corresponding test method and test class (`testBannerUrl` in `YoutubeChannelExtractorTest.CarouselHeader`).

This bug has been originally reported in [this comment in the issue of NewPipe v0.25.2's release](https://github.com/TeamNewPipe/NewPipe/issues/10284#issuecomment-1659726774).


chirag3103:
Build and test is failing on CI. Also you may have to explicitly add reviewers on your PR to get approvals

AudricV:
The test failures are unrelated to this PR (Bandcamp service) and are already fixed by #1082.

About reviews, team members working on the extractor look at PRs without having to request a review.

TobiGr (approved):

PR #1087, opened 2023-08-02
[YouTube] Improve and fix YoutubeJavaScriptExtractor

AudricV:

This PR improves and fixes a part of the extraction of YouTube's JavaScript base player file, by doing the following changes:

- Enhance the class' documentation;
- Fix the regular expression fallback on HTML embed watch page;
- Use HTML scripts tag search first instead of the regular expression approach, now used as a last resort;
- Compile regular expressions only once, in order to improve the performance of subsequent extraction calls when clearing the cache;
- Provide original exceptions when fetching or parsing pages on which the base JavaScript's player could be found failed, allowing clients to detect network errors when they are the cause of the failures for instance;
- Remove delegate method which was not taking a video ID and hardcoding one, as we can provide the video ID in all cases or do not provide a video ID at worse;
- Rename and make extraction methods package-private, as they are not intended to be used publicly.

The breaking changes introduced, which are internal to the extractor (the methods changed were not expected to be used by clients), have been applied where needed, in `YoutubeJavaScriptExtractorTest` and `YoutubeStreamExtractor` (in which an unneeded `initStsFromPlayerJsIfNeeded` call has been removed).

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeJavaScriptExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeJavaScriptExtractor.java
@@ -10,40 +10,62 @@ import org.schabi.newpipe.extractor.localization.Localization;
 import org.schabi.newpipe.extractor.utils.Parser;
 
 import javax.annotation.Nonnull;
+import java.util.regex.Pattern;
 
 /**
- * YouTube restricts streaming their media in multiple ways by requiring clients to apply a cipher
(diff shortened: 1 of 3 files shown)


AudricV:
Requesting a review from TheAssassin, in order to be sure that the wording around YouTube's IFrame API resource does not mean we use the YouTube IFrame API by itself and so that we agreed to YouTube API Terms of Service (the resource is used by external websites and even on YouTube static pages such as [How YouTube works](https://www.youtube.com/howyoutubeworks/)).

Stypox (approved):

PR #1088, opened 2023-08-03
Replace cryptographically secure random with regular random

FireMasterK:

There's no need for a cryptographically secure random, even though the JavaScript code uses one.


Stypox:
Out of curiosity, did you notice this because of performance hiccups?

FireMasterK:
> Out of curiosity, did you notice this because of performance hiccups?

Yes, I did notice it in my profiler. It wasn't necessarily a hiccup, but an improvement I saw could be helpful. `SecureRandom` seems to do blocking IO according to my profiler. (not sure why)

Stypox:
I guess the underlying implementation uses `/dev/urandom`, which is based on an entropy pool, which needs to fill up from various noise sources and sensors

Stypox (approved):

PR #1092, opened 2023-08-06
Improve channel tabs code

Stypox:
I implemented my suggestions from #1082 here, read the commit subjects for specific info. This completes the channel tabs PR, but can be merged with less priority.

I basically rewrote from scratch the channel tab tests, as you can simply inherit `DefaultListExtractorTest` and let the default implementation do the rest. All tests still pass. Furthermore, I split the age restricted channel tabs test into two tests, although this implies the common channel page is fetched twice.

We still need to decide whether we want to rename `ReadyChannelTabListLinkHandler` to something else. tldr it's a class that is used as a temporary ListLinkHandler which can only be instantiated internally (while other link handlers are instantiatable by anyone) in the extractor for specific purposes, e.g. to prevent additional fetches of data in channel tab extractors when such data has already been extracted in the channel extractor.

Let me know if https://github.com/TeamNewPipe/NewPipeExtractor/pull/1092/commits/e34739e1999ed1269c54b93d85c4cb4fa0c66aa3 is ok in your opinion. There is now a "Raw use of parameterized class 'InfoItemsPage'" warning, but it can be ignored since the type missing would be `<InfoItem>`, and `StreamInfoItem extends InfoItem`.


AudricV:
> Let me know if [e34739e](https://github.com/TeamNewPipe/NewPipeExtractor/commit/e34739e1999ed1269c54b93d85c4cb4fa0c66aa3) is ok in your opinion. There is now a "Raw use of parameterized class 'InfoItemsPage'" warning, but it can be ignored since the type missing would be `<InfoItem>`, and `StreamInfoItem extends InfoItem`.

Shouldn't you apply the same changes for next pages too?

Stypox:
> Shouldn't you apply the same changes for next pages too?

Right, my bad. Fixed.

AudricV:
> Right, my bad. Fixed.

The commit message isn't though.

Stypox:
> The commit message isn't though.

Fixed

AudricV (approved):
Thank you for these improvements and for this one line saving!

PR #1093, opened 2023-08-07
[YouTube] Support Shorts UI in playlists

### LINKED ISSUE #1077: [YouTube] Cannot get streams of playlists with only shorts

AudricV:
  I only used my test class on which no errors occurred during my tests.

This PR fixes the extraction of items in YouTube playlists returning a Shorts UI.

A new test in `YoutubePlaylistExtractorTest` has been added to test the extraction of items with this playlist UI, which uses the system playlist of uploaded shorts by the official YouTube channel (`UUSHBR8-60-B28hp2BmDPdntcQ`).

As one of the effects of YouTube using a shorts UI, no uploader info is available for each stream. The default tests expected that the uploader name was set, so I had to remove this check (I also removed the uploader URL one, as it was commented out).

Finally this PR removes an outdated A/B test comment in `YoutubePlaylistExtractor` and fixes a bug in the `assertNotEmpty` test method of the extractor's assertions class (`ExtractorAsserts`) which tested that the test failure message, annotated as nullable, was not null instead of the string to check (it was the initial cause of the stream items' tests failures of my test class added with this PR, as no uploader name was provided and a `NullPointerException` was thrown).

Fixes #1077.


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubePlaylistExtractor.java
```diff
@@ -327,17 +326,22 @@ public InfoItemsPage<StreamInfoItem> getInitialPage() throws IOException, Extrac
                 .map(content -> content.getObject("itemSectionRenderer")
                         .getArray("contents")
                         .getObject(0))
-                .filter(contentItemSectionRendererContents ->
-                        contentItemSectionRendererContents.has(PLAYLIST_VIDEO_LIST_RENDERER)
-                                || contentItemSectionRendererContents.has(
-                                "playlistSegmentRenderer"))
```
Stypox, line 333: Nice catch that "playlistSegmentRenderer" wasn't really used even before!

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubePlaylistExtractor.java
```diff
@@ -399,14 +403,26 @@ private Page getNextPageFrom(final JsonArray contents)
...
+                        if (richItemRenderer.has("content")) {
+                            final JsonObject richItemRendererContent =
+                                    richItemRenderer.getObject("content");
+                            if (richItemRendererContent.has(REEL_ITEM_RENDERER)) {
+                                collector.commit(new YoutubeReelInfoItemExtractor(
+                                        richItemRendererContent.getObject(REEL_ITEM_RENDERER),
+                                        timeAgoParser));
+                            }
+                        }
+                    }
```
Stypox, line 424: Shouldn't an error be added to the collector to signal that an item in the list was in an unrecognized format? Or is there other garbage inside the list that just needs to be ignored?

Stypox (changes requested):
LGTM, thanks! Just one thing to check

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubePlaylistExtractor.java
AudricV: It seems it was used for seasonal playlists: see https://github.com/TeamNewPipe/NewPipeExtractor/commit/8b3f90eb7e4a56ee7e33128af7cf8dd94663414a.

I don't remember now why I have added this though.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubePlaylistExtractor.java
AudricV: I am not sure, as there should be playlists video items only but YouTube return(ed?) suggestions of videos to add to users playlists when their owner(s) browse the playlists and is(are) logged with their account(s), so I don't know what elements they use in this case.

Stypox: Ok let's leave it this way and if users report of missing item take a look again

Stypox (approved):

PR #1094, opened 2023-08-08
[YouTube] Support more channel headers

AudricV:
  I only used my test class on which no errors occurred during my tests and a temporary one to test a part of my changes (see why below).

This PR adds support of more YouTube channel headers, the `pageHeaderRenderer` and `interactiveTabbedHeaderRenderer` ones, fixing crashes when extracting channels with these headers.

The first header is currently A/B tested (that's the reason why no test has been added it) on system topic channels such as the Gaming one and renders in the following way on the desktop web client:

PageHeaderRenderer YouTube desktop website

It only returns the info you can see (channel's name and its avatar).

The second header is used for gaming topic channels and only provides the channel's name, its banner and a poster as its "avatar".

A new test in `YoutubeChannelExtractorTest` has been added for this header, which uses the Minecraft system topic channel (`UCQvWX73GQygcwXOTSf_VDVg`).

Even if the support of these headers has been added, the tabs of the (system) channels on which the headers are returned are not supported.

Another header is also returned when you try to access some system topic channels with an invalid tab for these channels, such as the Gaming one with the Videos tab (it looks like a bug on YouTube's side): `feedTabbedHeaderRenderer`, which is almost empty and only contains the channel name (the response doesn't seem to render anything on the website). Its support is not added with this PR.

I also made a few other changes related to the headers support, which are described in the commits' message.


Stypox (approved):
Thank you! I pushed a small commit simplifying Optional call chains, since [`Optional.map()`](https://docs.oracle.com/en/java/javase/14/docs/api/java.base/java/util/Optional.html#map(java.util.function.Function)) accepts `null` values just fine (actually you did so in one place but in the other you used `flatMap` with `Optional.ofNullable`). I also included all cases in the `switch` statements so that it is clear for which header kinds the default branch is taken.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeChannelExtractor.java
```diff
@@ -158,53 +161,105 @@ public String getName() throws ParsingException {
...
+            if (header.headerType == HeaderType.PAGE) {
+                // No banner is available on pageHeaderRenderer headers
+                return null;
+            }
+
+            return Optional.ofNullable(header.json.getObject("banner")
+                    .getArray("thumbnails")
+                    .getObject(0)
+                    .getString("url"))
+                    .filter(url -> !url.contains("s.ytimg.com") && !url.contains("default_banner"))
```
Stypox, line 254: Why are you not returning a banner at all if the url contains `s.ytimg.com` or `default_banner`? This seems to date back to the initial commit of the extractor: [see here](https://github.com/AudricV/NewPipeExtractor/blame/7581c1200b7795d8fbb0769930f5981a962e347b/services/youtube/YoutubeChannelExtractor.java#L142)

AudricV: I wanted to keep the original behavior, that's the reason I didn't change this. Like you said, that's indeed useless.

PR #1095, opened 2023-08-12
[YouTube] Attempt to extract function from player only once

FireMasterK:


AudricV:
Why not storing a boolean indicated whether the code has been successfully loaded instead? We would still need to reset it where the JavaScript code is reset (the missing part of your PR), so it seems we have a design issue.

FireMasterK:
> Why not storing a boolean indicated whether the code has been successfully loaded instead? We would still need to reset it where the JavaScript code is reset (the missing part of your PR), so it seems we have a design issue.

The problem is we're failing to extract the function, rather than the code. When we extract the code, we also try to extract the function. If the code has been extracted, but the function has not, we have likely failed to extract the function in the past.

I can add a function to reset the JS extractor code.

PR #1104, opened 2023-09-17
[YouTube] Workaround Shorts UI for playlists by using a continuation for initial items

### LINKED ISSUE #1273: [YT] Playlists that only contains shorts/ShortsUI does not return any items
### Checklist

- [x] I am able to reproduce the bug with the latest version given here: [CLICK THIS LINK](https://github.com/TeamNewPipe/NewPipeExtractor/releases/latest).
- [x] I am aware that this issue is being opened for the NewPipe Extractor, NOT the [app](https://github.com/TeamNewPipe/NewPipe), and my bug report will be dismissed otherwise.
- [x] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipeExtractor/issues) or [closed](https://github.com/TeamNewPipe/NewPipeExtractor/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [x] This issue contains only one bug.
- [x] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Affected version

0.24.5

### Steps to reproduce the bug

Playlists that only contains YouTube shorts, e.g. https://www.youtube.com/playlist?list=UUSHBR8-60-B28hp2BmDPdntcQ does not return any items:

Image

This currently breaks ``YoutubePlaylistExtractorTest.ShortsUI#testRelatedItems``

AudricV:

This PR replaces fetching playlist items with an initial playlist continuation, in order to workaround the Shorts UI in playlists, restricting access to only the 100th first items, as no continuation is currently provided. Reel items returned don't also give upload date, uploader info and precise view count.

This required to add a protocol buffer definition file, for which its structure is based on reverse engineering of playlists continuations sent by WEB InnerTube client, received from InnerTube responses.

Java classes of this file are generated for the Java Lite runtime of Protobuf with the Protobuf Gradle plugin, as the lite version is enough for our use cases. This plugin ships in JARs Protobuf definitions, which should be avoided (I didn't found something for NewPipeExtractor working on assemble Gradle tasks, only for NewPipe).

As Protobuf generated classes are parsed by Checkstyle and do not follow our style rules at all, an exclusion rule has been for them.

As a result of this change, `testMoreRelatedItems` method of `YoutubePlaylistExtractorTest.ShortsUI` test class has been enabled.

Fixes #1273
Fixes TeamNewPipe/NewPipe#12341


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubePlaylistExtractor.java
```diff
@@ -64,17 +70,40 @@ public YoutubePlaylistExtractor(final StreamingService service,
...
+
+        final PlaylistContinuation playlistContinuation = PlaylistContinuation.newBuilder()
+                .setParameters(ContinuationParams.newBuilder()
+                        .setBrowseId("VL" + playlistId)
+                        .setPlaylistId(playlistId)
+                        .setContinuationProperties(PLAYLIST_CONTINUATION_PROPERTIES_BASE64)
+                        .build())
+                .build();
+
+        initialBrowseContinuationResponse = getJsonPostResponse("browse",
```
Stypox, line 99: I guess there is no way to avoid doing two network requests, right? Doesn't initialBrowseContinuationResponse also contain the metadata?

Stypox (commented):
Have you tried this to exclude the *.proto from the .jar? https://github.com/google/protobuf-gradle-plugin/issues/390

Other than that, code looks good to me. Thanks!

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubePlaylistExtractor.java
AudricV: Only legacy data is present (which is what we use almost every time, as on most playlists YouTube switched to view models, we do not support this data). I can also filter the continuations to only return videos, but this add more detection possibilities for YouTube.

Stypox: I do not understand, so you mean we could only do one request? Then I'd suggest doing only one call even if it means slightly degraded data (although it depends to which degree it's degraded)

Stypox: > I suspect YouTube will remove it at some point and I think the fact it is returned in all continuations is more a bug than something else.

Okok then let's not use and be future proof

> I don't think we should be worried on this adding more detection possibilities for YouTube on metadata endpoints

True

Stypox:
I rebased and fixed *.proto ending up in the JAR.

Before:

image

After:

image

PR #1106, opened 2023-09-18
Obtain stream length as a Duration

Isira-Seneviratne:


extractor/src/main/java/org/schabi/newpipe/extractor/localization/TimeAgoParser.java
```diff
@@ -68,45 +65,37 @@ public DateWrapper parse(final String textualDate) throws ParsingException {
...
-                .orElseThrow(() -> new ParsingException(
-                        "Could not parse duration \"" + textualDuration + "\""));
+            try {
+                duration = duration.plus(amount, parseChronoUnit(word));
+            } catch (final ParsingException ignored) {
+            }
+        }
+        if (duration.isZero()) {
+            throw new ParsingException("Could not parse duration \"" + textualDuration + "\"");
+        }
```
XiangRongLin, line 97: Can't `duration` not be uninitialized at the start and if it is still not initialized then throw the exception

XiangRongLin (commented):
I don't know what the policy around breaking changes is anymore, otherwise it looks fine. But it should be at least mentioned that its a breaking change in the PR itself

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamInfoItem.java
```diff
@@ -40,7 +42,7 @@ public class StreamInfoItem extends InfoItem {
     @Nullable
     private DateWrapper uploadDate;
     private long viewCount = -1;
-    private long duration = -1;
+    private Duration duration = Duration.ZERO;
```
TobiGr, line 45: Should the default value be `0` or `null`? There are valid points for both, e.g. null safety, a valid duration being > 0. ... The default value should be documented in the corresponding getters and setters.

Stypox: I think 0 is fine. We have been using 0 as an integer to indicate "no duration available", so the same can do for Duration, too.

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamInfoItem.java
```diff
@@ -76,11 +78,15 @@ public void setViewCount(final long viewCount) {
         this.viewCount = viewCount;
     }
 
-    public long getDuration() {
+    public Duration getDuration() {
```
TobiGr, line 81: Should be annotated as `@Nonnull`

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamInfoItem.java
```diff
@@ -76,11 +78,15 @@ public void setViewCount(final long viewCount) {
...
         return duration;
     }
 
-    public void setDuration(final long duration) {
+    public long getDurationInSeconds() {
+        return duration.toSeconds();
+    }
+
+    public void setDuration(final Duration duration) {
```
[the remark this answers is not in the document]
TobiGr: Param should be annotated as `@Nonnull`

Isira-Seneviratne:
I made some changes, this should no longer break existing uses of the methods.

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamInfoItem.java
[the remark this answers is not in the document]
Isira-Seneviratne: I've made some changes to the PR, it will no longer break existing applications.

Stypox:
I don't know if this PR is a good idea. While it's true that the correct data type would be `Duration`, just returning the plain old number of seconds also makes sense in my opinion. However, with `Duration` it would be possible to return even millisecond-precision durations, which is an advantage.

What do others think? Is it worth the effort to switch to `Duration` right now? Maybe it's better to do it if we ever will want to refactor the interfaces completely (in which case we wouldn't have to worry about breaking changes), but keeping track of deprecated methods and composite data types just adds complexity without clear advantages in this case. Users of the library can just use `Duration.ofSeconds()` if they really need a `Duration`.

E.g. in NewPipe we would need to change the database to properly adapt to this PR, which feels like a risky move without any advantage, which means in NewPipe we wouldn't really switch to using `Duration`, which makes this PR not useful (at least for NewPipe).

PR #1108, opened 2023-09-18
[YouTube] Refactor JavaScript usage and fix extraction of obfuscated signature deobfuscation function

AudricV:

This PR refactors the usage of YouTube's JavaScript base player file in the extractor.

It makes a single public class handling operations and caches of this file, `YoutubeJavaScriptPlayerManager`, instead of being split in multiple files with questionable designs (such as the one of the deobfuscation function for obfuscated signatures of streaming URLs from HTML5 clients in `YoutubeStreamExtractor`).

Extraction and parsing parts are delegated to different package-private classes:

- `YoutubeJavaScriptExtractor`, handling the extraction of JavaScript's base player file;
- `YoutubeThrottlingParameterUtils`, handling the extraction of throttling parameter of streaming URLs from HTML5's clients and the detection of this parameter for a given streaming URL;
- `YoutubeSignatureUtils`, handling the extraction of the signature timestamp property and the deobfuscation function for obfuscated signatures of streaming URLs from HTML5 clients.

Each extraction process has now a corresponding test.

It also improves the code on the fly, fixes the extraction of the deobfuscation function for obfuscated signatures, and makes use of a number when sending the signature timestamp in InnerTube's player requests instead of a string, in order to be consistent with HTML5 clients.

**This PR introduces important breaking changes**:

- `YoutubeJavaScriptExtractor` is not public anymore, the extraction of YouTube's JavaScript base player file isn't intended to be used outside of the extractor;
- `resetDeobfuscationCode` method of `YoutubeStreamExtractor` has been removed. Use the method `clearAllCaches` of `YoutubeJavaScriptPlayerManager` instead, which also clears other cached data;
- `YoutubeThrottlingDecrypter` has been renamed to `YoutubeThrottlingParameterUtils`, which is a package-private class and the methods `getCacheSize` and `clearCache` have been removed. Use respectively `getThrottlingParametersCacheSize` and `clearThrottlingParametersCache` methods of `YoutubeJavaScriptPlayerManager` instead.
- `DeobfuscateException` nested exception class in `YoutubeStreamExtractor` has been removed. A standard `ParsingException` will be thrown when the deobfuscation function couldn't be parsed or executed instead.

Fixes TeamNewPipe/NewPipe#10347.


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeJavaScriptPlayerManager.java
```diff
@@ -0,0 +1,334 @@
...
+        extractJavaScriptCodeIfNeeded(videoId);
+
+        try {
+            cachedSignatureTimestamp = Integer.valueOf(
+                    YoutubeSignatureUtils.getSignatureTimestamp(cachedJavaScriptPlayerCode));
+        } catch (final ParsingException e) {
+            // Store the exception for future calls of this method, in order to improve performance
+            sigTimestampExtractionEx = e;
+            throw e;
+        }
```
Stypox, line 102: What if `Integer.valueOf` fails? Can't you just catch any exception with `Throwable` in all places?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeSignatureUtils.java
```diff
@@ -0,0 +1,150 @@
...
+                    Parser.matchGroup1(SIG_DEOBF_HELPER_OBJ_NAME_REGEX, deobfuscationFunction);
+
+            final String helperObject = getHelperObject(javaScriptPlayerCode, helperObjectName);
+
+            final String callerFunction = "function " + DEOBFUSCATION_FUNCTION_NAME
+                    + "(a){return "
+                    + deobfuscationFunctionName
+                    + "(a);}";
+
+            return helperObject + deobfuscationFunction + callerFunction;
```
Stypox, line 97: I would add the semicolons here, instead of inside `getHelperObject` and `getDeobfuscateFunctionWithRegex`, for clarity. If the regex in `getHelperObject` needs the `;`, then you can keep it, and add it here anyway, a redundant `;` is valid JS and prevents future programming errors when e.g. changing the regex or the functions.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeSignatureUtils.java
```diff
@@ -0,0 +1,150 @@
...
+    };
+
+    private static final String STS_REGEX = "signatureTimestamp[=:](\\d+)";
+
+    private static final String DEOBF_FUNC_REGEX_START = "(";
+    private static final String DEOBF_FUNC_REGEX_END = "=function\\([a-zA-Z0-9_]+\\)\\{.+?\\})";
+
+    private static final String SIG_DEOBF_HELPER_OBJ_NAME_REGEX = ";([A-Za-z0-9_\\$]{2,})\\...\\(";
+    private static final String SIG_DEOBF_HELPER_OBJ_REGEX_START = "(var ";
+    private static final String SIG_DEOBF_HELPER_OBJ_REGEX_END = "=\\{(?>.|\\n)+?\\}\\};)";
```
Stypox, line 38: Why don't these need `@SuppressWarnings("RegExpRedundantEscape")`? They also use an escaped `}`

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeSignatureUtils.java
```diff
@@ -0,0 +1,150 @@
...
+        return functionBase + JavaScriptExtractor.matchToClosingBrace(
+                javaScriptPlayerCode, functionBase) + ";";
+    }
+
+    @Nonnull
+    private static String getDeobfuscateFunctionWithRegex(
+            @Nonnull final String javaScriptPlayerCode,
+            @Nonnull final String deobfuscationFunctionName) throws ParsingException {
+        final String functionPattern = DEOBF_FUNC_REGEX_START
+                + deobfuscationFunctionName.replace("$", "\\$")
```
Stypox, line 135: Here and in `getHelperObject` you used `.replace($, \$)`, while in `YoutubeThrottlingParameterUtils.getDeobfuscationFunctionName` you used `Pattern.quote()`. Is there a reason for this? I think `replace` is just fine, assuming that the variable name contains just `a-zA-Z0-9_$`, which is what the regexes enforce. (in theory js variables can be [as stupid as you want](https://stackoverflow.com/q/1661197))

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingParameterDeobfuscationTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+        for (final String videoId : videoIds) {
+            try {
+                final String deobfuscatedUrl =
+                        YoutubeJavaScriptPlayerManager.getUrlWithThrottlingParameterDeobfuscated(
+                                videoId, obfuscatedUrl);
+                assertNotEquals(obfuscatedUrl, deobfuscatedUrl);
+            } catch (final EvaluatorException e) {
+                fail("Failed to extract throttling parameter deobfuscation function for video "
+                        + videoId + "\n" + e);
```
Stypox, line 38: Won't this hide the exception stacktrace? If you just want to call `fail()` when the two urls are equal, just do `if (obfuscatedUrl.equals(obfuscatedUrl)) { fail(...) }`

Stypox (changes requested):
Thanks for this! The code is so much better and well organized. I pointed out a few small things I notices

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeJavaScriptPlayerManager.java
AudricV: > What if `Integer.valueOf` fails?

I need to handle at least `NumberFormatException`s here indeed, good catch!

> Can't you just catch any exception with `Throwable` in all places?

Is it really a good idea to catch all errors, including ones which could come from the JVM (i.e. instances of the `Error` class)? I don't think so, but I am not sure.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeSignatureUtils.java
AudricV: Done, if you meant `getDeobfuscateFunctionWithLexer` and `getDeobfuscateFunctionWithRegex`.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeSignatureUtils.java
AudricV: My IDE didn't complain about that, and I didn't disable the corresponding warning. Is a warning shown for you or someone else?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeSignatureUtils.java
AudricV: The `.replace($, \$)` calls have been not written by me, they have been introduced in the initial commit of the extractor. I don't know the reason.

I would personally use `Pattern.quote` to avoid any issues in the future, in the case YouTube starts to use weird names.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingParameterDeobfuscationTest.java
AudricV: I didn't write this code, but I think we can just use the [`fail` method accepting a `Throwable` as a second argument](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#fail(java.lang.String,java.lang.Throwable)) and remove the line break instead.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeSignatureUtils.java
Stypox: Yeah don't worry then, it was just to make sure

Stypox (approved):
Thank you!

Stypox:
I can confirm it works, I tested in NewPipe

PR #1113, opened 2023-10-01
[SoundCloud] Fix extraction of non-JPG images

AudricV:

Default image qualities were removed in image URLs with the `jpg` extension, causing the addition of the image suffix to full non-JPG images URLs and so to invalid image URLs.

Only the image quality name with its leading `-` character and the `.` character after the name is now removed and replaced by a string format replaced itself with the image quality name for each quality.

As the image suffixes do not contain the image extension, the name of image qualities lists has been adapted with these changes and some related comments have been also improved.

The documentation of the `ImageSuffix` class has been also updated in order to reflect the change of the non-requirement of the extension in the suffix property introduced by this fix.


Stypox (approved):
Thanks!

PR #1135, opened 2023-12-07
[YouTube] Implement emergency meta info

Stypox:

As seen in #1127, YouTube provides that meta info panel when users search for really sensitive content like suicide (e.g. "blue whale").

image

It contains:
- an encouragement as title (e.g. "Siamo con te" / "We are with you")
- a phone number as action (e.g.  "Chiama ..." / "Call ...")
- details about how to call the phone number (e.g. in which time ranges the phone number can be called)
- an url pointing to the website of an association (e.g. "Samaritans Onlus")

Since NewPipe's meta infos contain only one details field, I concatenated the phone number on a new line after the phone number details.

image

This is how it looks in NewPipe:

image

Also add a test that just checks if a meta info is properly extracted. Since this test depends on #1127, this PR includes those commits, too. I will cherry-pick once #1127 is merged.

Also move meta info extraction to separate file, since YoutubeParsingHelper was longer than 2000 lines which caused checkstyle issues.


TobiGr (approved):
LGTM

PR #1143, opened 2023-12-26
Avoid PeerTube accepting non-URLs

petlyh:
No NewPipe changes are needed.

This PR makes the PeerTube `onAcceptUrl` methods return `false` if the input is not a valid URL.
This is done by passing the input to the `URL` constructor and returning `false` if a `MalformedURLException` gets caught.

The PeerTube `onAcceptUrl` methods currently only checks whether the input contains path parts like `/videos/` or `/a/`.
This causes problems when they get called when NewPipes `SearchFragment.search` checks whether the users search query is an accepted URL, since non-URL search queries can contain those parts. TeamNewPipe/NewPipe#10695 is caused by the search query `orchestr/a/` being interpreted as a PeerTube Account URL because it contains `/a/`.

Fixes TeamNewPipe/NewPipe#10695


TobiGr:
Thank you! Can you also add test cases to ensure we won't introduce this unwanted behaviour again?

TobiGr (approved):
Thanks. I added the tests

extractor/src/test/java/org/schabi/newpipe/extractor/services/peertube/PeertubeLinkHandlerFactoryTestHelper.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.schabi.newpipe.extractor.exceptions.ParsingException;
+import org.schabi.newpipe.extractor.linkhandler.LinkHandlerFactory;
+import org.schabi.newpipe.extractor.linkhandler.ListLinkHandlerFactory;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+public class PeertubeLinkHandlerFactoryTestHelper {
+
+    public static void testDoNotAcceptNonURLs(LinkHandlerFactory linkHandler)
```
Stypox, line 11: These should be called `assert...` instead of `test...`, but anyway

TobiGr: Done: 9ab932e394f31a6687014d69d241dc1a56e4ee18

Stypox: Thanks!

PR #1148, opened 2023-12-30
[MediaCCC] Allow obtaining channel tab link handler

Stypox:
- [x] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe): TeamNewPipe/NewPipe#10717

This is a proper fix for https://github.com/TeamNewPipe/NewPipe/issues/10698. Now it is possible to obtain a channel tab link handler, and `null` is never returned. The comments scattered around the code should explain my approach better than I could do in the PR description ;-), but the important point is that I reused the existing channel link handler for channel tabs, too.

Fixes `TransactionTooLargeException` when switching away from NewPipe while there are MediaCCC channels open. Before I could reproduce this consistently, while with this PR the crash is fixed.

The fact that the channel and the channel tab have the same url made it so that NewPipe would cache them both under the same key. I opened TeamNewPipe/NewPipe#10717 to fix that, and you can try the APK there to test these changes.


TobiGr:
I guess this is a fix for https://github.com/TeamNewPipe/NewPipe/issues/10714, isn't it?

Stypox:
Probably it is, although the log poster by the user was fixed in your hotfix PR

TobiGr:
Ah, I just saw that the user listed 0.26.1 as affected version, but the logs indicate that the crash was happening on 0.26.0...

Issue #16, opened 2017-08-01
Testing and decouple extractor


mauriciocolli:
As @TheAssassin said [here](https://github.com/TeamNewPipe/NewPipeExtractor/pull/15#issuecomment-319029514), I want to see the tests working too.

I'm thinking in doing this:
- Move the unit tests to here
- Make this repository "buildable" with gradle
- Use [jitpack.io](https://jitpack.io) to import it to NewPipe as a dependency (I think that @theScrabi said that he wanted to include in maven right? For now, jitpack will be enough)

Currently, doesn't make much sense having the tests there, a big reason is:
- For each commit that is done there, the unit tests start **ALL** again testing this repository **that remains intact**, we are just repeating the tests without changing anything

Another reason would be that is easier to modify the extractor or include in another project, as it'll be a totally independent module, for now, it is a little bit coupled with NewPipe repository...

What do you guys think? If you have suggestions, feel free to let me know.

wb9688:
Well, I was thinking about doing the same thing after my PR is merged. I'll
also have to write unit tests for my service.

wb9688:
@mauriciocolli: The tests depends on [`Downloader.java`](https://github.com/TeamNewPipe/NewPipe/blob/master/app/src/main/java/org/schabi/newpipe/Downloader.java), what should I do with that?

theScrabi:
@mauriciocolli 
> I think that @theScrabi said that he wanted to include in maven right? For now, jitpack will be enough
Yes that was the idea to also put the backed into a maven repo.
jitpack.io is nice :)

PR #1149, opened 2024-01-04
[YouTube] Add trending position to StreamInfoItem

FineFindus:

Adds support for getting the trending position of the video, e.g. `#16 on Trending`.
Image showing the trending tag of a video

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
@@ -509,6 +509,15 @@ public final class YoutubeParsingHelper {
         }
     }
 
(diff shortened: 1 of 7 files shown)


AudricV:
Sorry, but unless there is an approach which allows getting this data from the stream items directly, we will not accept this PR, as we cannot fetch HTML pages or video metadata responses for each stream in a stream list, due to resources consumption and the increase of chances to flag the IP used by the extractor as an abusive one (which would result for instance to captchas and/or IP bans on videoplayback and/or InnerTube player endpoint).

FineFindus:
Well, that's sad to hear :(, but understandable. This is unfortunately the only occurrence of this data on the website, it's not sent in another API response. This may be different for other platforms, but I haven't looked into extracting data from them yet. If someone finds a better way to get the data, I'll be happy to work on it again. Closing it for now.

PR #1150, opened 2024-01-06
AudioStream: remove unused youtube-specific fields

Profpatsch:
These fields were added in
https://github.com/TeamNewPipe/NewPipeExtractor/pull/810 in preparation for a DASH support that never materialized … well it did materialize, but did not use any of these variables.

Since they are youtube-specific, it does not make much sense to export them in the generic API, lest people start depending on them as if they belonged to all backends.

Well, 4 variables are actually already depended on in places, they will have to be checked separately. But this is the low-hanging fruit.


extractor/src/main/java/org/schabi/newpipe/extractor/stream/AudioStream.java
```diff
@@ -33,16 +33,6 @@ public final class AudioStream extends Stream {
...
 
-    // Fields for DASH
-    private int itag = ITAG_NOT_AVAILABLE_OR_NOT_APPLICABLE;
-    private int bitrate;
-    private int initStart;
-    private int initEnd;
-    private int indexStart;
-    private int indexEnd;
-    private String quality;
-    private String codec;
```
FireMasterK, line 44: Hi, I was the one who added these fields originally, they are used in Piped's backend for providing sidX boxes for streaming with DASH.

Here's how an audio representation is created by a client: https://github.com/TeamPiped/Piped/blob/8ee2a1c20737ce9db84186e00849b405d6c097af/src/utils/DashUtils.js#L99-L117

TobiGr:
Closing to not brake Piped

PR #1151, opened 2024-01-07
LocaleCompat.forLanguageTag: return Optional if parsing fails

Profpatsch:
It’s not obvious that the function will fail in some cases and throw an `IllegalArgumentException`.

So instead of just failing if parsing fails, return an Optional that all callers have to decide what to do (e.g. the YoutubeExtractor can just ignore the locale in that case, like it does with most other fields in the json if they are unexpected).


Stypox (approved):
Thank you! I pushed a commit that uses ParsingException instead of IllegalArgumentException in places where the former makes more sense.

PR #1163, opened 2024-03-30
[YouTube] Support new comments data

AudricV:

This PR adds support for the new rolling out YouTube comments data.

However, support for formatting (italic, strikethrough, bold, ...) has been not already added to attributed descriptions, this means that this PR introduces a regression for comments (and the relevant tests will fail).

This should be done in this PR, which would also apply this fix for video descriptions (hence the draft status), I need help about that (I can of course provide data to add this support).

I also improved the current comments code by removing outdated comment renderer data.

Fixes TeamNewPipe/NewPipe#10808.


Stypox:
I added support for formatting in https://github.com/TeamNewPipe/NewPipeExtractor/pull/1163/commits/2eca6d3ae62dde0d0d3933ba01031ec07a1a940e, so this is now ready for review, let me know what you think @AudricV.
Here is a testing APK: [app-debug.zip](https://github.com/TeamNewPipe/NewPipeExtractor/files/14817788/app-debug.zip)

Theta-Dev:
I have just run the test build and I dont get any continuation pages, both for the main comments and comment replies.

AudricV:
I though that removing the copy the contents array containing the comments renderer, but it wasn't at all. I reverted this change and added a comment.

I rebased on top of the dev and improved the message of @Stypox's commit and reverted the change of the comment formatting video URL, which should go in a separate commit (maybe along mocks changes, even if I have a PR updating client versions and more which will touch at all mocks which will come soon).

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeDescriptionHelper.java
```diff
@@ -0,0 +1,255 @@
...
+    private YoutubeDescriptionHelper() {
+    }
+
+    public static final String LINK_CLOSE = "</a>";
+    public static final String STRIKETHROUGH_OPEN = "<s>";
+    public static final String STRIKETHROUGH_CLOSE = "</s>";
+    public static final String BOLD_OPEN = "<b>";
+    public static final String BOLD_CLOSE = "</b>";
+    public static final String ITALIC_OPEN = "<i>";
+    public static final String ITALIC_CLOSE = "</i>";
```
AudricV, line 30: Is it intended that these constants are public? They are not used outside the class.

Stypox: No, they should be private, thank you for noticing

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeDescriptionHelper.java
```diff
@@ -0,0 +1,255 @@
...
+     * </p>
+     *
+     * @param openers contains all of the places where a run begins, must have the same size of
+     *                closers, must be ordered by {@link Run#pos}
+     * @param closers contains all of the places where a run ends, must have the same size of
+     *                openers, must be ordered by {@link Run#pos}
+     * @param content the content to apply formatting to
+     * @return the formatted content in HTML
+     */
+    static String runsToHtml(
```
Stypox, line 118: This is supposed to be package-private though, because I wanted to add some unit tests

Theta-Dev:
I have also noticed that the new version does not trim the spaces, slashes and dots from YouTube-internal links.

Screenshot_20240406-130741

Here is another thing that should be addressed:

When I implemented the attributed description parser, YouTube used the short link format containing the video title or channel name only for their internal links. Now these links are used for various social media platforms, too. Without the icons shown on the actual YouTube page it is now impossible to distinguish these links (see the three `aespa_official` links in my screenshot).

So my suggestion is to use the original link text only for YT-internal links. For all other links we should display the original URL (trimmed if it is too long).

YouTube trims the URLs to a maximum of 40 characters (so if they are longer than 37 chars, the rest is cut off and `...` is added).

AudricV:
> So my suggestion is to use the original link text only for YT-internal links. For all other links we should display the original URL (trimmed if it is too long).

We can also add support for social icons instead as `img` HTML tags with the icon sources, though this would require a bit of client support.

Theta-Dev:
I would not do this since it would add too much complexity.

To make the links look good, we would not only need to show the icon image next to the link text but also put everthing into this rounded chip. Since that is not a standard HTML element we would need to add custom styling.

There are also social media icons like the new X logo that are not easily recognizable when they are just as tall as normal text.

Stypox:
I pushed a commit that uses the accessibility label instead of the link text, but only for urls that do have an accessibility label (i.e. only chips) and only for non-YouTube urls. What do you think? You can use this apk:  [NewPipe_v0.27.0_RC2_release.zip](https://github.com/TeamNewPipe/NewPipe/files/14899138/NewPipe_v0.27.0_RC2_release.zip).

Theta-Dev:
Looks good to me. Good idea to extract the service name from the accessibility text

Stypox:
Yes, my regex was a bit flaky in that screenshot. But I have already fixed it in the code (made the captured text as small as possible with `.*?`, so that the 0u00a0+ can grow as much as possible).

Stypox:
I pushed three commits:
- cleaned up YoutubeDescriptionHelper according to reviews and removed unneeded `isClose` field
- add tests for YoutubeDescriptionHelper
- avoid XSS attacks by escaping all content, you can test with a comment in [this](https://www.youtube.com/watch?v=8GjOvYh4udw) video, note that the old implementation (also used by Piped) did not escape everything, as seen in the [Piped counterpart](https://piped.video/watch?v=8GjOvYh4udw) of that video

Here is an APK: [Uploading app-release.zip…]()

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeDescriptionHelper.java
```diff
@@ -0,0 +1,316 @@
...
+
+    private static Function<String, String> getTransformContentFun(final JsonObject run) {
+        final String accessibilityLabel = run.getObject("onTapOptions")
+                .getObject("accessibilityInfo")
+                .getString("accessibilityLabel", "")
+                // accessibility labels are e.g. "Instagram Channel Link: instagram_profile_name"
+                .replaceFirst(" Channel Link", "");
+
+        final Function<String, String> transformContent;
+        if (accessibilityLabel.isEmpty() || accessibilityLabel.startsWith("YouTube: ")) {
```
AudricV, line 261: Same as above for the `YouTube: ` part.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeDescriptionHelper.java
```diff
@@ -0,0 +1,316 @@
...
+                        return;
+                    }
+                    final int end = start + length;
+
+                    if (run.has("strikethrough")) {
+                        openers.add(new Run(STRIKETHROUGH_OPEN, STRIKETHROUGH_CLOSE, start));
+                        closers.add(new Run(STRIKETHROUGH_OPEN, STRIKETHROUGH_CLOSE, end));
+                    }
+
+                    if (run.getBoolean("italic", false)) {
```
AudricV, line 304: The `getBoolean` method of JSON objects should already return `false` if the value isn't present, that's fine if the code stay as it is.

AudricV (commented):
`YoutubeDescriptionHelper` changes and the tests ones look good to me, I just have a few notes for which nothing could or should change. Thank you @Stypox!

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeDescriptionHelper.java
Stypox: Yeah I wanted to be explicit here

Iodine98:
@AudricV 
When will the new version of the app be released with the fix? Because the PR merges to `TeamNewPipe:dev` on the `NewPipeExtractor` repo rather than some branch on the `NewPipe` repo. See [TeamNewPipe/NewPipe/#10968](https://github.com/TeamNewPipe/NewPipe/issues/10968) (duplicate) and [TeamNewPipe/NewPipe/#10808](https://github.com/TeamNewPipe/NewPipe/issues/10808) (original)

Stypox:
@lodine98 see https://github.com/TeamNewPipe/NewPipe/issues/10930, also don't ping people

PR #1170, opened 2024-04-20
[YouTube] Support shows and page header on user channels

Stypox:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [ ] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).

Full credits to @AudricV

Fixes TeamNewPipe/NewPipe#5793, fixes TeamNewPipe/NewPipe#10826.

PR #1176, opened 2024-05-17
Extract Youtube Subscriptions for Pages

NWuensche:
In this PR, I added a way to get the subscription count for pages. Right now, "sometimes" when I open a youtube channel, it's type is "Page" and therefore no subscription count is displayed. However, one can extrat the subscription count from the metaData for pages.

With "sometimes", I mean that there seems to be a 25% change that when I open a youtube channel in the app, it's type is page. I haven't found any understandable ways on when this happens, it seems random to me. It can even happen that a channel that is not a page becomes a page when I restart the app. 


AudricV:
Thanks, but this should be already fixed by #1170, in a more dynamic and safer way than yours. Closing you pull request due to this reason.

PR #1179, opened 2024-05-19
Add support for on.soundcloud.com urls

### LINKED ISSUE #1178: Support for on.soundcloud.com
sometimes people link to soundcloud URLs in the form of `on.soundcloud.com`. 

example: https://on.soundcloud.com/a8QmYdMnmxnsSTEp9

Would be cool if this opened in newpipe, or if newpipe even supported opening in this kind of URL. currently clicking on this url in android results in opening in the browser which redirects to an "open this url in app" which doesn't work.

mjsir911:
Fixes #1178


mjsir911:
I'm not sure what's going on with the failing tests.

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
```diff
@@ -185,8 +190,20 @@ public static String resolveUrlWithEmbedPlayer(final String apiUrl) throws IOExc
...
     public static String resolveIdWithWidgetApi(final String urlString) throws IOException,
             ParsingException {
-        // Remove the tailing slash from URLs due to issues with the SoundCloud API
         String fixedUrl = urlString;
+
+        // if URL is an on.soundcloud link, do a request to resolve the redirect
+
+        if (ON_URL_PATTERN.matcher(fixedUrl).find()) {
+            try {
+                fixedUrl = NewPipe.getDownloader().head(fixedUrl).latestUrl();
```
AudricV, line 199: The latest URL we get contains tracking parameters. From the URL you linked in the issue you opened (#1178, `https://on.soundcloud.com/a8QmYdMnmxnsSTEp9`), I get redirected to:

`https://soundcloud.com/katagiri/ultimate-dragon-forest-kingdom-64-definitive-edition-game-of-the-year-mix?ref=clipboard&p=i&c=0&si=A1822F15965A4D9B8EEBDE9DC7524F10&utm_source=clipboard&utm_medium=text&utm_campaign=social_sharing`

You should remove them by removing any query parameter (I don't think they are useful in other places like artists, playlists and sets, but please correct me if I am wrong). This is not an issue for now as we do not support "private" tracks, i.e. tracks with secret tokens.

AudricV (dismissed):

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
mjsir911: oh, good catch! I guess this isn't explicitly necessary since the existing tests pass but to remove tracking and such I do agree this is good.

Sypersugma:
Бля,нечего не работает обновите нью пайп пж!

PR #1181, opened 2024-05-24
Throw upon cleartext HTTP queries

fynngodau:

NewPipe presently does not have a network policy (see [manifest](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/AndroidManifest.xml)), hence [from Android 9 onwards](https://developer.android.com/privacy-and-security/security-config#CleartextTrafficPermitted), it is not ever possible for it to make cleartext HTTP queries. Thus as of now, to avoid errors, the extractor must make sure that all queries it sends are encrypted HTTPS queries.

I propose explicitly forcing the extractor itself to refuse HTTP queries. This extends the guarantee that no HTTPS queries are made towards all other downstreams as well, and allows testing for bugs related to missing HTTPS queries from within the extractor's tests. https://github.com/TeamNewPipe/NewPipe/issues/11074 is such a bug that could have been caught through tests.

The alternative would be to transparently upgrade queries to HTTPS from within the `Downloader` class. However, I would prefer if extractors ensured that they never send HTTPS queries in the first place.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/downloader/Downloader.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/downloader/Downloader.java
@@ -3,6 +3,7 @@ package org.schabi.newpipe.extractor.downloader;
 import org.schabi.newpipe.extractor.NewPipe;
 import org.schabi.newpipe.extractor.exceptions.ReCaptchaException;
 import org.schabi.newpipe.extractor.localization.Localization;
+import org.schabi.newpipe.extractor.utils.Utils;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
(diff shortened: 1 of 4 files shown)


fynngodau:
Nicely, the tests fail because https://github.com/TeamNewPipe/NewPipeExtractor/pull/1177 is not merged. :slightly_smiling_face:

litetex:
Note: I'm not sure if this is required or a good idea because when e.g. running services like PeerTube locally (or doing some traffic inspection) this will cause problems.

NewPipeExtractor is also used in other services which might be affect by this.

I also don't think that the current situation has caused any problems so far.

Therefore I guess it's best to close/ignore this PR for now.

Stypox:
> e.g. running services like PeerTube locally (or doing some traffic inspection) this will cause problems.

Mmmh, in that case one could just build a test version of NewPipe with this feature disabled

PR #1196, opened 2024-07-18
Use the new URL encode/decode methods introduced in Java 10

Isira-Seneviratne:
----
Update the `Utils` encode/decode methods to call the new URL [encode](https://developer.android.com/reference/java/net/URLEncoder#encode(java.lang.String,%20java.nio.charset.Charset))/[decode](https://developer.android.com/reference/java/net/URLDecoder#decode(java.lang.String,%20java.nio.charset.Charset)) methods introduced in Java 10. These methods do not require `UnsupportedEncodingException` to be caught.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ If you're using Gradle, you could add NewPipe Extractor as a dependency with the
 -dontwarn org.mozilla.javascript.tools.**
 ```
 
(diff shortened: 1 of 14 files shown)


Isira-Seneviratne:
> This refactor deal is right up your alley, isn't it? xD You've been submitting optimisations since the very beginning.

Haha, yeah 😂

Isira-Seneviratne:
> AH wait, are those Constants finally desugared correctly?

Yes: https://developer.android.com/studio/write/java11-nio-support-table

PR #1197, opened 2024-07-19
[YouTube] Workaround HTTP 403s on streaming URLs of WEB client (after some time or instantly for some JavaScript players), update clients info

AudricV:

This PR works around an anti-bot token, for which its requirement is A/B tested on the WEB client. In this test, streaming URLs of this client return HTTP errors 403 if the token is not provided after some time.

It also allows to not fetch the JavaScript player for non-age restricted videos, reducing data usage. The TVHTML5 embed client is now also only fetched in the case of age-restricted videos.

:warning: The methods `forceFetchAndroidClient` and `forceFetchIosClient` of `YoutubeStreamExtractor` **have been removed**, as they are now not needed anymore. This is a **breaking change** for users of these methods.

The PR changes also break the extraction of appropriate error states for private and deleted videos and invalid video IDs (an invalid player response message is thrown instead of a content not available one). This should fixed before this PR is merged, but I can't find a solution, so I will need help for that.

Clients info have been updated, this means that mocks have to be updated and several tests unrelated to my changes are failing. I fixed one in this PR (at least for the connection I used to ran the tests), `YoutubeSearchExtractorTest.CrisisResources`.

Fixes TeamNewPipe/NewPipe#11191.


Stypox:
> The PR changes also break the extraction of appropriate error states for private and deleted videos and invalid video IDs (an invalid player response message is thrown instead of a content not available one). This should fixed before this PR is merged, but I can't find a solution, so I will need help for that.

Can't you just use `checkPlayabilityStatus()` like before? This is a response for a private video, which contains information about which error to throw:

```
{playabilityStatus={contextParams=..., messages=[This is a private video. Please sign in to verify that you may see it.], status=LOGIN_REQUIRED, errorScreen={playerErrorMessageRenderer={reason={simpleText=Private video}, thumbnail=..., icon={iconType=ERROR_OUTLINE}, subreason={simpleText=Sign in if you've been granted access to this video}, ...
```

This is the response for an invalid id:

```
{playabilityStatus={reason=Video unavailable, contextParams=Q0FBU0FnZ0E=, status=ERROR, errorScreen={playerErrorMessageRenderer={reason={simpleText=Video unavailable}, thumbnail=..., icon={iconType=ERROR_OUTLINE}}}}}
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -796,64 +794,65 @@ public void onFetchPage(@Nonnull final Downloader downloader)
...
+        final JsonObject webPlayerResponse = YoutubeParsingHelper.getWebPlayerResponse(
+                localization, contentCountry, videoId);
 
-        if (playerResponse == null) {
-            throw new ExtractionException("Could not get playerResponse");
+        // FIXME: checking the player response validity here makes the extractor throwing the
+        //  exception above for deleted and private videos and invalid video IDs
+        if (isPlayerResponseNotValid(webPlayerResponse, videoId)) {
+            throw new ExtractionException("Initial WEB player response is not valid");
         }
```
Stypox, line 805:
```suggestion
        if (isPlayerResponseNotValid(webPlayerResponse, videoId)) {
            checkPlayabilityStatus(webPlayerResponse,
                    webPlayerResponse.getObject("playabilityStatus"));
            throw new ExtractionException("Initial WEB player response is not valid");
        }
```

Stypox (approved):

offshade:
How does one get this fix?

PR #1198, opened 2024-07-20
Use TreeMap in Response

Isira-Seneviratne:
----
Store the response headers in a [`TreeMap`](https://docs.oracle.com/javase/8/docs/api/java/util/TreeMap.html). This allows the header value retrieval to be optimised as `TreeMap` orders its keys using a comparator.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/downloader/Response.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/downloader/Response.java
@@ -5,6 +5,7 @@ import javax.annotation.Nullable;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
 
 /**
  * A Data class used to hold the results from requests made by the Downloader implementation.
@@ -14,17 +15,18 @@ public class Response {
     private final String responseMessage;
     private final Map<String, List<String>> responseHeaders;
     private final String responseBody;
-
     private final String latestUrl;
 
     public Response(final int responseCode,
                     final String responseMessage,
-                    @Nullable final Map<String, List<String>> responseHeaders,
+                    @Nonnull final Map<String, List<String>> responseHeaders,
                     @Nullable final String responseBody,
                     @Nullable final String latestUrl) {
         this.responseCode = responseCode;
         this.responseMessage = responseMessage;
-        this.responseHeaders = responseHeaders == null ? Collections.emptyMap() : responseHeaders;
+
+        this.responseHeaders = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+        this.responseHeaders.putAll(responseHeaders);
 
         this.responseBody = responseBody == null ? "" : responseBody;
         this.latestUrl = latestUrl;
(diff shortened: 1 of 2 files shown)


Isira-Seneviratne:
I've revised the code a bit, to reuse the map if it is already a SortedMap.

Isira-Seneviratne:
> A related test is failing, something is wrong with the changes
> 
> ```
> Invalid response type (got "null", excepted a JSON response) (response code 200)
> org.schabi.newpipe.extractor.exceptions.ExtractionException: Invalid response type (got "null", excepted a JSON response) (response code 200)
> 	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeSuggestionExtractor.suggestionList(YoutubeSuggestionExtractor.java:71)
> 	at org.schabi.newpipe.extractor.services.youtube.YoutubeSuggestionExtractorTest.testIfSuggestions(YoutubeSuggestionExtractorTest.java:55)
> ```

I wasn't able to reproduce the issue locally, the header value is being returned as expected.

extractor/src/main/java/org/schabi/newpipe/extractor/downloader/Response.java
```diff
@@ -2,29 +2,35 @@
...
-    private final Map<String, List<String>> responseHeaders;
+    private final SortedMap<String, List<String>> responseHeaders;
     private final String responseBody;
-
     private final String latestUrl;
 
     public Response(final int responseCode,
                     final String responseMessage,
-                    @Nullable final Map<String, List<String>> responseHeaders,
+                    @Nonnull final Map<String, List<String>> responseHeaders,
```
Stypox, line 22: Why is this changed to be `@Nonnull`? Isn't this a breaking change?

extractor/src/main/java/org/schabi/newpipe/extractor/downloader/Response.java
```diff
@@ -2,29 +2,35 @@
...
         this.responseCode = responseCode;
         this.responseMessage = responseMessage;
-        this.responseHeaders = responseHeaders == null ? Collections.emptyMap() : responseHeaders;
+
+        if (responseHeaders instanceof SortedMap) {
+            this.responseHeaders = (SortedMap<String, List<String>>) responseHeaders;
+        } else {
+            this.responseHeaders = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+            this.responseHeaders.putAll(responseHeaders);
+        }
```
Stypox, line 33: Mmmh, this still means that all items are reprocessed when building a `Response` in case the headers were not already in a `SortedMap`. Also, how can you be sure that `responseHeaders` was built using `String.CASE_INSENSITIVE_ORDER`?

I would just build a new `TreeMap` every time.

Stypox (changes requested):
Somewhere we have to iterate through all headers anyway: either in the `Response` constructor, or in the `getHeader()` function, since we don't know with which `Map` the `Response` is being built (it might be a `SortedMap` sorted according to a wrong criterion). So I would suggest recreating a new `TreeMap` every time in the `Response` constructor, since that's called only once, while `getHeader` might be called multiple times. 

Replying to @TobiGr, I think we can't use a `HashMap` if we want a sorted map. But in any case, given the small amount of headers that there usually are (<20), the `O(log n)` of the `TreeMap` is actually much faster than the `O(1)` of the `HashMap`: hash maps have a big constant in the `O(1)` and get faster only after n grows large enough.

Another option would be to create a hash map where keys are already lowercased in the constructor, and then just accessing the map with `name.lowercase()` in `getHeader()`.

In general I don't think this change is so important anyway, since the `O(n)` we had in `getHeader()` is not so bad given that `n` is small, and `getHeader()` is basically never used anyway:
image

So I would just keep the old `O(n)` behavior in `getHeader()`, but `O(1)` in the constructor, rather than switching to `O(n)` in the constructor (which is executed **all the time**) and `O(log n)`/`O(1)` in `getHeader()` (which is basically never called).

PR #1208, opened 2024-08-05
[Youtube] Add linked album for ytmusic searches

wispl:

Adds the linked album to the song when doing a ytmusic search. This is a draft and the current change is not visible. Some questions:

1. To expose the linked album for use, is it alright to modify StreamInfoItem to add a new method called getPlaylist?
2. Currently the code just returns the browseId of the album, is there any preferred structure to also return the name of the album as well?

The browseId currently does nothing, hopefully I can try to get the browse endpoint working for ytmusic. I also plan to add a method to get the linked artist as well.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicSongOrVideoInfoItemExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicSongOrVideoInfoItemExtractor.java
@@ -174,4 +174,26 @@ public class YoutubeMusicSongOrVideoInfoItemExtractor implements StreamInfoItemE
             throw new ParsingException("Could not get thumbnails", e);
         }
     }
+
+    @Nonnull
+    public String getPlaylist() {
+        if (searchType.equals(MUSIC_SONGS)) {
+            for (final Object item : descriptionElements) {
+                final JsonObject browseEndpoint = ((JsonObject) item)
+                        .getObject("navigationEndpoint")
+                        .getObject("browseEndpoint");
+
+                final String type = browseEndpoint
+                        .getObject("browseEndpointContextSupportedConfigs")
+                        .getObject("browseEndpointContextMusicConfig")
+                        .getString("pageType");
+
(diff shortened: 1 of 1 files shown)


Akash-Sriram:
is this still in progress ?

wispl:
Kind of, it has been a while so I have to see if the changes even work now.

wispl:
I added a new method (getPlaylistId) to StreamInfoItem. This PR introduces an useful value for the playlistId only for youtube music searches which is the browseId.

wispl:
A better idea might be to replace the playlistId with the url of the playlist?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicSongOrVideoInfoItemExtractor.java
```diff
@@ -174,4 +174,26 @@ public List<Image> getThumbnails() throws ParsingException {
             throw new ParsingException("Could not get thumbnails", e);
         }
     }
+
+    @Override
+    public String getPlaylistId() throws ParsingException {
```
Stypox: This should be `getPlaylistUrl()` and return the playlist URL, rather than the ID, just like we have `getUploaderUrl()`. Also, you could also add `getPlaylistName()` since YouTube music seems to provide that information too.

Stypox (changes requested):
Thank you! Sorry for not reviewing earlier, I somehow didn't notice this PR.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicSongOrVideoInfoItemExtractor.java
wispl: Sorry for the really long delay (been busy this semester), I can change the id to a playlist url, but I can't seem to find where the playlist name is exposed, at least for songs.

Akash-Sriram:
wish someone should added this feature is so important for music lover

PR #1221, opened 2024-09-12
[YouTube] Fix extraction of Shorts in channels and remove visitor data usage

AudricV:

This PR fixes the extraction of Shorts in their channel tab due to a new UI element being rolled-out replacing the `reelItemRenderer`, `shortsLockupViewModel`. It doesn't provide more information than a `reelItemRenderer`.

It also removes the usage of a visitor data in continuations of Shorts channel tabs, which used to be mandatory to get valid continuations with more Shorts. It isn't the case anymore, and the visitor data wasn't used since 5a6da5f43ebb0a612e6ffdd3feb90d78b3dd383e, due to a lack of an index update with the channel IDs changes in this commit (the channel verified status was used instead).

The method `prepareDesktopJsonBuilder` of `YoutubeParsingHelper` with the visitor data argument isn't now used directly, but has been kept for now as it is (it will be very likely used with the PoToken changes for streaming URLs).

Due to these visitor data changes, the mocks of `YoutubeChannelTabExtractorTest.Shorts` test class had to be updated.

Fixes TeamNewPipe/NewPipe#11525.


Stypox (approved):
Code looks good to me, although I didn't test. Thank you!

PR #1222, opened 2024-09-15
[YouTube] Fix serialization of Videos channel tab when it is already fetched

AudricV:
- [ ] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API. Not needed, the API changes are on code that is not a part of the API that users should use (that's the case of NewPipe).

This PR fixes the ability to serialize `ReadyChannelTabListLinkHandler`s for YouTube channels. It was due to the usage of the `channelHeader` field of `YoutubeChannelExtractor`, as explained in https://github.com/TeamNewPipe/NewPipe/issues/11356#issuecomment-2266657671. Even if this usage has been removed by copying the instance members into a new object, the `ChannelHeader` class of `YoutubeChannelHelper` was not serializable, so crashes would still happen if I didn't make this class serializable.

`Optional` as fields and parameters of methods have been removed in the related extractor code as it is not a good practice. This simplifies in some places channel info extraction code.

This PR and the issue it fixes also raise the question of the limits of the serialization system we use, both in the app and in the extractor.

Debug NewPipe APK based on app commit `035c394cf6704c0af9077c01ebfa55b810d02140` with changes from commit 422df7002e6aa2d2096b4ef0c1546002823445b6: [app-debug.zip](https://github.com/user-attachments/files/17006047/app-debug.zip)

Fixes TeamNewPipe/NewPipe#11356.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeChannelHelper.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeChannelHelper.java
@@ -11,6 +11,7 @@ import org.schabi.newpipe.extractor.localization.Localization;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.io.IOException;
+import java.io.Serializable;
 import java.nio.charset.StandardCharsets;
 import java.util.Optional;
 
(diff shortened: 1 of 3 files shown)


Stypox (approved):
Nice catch! I can't really reproduce the crash on NewPipe, but it totally makes sense that it might happen. Thank you for noticing and fixing!

PR #1246, opened 2024-12-07
Use BufferedReader#lines()

Isira-Seneviratne:
Use the [`lines()`](https://developer.android.com/reference/java/io/BufferedReader#lines()) method in `BufferedReader`, which is desugared for Android versions below 7.0 (when it was first introduced in Android).

----


litetex (approved):
Thank you for the PR!

LGTM
Did some rebasing and polishing. Test are working and import also works with NewPipe itself.

Let's get this merged

PR #1247, opened 2024-12-12
[YouTube] Potokens support implementation

FireMasterK:

Thanks to @AudricV for doing most of the work here! :)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/PoTokenProvider.java
```diff
@@ -0,0 +1,113 @@
...
+     * There should be only poToken needed, for the player requests.
+     * </p>
+     *
+     * @return a {@link PoTokenResult} specific to the ANDROID InnerTube client
+     */
+    @Nullable
+    PoTokenResult getAndroidClientPoToken(String videoId);
+
+    /**
+     * Get a {@link PoTokenResult} specific to the Android app, a.k.a. the ANDROID InnerTube client.
```
FineFindus, line 97: I think this is for iOS, not Android?

AudricV:
Sorry for the late reply.

No, it was fixed locally when I resolved the conversations. I thought I would be faster to push my changes, but this wasn't the case.

This PR is superseded by #1272.

PR #1250, opened 2024-12-22
Add media.ccc specific parsing of recordings

Profpatsch:
Instead of pushing the `media.ccc` recordings straight into the StreamExtractor straightjacket (which does not really provide the correct API for handling the semantics of the media.ccc data), we parse into an intermediate structure first that contains all relevant information about the different streams.

This is required for correct handling of the different languages.

It should not change the API, since its specfic to the media.ccc Extractor. I just type-cast in frankenplayer at the moment.
I skipped adding getters/setters in the MediaCCCRecording structs, once this repository is migrated to Kotlin (?), it would just be a data class.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCStreamExtractor.java
@@ -20,6 +20,7 @@ import org.schabi.newpipe.extractor.exceptions.ParsingException;
 import org.schabi.newpipe.extractor.linkhandler.LinkHandler;
 import org.schabi.newpipe.extractor.localization.DateWrapper;
 import org.schabi.newpipe.extractor.localization.Localization;
+import org.schabi.newpipe.extractor.services.media_ccc.extractors.data.MediaCCCRecording;
 import org.schabi.newpipe.extractor.services.media_ccc.linkHandler.MediaCCCConferenceLinkHandlerFactory;
 import org.schabi.newpipe.extractor.services.media_ccc.linkHandler.MediaCCCStreamLinkHandlerFactory;
 import org.schabi.newpipe.extractor.stream.AudioStream;
@@ -28,15 +29,18 @@ import org.schabi.newpipe.extractor.stream.StreamExtractor;
 import org.schabi.newpipe.extractor.stream.StreamType;
 import org.schabi.newpipe.extractor.stream.VideoStream;
 import org.schabi.newpipe.extractor.utils.JsonUtils;
-import org.schabi.newpipe.extractor.utils.LocaleCompat;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 public class MediaCCCStreamExtractor extends StreamExtractor {
     private JsonObject data;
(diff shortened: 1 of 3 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCStreamExtractor.java
```diff
@@ -167,18 +162,119 @@ public List<VideoStream> getVideoStreams() throws ExtractionException {
...
-                        .setResolution(recording.getInt("height") + "p")
+                        .setResolution(recording.height + "p")
                         .build());
-            }
         }
 
         return videoStreams;
     }
 
+    public List<MediaCCCRecording> getRecordings() {
```
Stypox, line 176: Sorry for the late review. You should probably cache the value returned by `getRecordings()` to avoid recalculating it in `getVideoStreams()`, `getAudioStreams()` and `getVideoOnlyStreams()`

extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCStreamExtractor.java
```diff
@@ -100,64 +104,55 @@ public List<Image> getUploaderAvatars() {
...
     @Override
     public List<VideoStream> getVideoStreams() throws ExtractionException {
-        final JsonArray recordings = data.getArray("recordings");
+
+        final List<MediaCCCRecording.Video> recordings = getRecordings().stream()
+                .flatMap(r ->
+                        r instanceof MediaCCCRecording.Video
+                                ? Stream.of((MediaCCCRecording.Video) r)
+                                : Stream.empty()
+                )
```
Stypox, line 147: (also in other places)
```suggestion
                .filter(MediaCCCRecording.Video::isInstance)
                .map(MediaCCCRecording.Video::cast)
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/data/MediaCCCRecording.java
```diff
@@ -0,0 +1,72 @@
...
+        public int height;
+    }
+
+    /** Some talks have multiple kinds of video. */
+    enum VideoType {
+        /** The main recording of a talk/event. */
+        MAIN,
+        /** A side-recording of a talk/event that has the slides full-screen.
+         * Usually if there is a slide-recording there is a MAIN recording as well */
+        SLIDES
```
Stypox, line 34: Probably we should add something like [this](https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/main/java/org/schabi/newpipe/extractor/stream/AudioTrackType.java) but for videos, to distinguish between the main and secondary types

Stypox (changes requested):
Sorry for the late review, I think this PR is well organized. Thanks!

extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/MediaCCCStreamExtractor.java
Profpatsch: Oh, but it’s not doing a network request here, the `data` object will be used for both function calls as long as the object is preserved between them. The Json extracting code itself should be on the order of microseconds.

Stypox: True, it's not important, it just feels bad to do the same extraction 3 times

extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/extractors/data/MediaCCCRecording.java
Stypox: True, but it's something that should be exposed at the extractor-level APIs, not be specific for every service. This way it would be easy in clients to add a switch that allows choosing the video track type.

Profpatsch: Yes, maybe, once we figured out which parts are specific to media.ccc.de and which ones can be abstracted :)

I think a major issue with NPE is that it’s trying to unify the services too much, and then we get lots of workarounds on the NewPipe side cause we have to reconstruct specific service behaviour again that was unified on the NPE side. I’d rather cast on the class and then have some service-specific functionality available if I need it in the player.

Stypox:
I would be ok with merging this as-is (after you fix my [other comment](https://github.com/TeamNewPipe/NewPipeExtractor/pull/1250#discussion_r1971654583)), since it's not adding any new functionality for now but just adds the ability to load all recordings. Going forward, I believe it's important to make the API general (especially when it comes to just classifying the type of audio/video, be it slides, normal video, voiceover, etc.), but to also allow subclasses with service-specific information as suggested by Audric in https://github.com/TeamNewPipe/NewPipeExtractor/issues/858#issuecomment-3289753673 .

PR #1253, opened 2024-12-29
[YouTube] Fixes for n param deobfuscation function

### LINKED ISSUE #1252: Could not find throttling deobfuscation function
### Checklist

- [X] I am able to reproduce the bug with the latest version given here: [CLICK THIS LINK](https://github.com/TeamNewPipe/NewPipeExtractor/releases/latest).
- [X] I am aware that this issue is being opened for the NewPipe Extractor, NOT the [app](https://github.com/TeamNewPipe/NewPipe), and my bug report will be dismissed otherwise.
- [X] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipeExtractor/issues) or [closed](https://github.com/TeamNewPipe/NewPipeExtractor/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [X] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [X] This issue contains only one bug.
- [X] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Affected version

0.24.3

### Steps to reproduce the bug

Call `YoutubeJavaScriptPlayerManager.getUrlWithThrottlingParameterDeobfuscated`

### Expected behavior

It should return the url with the deobfuscated throttling parameter

### Actual behavior

It throws a ParsingException

### Screenshots/Screen recordings

_No response_

### Logs

```
org.schabi.newpipe.extractor.exceptions.ParsingException: Could not find deobfuscation function with any of the known patterns in the base JavaScript player code
    at org.schabi.newpipe.extractor.services.youtube.YoutubeThrottlingParameterUtils.getDeobfuscationFunctionName(YoutubeThrottlingParameterUtils.java:134)
    at org.schabi.newpipe.extractor.services.youtube.YoutubeJavaScriptPlayerManager.getUrlWithThrottlingParameterDeobfuscated(YoutubeJavaScriptPlayerManager.java:242)
    ...
Caused by: org.schabi.newpipe.extractor.utils.Parser$RegexException: Failed to find pattern "[a-zA-Z0-9$_]="nn"\[\+[a-zA-Z0-9$_]+\.[a-zA-Z0-9$_]+],[a-zA-Z0-9$_]+\([a-zA-Z0-9$_]+\),[a-zA-Z0-9$_]+=[a-zA-Z0-9$_]+\.[a-zA-Z0-9$_]+\[[a-zA-Z0-9$_]+]\|\|null\).+\|\|([a-zA-Z0-9$_]+)\(""\)"
    at org.schabi.newpipe.extractor.utils.Parser.matchMultiplePatterns(Parser.java:96)
    at org.schabi.newpipe.extractor.services.youtube.YoutubeThrottlingParameterUtils.getDeobfuscationFunctionName(YoutubeThrottlingParameterUtils.java:131)
    at org.schabi.newpipe.extractor.services.youtube.YoutubeJavaScriptPlayerManager.getUrlWithThrottlingParameterDeobfuscated(YoutubeJavaScriptPlayerManager.java:242) 
    ...
```

### Additional information

_No response_

gechoto:

Fixes #1252

Changes:
- new regex for extracting the name of the n param deobfuscation function
- added fixup for the n param deobfuscation functions code to prevent an early return

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingParameterUtils.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingParameterUtils.java
@@ -58,14 +58,16 @@ final class YoutubeThrottlingParameterUtils {
              * The third regex matches the following text, where we want SDa and the array index
              * accessed:
              *
-             * WL(a),c=a.j[b]||null)&&(c=SDa[0](c)
+             * ,Vb(m),W=m.j[c]||null)&&(W=cvb[0](W),m.set(c,W)
              */
-            Pattern.compile(MULTIPLE_CHARS_REGEX + "\\("
+            Pattern.compile("," + MULTIPLE_CHARS_REGEX + "\\("
                     + MULTIPLE_CHARS_REGEX + "\\)," + MULTIPLE_CHARS_REGEX + "="
                     + MULTIPLE_CHARS_REGEX + "\\." + MULTIPLE_CHARS_REGEX + "\\["
-                    + MULTIPLE_CHARS_REGEX + "]\\|\\|null\\)&&\\(" + MULTIPLE_CHARS_REGEX + "=("
-                    + MULTIPLE_CHARS_REGEX + ")" + ARRAY_ACCESS_REGEX
-                    + "\\(" + MULTIPLE_CHARS_REGEX + "\\)"),
+                    + MULTIPLE_CHARS_REGEX + "]\\|\\|null\\)&&\\(\\b" + MULTIPLE_CHARS_REGEX + "=("
+                    + MULTIPLE_CHARS_REGEX + ")" + ARRAY_ACCESS_REGEX + "\\("
+                    + SINGLE_CHAR_VARIABLE_REGEX + "\\)," + MULTIPLE_CHARS_REGEX
+                    + "\\.set\\((?:\"n+\"|" + MULTIPLE_CHARS_REGEX + ")," + MULTIPLE_CHARS_REGEX
+                    + "\\)"),
 
             /*
              * The fourth regex matches the following text, where we want rma:
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingParameterUtils.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingParameterUtils.java
@@ -42,21 +42,6 @@ final class YoutubeThrottlingParameterUtils {
                     + MULTIPLE_CHARS_REGEX + ")" + ARRAY_ACCESS_REGEX + "\\("),
 
             /*
-             * This regex matches the following text, where we want Wma:
-             *
-             * a.D&&(b="nn"[+a.D],WL(a),c=a.j[b]||null)&&(c=SDa[0](c),a.set(b,c),SDa.length||Wma("")
-             *
-             * UPDATE: This is broken and Wma is not the function needed anymore.
-             * We need the function which is in SDa[0] instead.
-             */
-//            Pattern.compile(SINGLE_CHAR_VARIABLE_REGEX + "=\"nn\"\\[\\+" + MULTIPLE_CHARS_REGEX
-//                    + "\\." + MULTIPLE_CHARS_REGEX + "]," + MULTIPLE_CHARS_REGEX + "\\("
-//                    + MULTIPLE_CHARS_REGEX + "\\)," + MULTIPLE_CHARS_REGEX + "="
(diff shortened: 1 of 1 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingParameterUtils.java
```diff
@@ -30,17 +30,31 @@ final class YoutubeThrottlingParameterUtils {
...
+             *
+             * UPDATE: This is broken and Wma is not the function needed anymore.
+             * We need the function which is in SDa[0] instead.
+             */
+//            Pattern.compile(SINGLE_CHAR_VARIABLE_REGEX + "=\"nn\"\\[\\+" + MULTIPLE_CHARS_REGEX
+//                    + "\\." + MULTIPLE_CHARS_REGEX + "]," + MULTIPLE_CHARS_REGEX + "\\("
+//                    + MULTIPLE_CHARS_REGEX + "\\)," + MULTIPLE_CHARS_REGEX + "="
+//                    + MULTIPLE_CHARS_REGEX + "\\." + MULTIPLE_CHARS_REGEX + "\\["
+//                    + MULTIPLE_CHARS_REGEX + "]\\|\\|null\\).+\\|\\|(" + MULTIPLE_CHARS_REGEX
+//                    + ")\\(\"\"\\)"),
```
Stypox, line 57: Why did you remove the previous regex? It's better if you keep it, so in case YouTube reverts something it's already there and it works.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingParameterUtils.java
```diff
@@ -214,4 +237,17 @@ private static String validateFunction(@Nonnull final String function) {
         JavaScript.compileOrThrow(function);
         return function;
     }
+
+    @Nonnull
+    private static String fixupFunction(@Nonnull final String function)
```
Stypox, line 242: Mmmh, this function is specific to the new player code, right? Maybe it would be best to return the original `function` if the regexes don't find anything.

Also, could you add a comment explaining what `fixupFunction` does with an example?

Stypox (changes requested):
Thank you!

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingParameterUtils.java
AudricV: > (Also it might be a good idea to add some additional checks to validate it found the correct function but I consider this out of scope for this PR.)

That's a pretty good idea, however if you are using other clients not requiring deobfuscation like it is the case currently (we do not use HTML5 clients for now except for age-restricted videos, but this is a broken workaround that will be removed soon), you are preventing streams extraction. This already happened in past. A general error/logging warning is something that would fit the best.

> My suggestion for now:
> I will add it back in but swap it with the next one.
> What do you think?

Keep the current third version as it is and move the current first one, potentially finding the wrong function, at the third place and the current regexes from this position to the last one after the new third regex. Let me know if I am not clear.

AudricV:
Is it really needed to add a new regex matching partially what the current second one does? I think it could increase the chances to extract the wrong function, as there is nothing specific about the `n` param itself like the `nn` string.

ildar:
I see the merging is blocked. Is it possible to advance it? AFAIK some downstreams depend on it.

Also, @gechoto , could you please rebase it on top of the latest release?

thanks guys, your work is hard to overestimate for many many ppl!

ildar:
thanks a lot

Stypox:
Mmmh, I don't see the point of not putting the most recent regex in first place. YouTube is unlikely to rollback to the previous player version.
I pushed two commits, one to match more context around the function we are looking for in the new regex (see [yt-dlp](https://github.com/yt-dlp/yt-dlp/blob/de82acf8769282ce321a86737ecc1d4bef0e82a7/yt_dlp/extractor/youtube.py#L3320)'s regex), and another to remove "first regex", "second regex", ... in the comments (which are useless imho).

AudricV (approved):

Stypox:
Yes, the point is that every regex is made specifically for one player version from YouTube, and the current version only matches the third regex (and none of the others), so it doesn't make sense not to put that one regex in first position. It's not like YouTube is going to rollback to the previous player (I mean, it happened once I think, and that's why we have the other regexes in the list).

Stypox (approved):
Nevermind, we can always change it later

PR #1262, opened 2025-01-19
[YouTube] update iOS client, add visitor data to requests

Theta-Dev:
- [X] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [X] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).
- [X] No API changes

I have fixed the issue with the iOS client returning invalid responses
Apparently the only issue with the iOS client is that we were not sending it visitor data cookies. Since the visitor data cookies are just protobuf objects with a random ID, timestamp and country code, they can be randomly generated and dont need to be fetched from YouTube.

Fixes #1259
Fixes TeamNewPipe/NewPipe/issues/11934

PR #1212, opened 2024-08-19
Remove unused method in TimeAgoParser

Isira-Seneviratne:
* Removed the `parseDuration` method in `TimeAgoParser`, as it was not being used anywhere in the extractor or NewPipe.
* Updated the test cases for `TimeAgoParser`.
----
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).

PR #1263, opened 2025-01-23
Partial followup for #1262

litetex:
* ~Fixed Checkstyle~ (Already merged via https://github.com/TeamNewPipe/NewPipeExtractor/pull/1212)
* Use ``UncheckedIOException`` instead of ``RuntimeException`` as it's more specific
* Use correct iPhone model in docs
* Documentation has to be done by @Theta-Dev as I didn't write that code and have no deeper insight into this


litetex:
> @litetex do you want to include it in this PR?

@gechoto 
I think it would be better to have another PR, since this is kind of off-topic and I also can't review my own PRs ;)

Stypox (approved):

PR #1272, opened 2025-02-01
[YouTube] Refactor player clients, add support for poTokens, extract visitor data from the service and more

AudricV:
- [x] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API. TeamNewPipe/NewPipe#11955

This PR, which supersedes #1247, adds support for poTokens, refactor clients used and extract visitor data from the service itself, as generated visitor data do not work anymore to get valid player responses (rolling out in progress as of writing the first version of this PR description).

It also replaces age-restricted videos streams' fetching with the web embed player, which allows to play a very few age-restricted videos anonymously. This is the best we can do with what we can use (unless some method is still present and we haven't discovered it, but I have a lot of doubts).

TODO:

- [ ] describe in a more detailed way changes in this PR?
- [ ] add Javadocs
- [ ] update mocks
- [ ] add a poToken implementation for tests (in this PR?)

Fixes TeamNewPipe/NewPipe#11980.
Fixes TeamNewPipe/NewPipe#11803.
Fixes TeamNewPipe/NewPipe#11486.
Fixes partially TeamNewPipe/NewPipe#11382.


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -104,6 +101,10 @@
 
 public class YoutubeStreamExtractor extends StreamExtractor {
 
+    @Nullable
+    private static PoTokenProvider poTokenProvider;
+    private static boolean fetchIosClient;
```
kitterion, line 106: I'm sorry for a sort of a drive-by comment, but is it possible to not make these `static`? For example, have them in `YoutubeService` and pass them along when getting the stream extractor? It's trivial to make them static downstream if needed, and in the typical usage, where the `YoutubeService` is already static, they will also be effectively static with something like `ServiceList.YouTube.setPoTokenProvider(...)`.

My use case is that running the stream extractor is an expensive operation taking multiple seconds, so if I need to run multiple at the same time, I do it from multiple threads using different service and extractor instances.
It would be nice to not have to worry about synchronization here, reduce global state, and have the ability to have extractors with different providers.

Stypox: I think we have static fields in various places of the extractor, so this issue is not limited to this PR. Please open a pull request to fix this across the whole extractor ;-).

> My use case is that running the stream extractor is an expensive operation taking multiple seconds, so if I need to run multiple at the same time, I do it from multiple threads using different service and extractor instances.
It would be nice to not have to worry about synchronization here, reduce global state, and have the ability to have extractors with different providers.

I am not sure how your setup works, but can't you first set the static fields and then start multiple threads? Then if `poTokenProvider` is thread-safe (NewPipe's one is) everything that uses it should also be thread safe. Or am I missing something?

kitterion: > I think we have static fields in various places of the extractor, so this issue is not limited to this PR. Please open a pull request to fix this across the whole extractor ;-)

Sure, I'll do that :), I thought it was fine already for some reason. Gotta at least fix it for my own sanity.
However, this is part of a public facing api, and I'd like not to break people unnecessarily that will start to depend on this.

> I am not sure how your setup works, but can't you first set the static fields and then start multiple threads? Then if poTokenProvider is thread-safe (NewPipe's one is) everything that uses it should also be thread safe. Or am I missing something?

True, it's not impossible to adapt, what I'm saying is that with this change I'm _forced_ to make it thread-safe and I'm _forced_ to have only one provider for the whole program. It seems unnecessary, I would like to have other options. Like having extractors with different providers or different settings could conceivably be useful in case some of them don't work.
A library probably shouldn't force these decisions onto the users. Besides, the users can always make a static provider if they wish.
If I'm reading it correctly, NewPipe does exactly that: makes its own static provider.

FireMasterK: Could you help me understand your use-case more? Why would you need more than one provider for all of your extractors?

In case your `PoTokenProvider` isn't thread-safe, you could just have a field, and use a `synchronized` block for the unsafe activity. Would it be more helpful to have documentation somewhere for this?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -900,128 +891,213 @@ private void checkPlayabilityStatus(final JsonObject youtubePlayerResponse,
...
-     * Fetch the Android Mobile API and assign the streaming data to the androidStreamingData JSON
-     * object.
-     */
-    private void fetchAndroidMobileJsonPlayer(@Nonnull final ContentCountry contentCountry,
-                                              @Nonnull final Localization localization,
-                                              @Nonnull final String videoId)
+    private void fetchHtml5Client(@Nonnull final Localization localization,
+                                  @Nonnull final ContentCountry contentCountry,
+                                  @Nonnull final String videoId,
+                                  @Nullable final PoTokenProvider poTokenProviderInstance,
```
FireMasterK, line 907: Why don't we simply do a null-check rather than having 2 fields for this? (in all the fetch methods)

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/PoTokenProvider.java
```diff
@@ -0,0 +1,120 @@
...
+     * <p>
+     * Note that YouTube desktop website generates two {@code poToken}s:
+     * - one for the player requests {@code poToken}s, using the videoId as the minter value;
+     * - one for the streaming URLs, using a visitor data for logged-out users as the minter value.
+     * </p>
+     *
+     * @return a {@link PoTokenResult} specific to the WEB InnerTube client
+     */
+    @Nullable
+    PoTokenResult getWebClientPoToken(String videoId);
```
FireMasterK, line 52: Could we want to create a potoken with some other data other than the videoId?

https://github.com/LuanRT/BgUtils/blob/main/examples/node/innertube-challenge-fetcher-example.ts
Over here, for example, the `pot` url parameter is based on a 'session' PoToken based on the visitorData, while the the one sent to the player request is based on the videoId?

In that case, why does `PoTokenResult` need to have a `visitorData`?

Is having 2 different PoTokens the behavior on YouTube, or is what this PR does correct?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
kitterion: Is it reasonable to want to switch fields like `fetchIosClient` or return values of `PoTokenProvider` methods at runtime? I feel like it could be because youtube requirements are constantly changing and it'd be nice to have multiple configurations available and switch between them at runtime if one doesn't work, but maybe I'm wrong.

If it's not reasonable, then one time initialization of read-only statics is perfectly fine. If it is, then for example there's currently no thread safe way to change the value of `fetchIosClient` without synchronizing calls to the extractor.

There's another issue that I spotted but I'm not sure if it's relevant: I don't see how to switch what the `PoTokenProvider` methods return in a consistent manner and without a race condition. Specifically, `PoTokenProvider` currently requires implementations to essentially return 4 values (some of them potentially `null`). There's no way to consistently switch them to another set of 4 values in a multi-threaded environment. The reason is that `onFetchPage` simply calls methods in order, and the values can change between calls even if I synchronize everything in a `PoTokenProvider` on my end.
If this is not an issue, still having to synchronize everything is not great. It's error prone, requires you to think about what you're doing.

I avoid having to think about this because in my use case threads don't really know about other threads, they use thread-local instances. That way there's really no synchronization, data races, inconsistencies, etc., it's very convenient. Though I'm told this is not currently safe and will need some fixing.
(My actual use case is a bit more ~~cursed~~ involved, java is not the main language, I call it through jni, even the threads are created outside of the jvm and then attached to it. So having to juggle multiple languages with threads in the mix, I really try to minimize the possibility of things that can go wrong.)

So if I were to summarize my approach here, it's: why worry about any of this if you can just avoid it altogether and write code as if it's single threaded.
It eliminates a whole class of errors, I'm a big fan.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/PoTokenProvider.java
AudricV: > Is having 2 different PoTokens the behavior on YouTube, or is what this PR does correct?

YouTube desktop website uses indeed two poTokens: one minted with a visitorData, one with a videoId.

The visitorData must be the same for player requests and for the visitorData poToken, used in streaming URLs.

On embeds, that's currently not the case: it uses the old behavior of sending a visitorData poToken for player and streaming URLs requests.

As the extractor doesn't support login with a Google account, you only need to create poTokens from two properties: a visitorData and a videoId (for logged-in users, as written in BgUtils repo, it uses account's dataSyncId).

So maybe we should generate the visitorData at NPE and then pass it to the interface, which will be used by implementations. This also prevents clients to reuse the same visitorData poToken, which is good for privacy but not really for speed and resources consumption.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
AudricV: It was to avoid recomputing the field nullability multiple times (also in a previous version when I was working on this, the `fetchHtml5Client` method content was directly in the `onFetchPage` method if I remember properly), but maybe this should be avoided.

Stypox: I'm pretty sure passing one more parameter around is slower than redoing the null check. Actually, I would guess that `if (o != null)` is as fast as `if (b)`, since both perform a memory load and then use the `zero` flag to determine whether to jump or not (assuming `null` is represented as 0).

Stypox:
I am merging this as it does work well (many users tested it in NewPipe) and we need to make a NewPipe release asap. I will publish the extractor release as a pre-release. Please create PRs to fix/improve/finish this PR and then we can do a proper release.

Issue #1276, opened 2025-02-11
Mock downloader for everything


litetex:
Currently the mock downloader is only used for YT.

We should use it everywhere, so that when e.g. Bandcamp is down or cannot be reached, one can still run the test

PR #1277, opened 2025-02-12
Fix tests and cleanup other stuff

### LINKED ISSUE #1264: Make the tests and CI work again
The tests and CI are broken for a loooonnnng time.

These should be fixed.

Things that should be done:
* Mock/Saved response
  * Fix tests with stored/missing responses
* Real
  * Throttle down YT tests to not run into rate-limits
  * Exclude YT test from scheduled nightly tests since they YT blocks GitHub's runner anyway

litetex:

Fixes #1264

Overview of the changes:
* Implemented Ratelimiting for REAL and RECORDING downloader
  * Currently YT and Bandcamp require this
  * Using a fork of guava's Ratelimiter
  * Removed old "rate-limiter" from RECORDING
* Use a more modern TLS configuration
  * [This fixes Bandcamp tests not working on Windows 10 with Java 17+. Their CDN (Fastly) is likely blocking the outdated offered ciphers.](https://github.com/TeamNewPipe/NewPipeExtractor/commit/351174a85f86658371a82041befa7a1a21104c4e)
* Fixed tests
* Disabled some broken tests that currently have no fixes
  * for example: YT's age-restriced videos
* Re-enabled a few disabled tests
* Updated test dependencies
* [Removed YT channel mixes as these no longer exist](https://github.com/TeamNewPipe/NewPipeExtractor/commit/451bb1315c9171e574f5da527e58fc508fe16b43)
* [Use new Bandcamp autocomplete API](https://github.com/TeamNewPipe/NewPipeExtractor/pull/1277/commits/1966ad1e529594f22ebbb8d35c51302024bb24fc)
* [Removed unused visitorData generation code](https://github.com/TeamNewPipe/NewPipeExtractor/pull/1262#issuecomment-2633421077)
* YT now has 100% mock test coverage. This should be further improved with #1276 in the future
* Cleaned up a ton of smaller code/code style problems
* Updated gradle to latest version (8.12.1) to fix compatibility with newer Java Versions

⚠️ Note:
YT still sometimes randomly fails on ``testMoreItems`` when doing real test (I suspect a new A/B test being rolled out).
I also observed this behavior inside the app but with a very low probability (1:20+)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeStreamLinkHandlerFactory.java
```diff
@@ -114,7 +114,7 @@ public String getId(final String theUrlString)
         try {
             url = Utils.stringToURL(urlString);
         } catch (final MalformedURLException e) {
-            throw new IllegalArgumentException("The given URL is not valid");
+            throw new ParsingException("The given URL is not valid", e);
```
Stypox, line 117: @AudricV is this change ok?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicArtistInfoItemExtractor.java
```diff
@@ -61,10 +62,14 @@ public String getUrl() throws ParsingException {
...
-                .getObject(2)
+        final JsonArray flexColumns = artistInfoItem.getArray("flexColumns");
+        final JsonArray runs = flexColumns
+                .getObject(flexColumns.size() - 1)
                 .getObject("musicResponsiveListItemFlexColumnRenderer")
-                .getObject("text"));
+                .getObject("text")
+                .getArray("runs");
+        final String subscriberCount = runs.getObject(runs.size() - 1)
+                .getString("text");
```
Stypox, line 72: I think you can use YoutubeParsingHelper.getTextFromObject()

Stypox (changes requested):
Thank you for all of these changes! Maybe some of the stuff could have been broken up in multiple PRs but I guess it would have taken too much time. The commits were well separated, I reviewed them one by one and they look good to me. :-)

Small criticism: please put `[YouTube]`/`[BandCamp]`/... at the beginning of service-specific commit messages.

Maybe rate-limiting should be added to NewPipe too, since we currently already do some rate limiting e.g. in the feed manager.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeStreamLinkHandlerFactory.java
AudricV: I have no idea 🤷‍♂️

extractor/src/main/java/org/schabi/newpipe/extractor/playlist/PlaylistInfo.java
```diff
@@ -42,12 +42,6 @@ public enum PlaylistType {
          */
         MIX_MUSIC,
 
-        /**
-         * A mix made only of streams from (or related to) the same channel, for example YouTube
-         * channel mixes
-         */
-        MIX_CHANNEL,
-
```
AudricV, line 50: Are you really sure that channel mixes are gone? Even if they are gone from tests, maybe they aren't on the whole platform. This concept may be applicable to other services too.

This is a breaking change, this should precised in the PR description, so you're also required to add if you tested the changes in NewPipe and if you want to make a PR in the app repo to fix breaking changes, if applicable (from the PR template).

extractor/src/main/java/org/schabi/newpipe/extractor/utils/ProtoBuilder.java
AudricV, line 1: Note: this could be still useful to generate manually continuations of playlists to extract all videos of shorts UI playlists.

#1104 uses Google's protobuf classes directly, so using this class may be a lighter alternative.

If you want to have a look at this, free feel to do so, note the continuation format has slightly changed, but the workaround idea should still work.

AudricV (dismissed):

extractor/src/main/java/org/schabi/newpipe/extractor/playlist/PlaylistInfo.java
Stypox: Yeah it might make sense to keep the code, you never know if YouTube might readd them

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicArtistInfoItemExtractor.java
litetex: That was used there before however I did not choose to use it here.

runs returns an array with 3 items containing:
0: <artistname>
1: •
2: <subscriperCount>

There is 2 problems with that:
1. It's inefficient
2. If for some reason the artist now has a name like "2k" the extraction malfunctions
 
I can add this as a comment to the code if you wish ;)

extractor/src/main/java/org/schabi/newpipe/extractor/playlist/PlaylistInfo.java
litetex: I just now noticed that this is inside ``org.schabi.newpipe.extractor.playlist`` and not in inside ``youtube``. Thought it was in ``youtube`` (because these mix type only exist for YT and nothing else).

I will reintroduce the enum but deprecate it.

extractor/src/main/java/org/schabi/newpipe/extractor/utils/ProtoBuilder.java
litetex: That sounds like it should be outsourced into a new issue to me.

I also think it's not useful to keep this unused code as it also has 0 documentation on what it does. This was [already highlighted](https://github.com/TeamNewPipe/NewPipeExtractor/pull/1262#pullrequestreview-2565796918) a month ago and we never got a followup PR :(

Otherwise the [same arguments as above](https://github.com/TeamNewPipe/NewPipeExtractor/pull/1277#discussion_r1955169657) also apply here.

litetex:
@Stypox @AudricV 
I think if there are no further objections this can be merged so that we can go on and fix other stuff :)

Stypox (approved):
> I can add this as a comment to the code if you wish ;)

Yes please :-)

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicArtistInfoItemExtractor.java
litetex: Added the comment 👍

Stypox (approved):

PR #1279, opened 2025-02-15
Only include ``jsr305`` and not complete spotbugs

### LINKED ISSUE #1278: Remove spotbugs
SpotBugs is currently used and shipped as a dependency, however we only required ``javax.annotation.Nonnull`` or ``javax.annotation.Nullable``.
These are provided by [``jsr305``](https://mvnrepository.com/artifact/com.google.code.findbugs/jsr305/3.0.2) and not SpotBugs.

In the future this should be potentially replaced by [``jakarta-annotations``](https://projects.eclipse.org/projects/ee4j.ca) which is the successor to this topic.

litetex:
Fixes #1278

Only ``javax.annotation.Nonnull`` or ``javax.annotation.Nullable`` are currently in used which means that spotbugs can and should be removed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,7 +28,7 @@ allprojects {
 
     ext {
         nanojsonVersion = "1d9e1aea9049fc9f85e68b43ba39fe7be1c1f751"
-        spotbugsVersion = "4.8.6"
+        jsr305Version = "3.0.2"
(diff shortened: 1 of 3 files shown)


Stypox (approved):
Seems to work as intended:
image

PR #1288, opened 2025-03-26
Remove usage of reflection in timeago-parser module

thigg:
We are trying to use the NewPipe extractor library for a native app for the mobile linux OS SailfishOS.
For this we intend to use graal's native-image.
When reflection is used, this needs an entry in a json file for each of the classes that may be used with reflection.

I created the put lines for the hashmap with a simple bashscript.
I think this should be totally equal (in terms of functionallity and maintainability) to the current solution without the drawback (for us) that it uses reflection.


Stypox:
All of the java files under the `patterns/` folder were autogenerated by https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/timeago-parser/raw/java/GeneratePatternClasses.java, could you instead change that class to also autogenerate a java file which contains the hashmap, instead of manually hardcoding it?

dvag-thilo-kogge:
Sure, sounds fun

thigg:
@Stypox I updated the PR to generate the map with the generator.
While working at the generator I also moved it into its own gradle submodule, I hope this was okay?

thigg:
removed the line

timeago-generator/src/main/java/com/github/TeamNewPipe/Utils.java
```diff
@@ -1,3 +1,5 @@
+package com.github.TeamNewPipe;
```
Stypox: The package name should be `org.schabi.newpipe.timeago_generator` I believe (also in the other `timeago-generator` classes).

timeago-parser/src/main/java/org/schabi/newpipe/extractor/timeago/PatternMap.java
```diff
@@ -0,0 +1,100 @@
...
+        patternMap.put("ur", ur.getInstance());
+        patternMap.put("uz", uz.getInstance());
+        patternMap.put("vi", vi.getInstance());
+        patternMap.put("zh_CN", zh_CN.getInstance());
+        patternMap.put("zh_HK", zh_HK.getInstance());
+        patternMap.put("zh_TW", zh_TW.getInstance());
+        patternMap.put("zu", zu.getInstance());
+    }
+
+    public static PatternsHolder getPattern(String languageCode) {
```
Stypox:
```suggestion
    public static PatternsHolder getPattern(final String languageCode) {
```

Stypox (approved):
Thank you! I tested importing the project in Android Studio and running the generator and everything worked without issues. I solved the two comments myself with two commits.

thigg:
Thank you. That was pleasant :)

PR #1317, opened 2025-07-07
[Soundcloud] Add support for AAC 160k unprotected stream and update Creative Commons stream test 

AudricV:

This PR replaces Creative Commons' test in `SoundcloudStreamExtractorTest` and adds support for an AAC HLS stream with an average bitrate of 160kbps, which isn't returned in DRM-protected tracks, of course.

### DIFF
--- a/extractor/src/test/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudStreamExtractorTest.java
+++ b/extractor/src/test/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudStreamExtractorTest.java
@@ -141,15 +141,15 @@ public class SoundcloudStreamExtractorTest {
         @Override public String expectedCategory() { return "Dance"; }
     }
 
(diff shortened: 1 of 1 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
```diff
@@ -240,7 +240,9 @@ private void extractAudioStreams(@Nonnull final JsonArray transcodings,
                         } else if (preset.contains("opus")) {
                             builder.setMediaFormat(MediaFormat.OPUS);
                             builder.setAverageBitrate(64);
-                            builder.setDeliveryMethod(DeliveryMethod.HLS);
```
TobiGr, line 243: Was this removed because it is already covered by the protocol check in lines 231-233?

AudricV: Exactly.

PR #1319, opened 2025-07-07
[PeerTube] Detect video-only streams

AudricV:

This PR detects video-only streams of PeerTube, introduced in [the version 6.3.0 of the software](https://github.com/Chocobozzz/PeerTube/releases/tag/v6.3.0).

This is the extractor part of TeamNewPipe/NewPipe#11601.


extractor/src/main/java/org/schabi/newpipe/extractor/services/peertube/extractors/PeertubeStreamExtractor.java
```diff
@@ -651,12 +651,13 @@ private void addNewVideoStream(@Nonnull final JsonObject streamJsonObject,
         final String extension = url.substring(url.lastIndexOf(".") + 1);
         final MediaFormat format = MediaFormat.getFromSuffix(extension);
         final String id = resolution + "-" + extension;
+        final boolean isVideoOnly = !streamJsonObject.getBoolean("hasAudio");
```
TobiGr: is that backwards compatible? What happens if this field is not present? Looks to me like the file would be considered to be videoOnly then. Not sure if this is intended. LGTM apart from that

AudricV: It wasn't indeed, I fixed this by checking for the presence of the `hasAudio` field, but we could also use `getBoolean("hasAudio", true)`.

PR #1322, opened 2025-07-08
Validate HTTP response codes across the extractor [1]

absurdlylongusername:

### Initial PR to introduce validation of HTTP response codes for network requests throughout the extractor

(Part of #1472)

Almost all code that uses the Downloader does not do any checking on the response code: therefore errors and exceptions thrown due to error response code give false information (e.g. tests failing due to 404 were giving error about `clientId` when the error was 404).

TL;DR this PR adds methods in Response.java for validating response codes and uses it in some places for soundcloud extraction. These methods throw exceptions with information about the error code, and propagate upwards to the call site to print a more useful stack trace for debugging.

Read discussions below for full context.

I don't like chonky PRs so I will not do too much refactoring in this PR, but will split it up into multiple PRs


TobiGr (commented):
Thaank you. Changes LGTM, but conflicts need to be resolved first.

litetex (dismissed):
I don't think this can be merged in the current state:
* There are Todo comments
* The changes in the test will be irrelevant after #1332 
* Response codes that are ok are 2XX response codes
* I'm really not sure about another HttpUtils method and a new Exception class

I think the most simple approach would be to have a ``boolean Response#isOk()`` method and then use this method to control the underlying use cases accordingly.

absurdlylongusername:
@litetex 

> - There are Todo comments

The comment was kind of intentionally left in there in the hopes that whoever reviewed it would provide an explanation and then I would remove it.

> - Response codes that are ok are 2XX response codes

Not exactly sure what you mean by this? Do you mean that 2xx is the only "success" code range all network calls throughout the codebase?

>  - I'm really not sure about **another** HttpUtils method and a new Exception class
> I think the most simple approach would be to have a boolean Response#isOk() method and then use this method to control the underlying use cases accordingly.

Do you mean another as in having two `validateResponseCode` methods, one in `Response` and one in `HttpUtils`, or do you mean that there's another `HttpUtils` class somewhere (because I can't seem to find one)?

And does this also mean we shouldn't have a `HttpResponseException`? If so, why? I think having a specific exception for Http errors is better than throwing a generic `IOException`, and also makes for better logging and diagnostics, because right now our code does not give any useful information at all for what causes an error if the cause is an invalid response.

* * *
My idea _was_ to have a generic method we can use to check for expected response codes instead of just checking it's 2xx or not 2xx, because I was of the assumption that:

1. We are definitely not validating response codes anywhere in the codebase (I can't see anywhere we do this)
2. There may be scenarios where we want to treat a specific 3xx or 4xx as not an error and it's not efficient to catch an exception in these cases and _then_ check the response code (and then throw _another_ exception with info about the http error that will get propagated upwards), as I believe we should never throw exception if we don't have to, and exceptions should only be thrown for unexpected behaviour, or in the same scenarios we already throw exceptions.


The end goal is that **if a network call ever fails, we want to know why**, using code that is:

- As small as possible
- Easy to understand
- Easy to use
- Easy to extend

If there's only one place in code where non-2xx codes shouldn't be considered an error then we could handle that case explicitly at that call site, but if it's at least 3 times then I think we should have more than a simple `isOk` method. 
Could perhaps have other methods like Response#isRedirect for 3xx, Response#isClientError for 4xx, Response#isServerError instead of a single generic method.

Since I am almost certain 1. is true, I am more than happy to trawl through the entire codebase for every http call we make (not jusr for `get`, but for `post` and `head` too) and add response code validation and see which approach makes the most sense.

I could update this PR with these changes and/or split it into multiple PRs if it is too big.

Thoughts?

extractor/src/main/java/org/schabi/newpipe/extractor/utils/HttpUtils.java
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Validates the response codes for the given {@link Response}, and throws 
+     * a {@link HttpResponseException} if the code is invalid
+     * @param response The response to validate
+     * @param validResponseCodes Expected valid response codes
+     * @throws HttpResponseException Thrown when the response code is not in {@code validResponseCodes},
+     * or when {@code  validResponseCodes} is empty and the code is a 4xx or 5xx error.
+     */
+    // CHECKSTYLE:ON
+    public static void validateResponseCode(final Response response,
```
Stypox: I would inline this directly in `Response`, especially because it's first argument is a `Response`

extractor/src/test/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudChannelTabExtractorTest.java
Stypox, line 1: Yeah changes in this file should be reverted as they were superseded

extractor/src/main/java/org/schabi/newpipe/extractor/exceptions/HttpResponseException.java
```diff
@@ -0,0 +1,15 @@
+package org.schabi.newpipe.extractor.exceptions;
+
+import java.io.IOException;
+import org.schabi.newpipe.extractor.downloader.Response;
+
+public class HttpResponseException extends IOException {
```
Stypox: This should be an `ExtractorException`, since the extractor received an unexpected response. There was no error in data input/output, since the request went through just fine at the network layer.

extractor/src/main/java/org/schabi/newpipe/extractor/downloader/Response.java
```diff
@@ -80,4 +83,21 @@ public String getHeader(final String name) {
 
         return null;
     }
+    // CHECKSTYLE:OFF
```
Stypox: Why checkstyle off?

Stypox (changes requested):
Thanks! We do need to check response codes more often. For example, PeerTube sometimes throws 429 Too Many Requests (https://github.com/TeamNewPipe/NewPipe/issues/2979), but that shouldn't be treated as a recaptcha by the `Downloader` (while that's what happens now). It should be the PeerTube service's job to throw a normal ExtractorException for 429s, and instead the YouTube service's job to throw a RecaptchaException (along with a proper URL that users can use to solve the captcha, instead of having the `Downloader` guess that URL).

It would be good if you could take the above paragraph into consideration for this PR, and properly validate the HTTP response codes of all requests made by all services in the extractor, so we don't have to change the design once again later.

extractor/src/main/java/org/schabi/newpipe/extractor/downloader/Response.java
absurdlylongusername: IIRC, checkstyle was complaining about line being too long in the comment below, so I had to disable it

extractor/src/main/java/org/schabi/newpipe/extractor/exceptions/HttpResponseException.java
absurdlylongusername: So, problem with this. If this is changed to `ExtractionException`, then that means any code that validates HttpResponseException will also now need to have `throws ExtractionException`.

Is this something we want or not?

Stypox: Yeah I think this would be ok

extractor/src/main/java/org/schabi/newpipe/extractor/downloader/Response.java
Stypox: Not really important, but why don't you make the lines shorter then? Here you don't have long URLs

extractor/src/main/java/org/schabi/newpipe/extractor/downloader/Response.java
```diff
@@ -30,6 +33,29 @@ public Response(final int responseCode,
...
+    /**
+     * Validates the response codes for the given {@link Response}, and throws
+     * a {@link HttpResponseException} if the code is invalid
+     * @param response The response to validate
+     * @param validResponseCodes Expected valid response codes
+     * @throws HttpResponseException Thrown when the response code is not in {@code validResponseCodes},
+     * or when {@code  validResponseCodes} is empty and the code is a 4xx or 5xx error.
+     */
+    // CHECKSTYLE:ON
+    public static void validateResponseCode(final Response response,
```
Stypox: Why not inline this directly within the non-static `Response::validateResponseCode`? This is basically a method on the `Response` class since it takes `Response` as the first argument, so you might as well make it non-static.

extractor/src/main/java/org/schabi/newpipe/extractor/downloader/Response.java
```diff
@@ -80,4 +106,22 @@ public String getHeader(final String name) {
...
+     * before getting the code/body/latestUrl/etc.
+     * Validates the response codes for the given {@link Response}, and throws a {@link HttpResponseException} if the code is invalid
+     * @see Response#validateResponseCode(Response, int...)
+     * @param validResponseCodes Expected valid response codes
+     * @return {@link this} response
+     * @throws HttpResponseException Thrown when the response code is not in {@code validResponseCodes},
+     * or when {@code  validResponseCodes} is empty and the code is a 4xx or 5xx error.
+     */
+    // CHECKSTYLE:ON
+    public Response validateResponseCode(final int... validResponseCodes)
```
Stypox: In order to support the needs of other services (like youtube), I'd add the following methods:
```java
public Response throwIfResponseCode(int errorCode, final Function<Response, HttpResponseException> errorSupplier);
public Response throwIfResponseCodeInRange(final int errorCodeMin, final int errorCodeMax, final Function<Response, HttpResponseException> errorSupplier);
```

I'd also rename this method to `throwIfInvalidResponseCode()`

Then e.g. for YouTube we'd be able to do:
```java
response
    // "https://www.youtube.com" is the URL to solve the captcha
    .throwIfResponseCode(429, r -> new ReCaptchaException(r, "https://www.youtube.com"))
    .throwIfInvalidResponseCode()
```

While in PeerTube

```java
response
    .throwIfResponseCode(429, r -> new TooManyRequestsException(r))
    .throwIfInvalidResponseCode()
```

And we'd remove the `ReCaptchaException` checks in the `Downloader`. Also we'd have `ReCaptchaException` and `TooManyRequestsException` extend `HttpResponseException`.

What do you think?

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
```diff
@@ -189,6 +194,7 @@ public static String resolveUrlWithEmbedPlayer(final String apiUrl) throws IOExc
      *
      * @return the resolved id
      */
+    // TODO: what makes this method different from the others? Don' they all return the same?
```
Stypox: Which method would be the same as which others? They all do different things imo

Stypox (changes requested):

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
absurdlylongusername: At the time, I didn't have proper understanding of how the other resolve methods worked. But now I do. This resolve using the widget and the resolve for above using the normal API both return exactly the same thing, but this one just filters for the `Id`. 

This one also doesn't take a `Downloader` though. I believe there was some reason somewhere for using widget API alongside the api-v2 endpoint, but I can't remember.

extractor/src/main/java/org/schabi/newpipe/extractor/downloader/Response.java
absurdlylongusername: My intention was to allow for `response.validateResponseCode` and `validateResponseCode(response)` depending on the format of the call site.

But in hindsight, I now recall that in C# (my main language) there is [`EnsureSuccessStatusCode`](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpresponsemessage.ensuresuccessstatuscode?view=net-9.0) which can be used on the response itself, rather than a separate method.

So yeah I think you're right, validation can just be done inline like how you have mentioned in your other comment

extractor/src/main/java/org/schabi/newpipe/extractor/downloader/Response.java
absurdlylongusername: Will do

extractor/src/main/java/org/schabi/newpipe/extractor/downloader/Response.java
absurdlylongusername: I like this.

However, I think `throwIfInvalidResponseCode` should be called `ensureValidResponseCode` or `ensureSuccessResponseCode` if it's going to be the same as `validateResponseCode` which takes in a list of valid response codes. Consider these:

- `throwIfInvalidResponseCode(200, 201)` 
- `ensureValidResponseCode(200, 201)`

Reading the second, it is easier to understand that 200 and 201 are the valid response codes, but one could presume with the first that 200 and 201 are _invalid_ response codes and it should throw if one of those are returned.

Thoughts?

absurdlylongusername:
> Thanks! We do need to check response codes more often. For example, PeerTube sometimes throws 429 Too Many Requests ([TeamNewPipe/NewPipe#2979](https://github.com/TeamNewPipe/NewPipe/issues/2979)), but that shouldn't be treated as a recaptcha by the `Downloader` (while that's what happens now). It should be the PeerTube service's job to throw a normal ExtractorException for 429s, and instead the YouTube service's job to throw a RecaptchaException (along with a proper URL that users can use to solve the captcha, instead of having the `Downloader` guess that URL).
> 
> It would be good if you could take the above paragraph into consideration for this PR, and properly validate the HTTP response codes of all requests made by all services in the extractor, so we don't have to change the design once again later.

@Stypox Some problems.

In [`BandcampChannelLinkHandlerFactory.getId`](https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/main/java/org/schabi/newpipe/extractor/services/bandcamp/linkHandler/BandcampChannelLinkHandlerFactory.java#L33-L48) if I want to add HTTP response code validation, then this means that I need to handle the HttpResponseException.

Now, `LinkHandlerFactory.getId` and other base methods all throw `ParsingException`, but not `ExtractionException`. `getId` in the bandcamp class catches `IOException | ReCaptchaException | ArrayIndexOutOfBoundsException | JsonParserException` and swallows them all into a `ParsingException`. 

Semantically, only `JsonParserException` is a parsing exception, and I don't know the rationale behind catching `ArrayIndexOutOfBoundsException`.

Now, what exactly is the difference between an `ExtractionException` and a `ParsingException`?

`LinkHandlerFactory` and pretty much all the other abstract base class extractors and similar classes all throw `ParsingException`, and they do that by swallowing up all other exceptions and wrapping it in a `ParsingException`. 

If we add http response validation throughout the extractor, should the `HttpResponseException` be swallowed and wrapped in a `ParsingException`? Because as far as I can see that is what happens throughout the codebase regardless of the type of exception thrown when we make network requests. 

How exactly do we want to do this, so that we can have correct semantic exceptions?

extractor/src/main/java/org/schabi/newpipe/extractor/downloader/Response.java
Stypox: `ensureValidResponseCode` is a better name indeed

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
Stypox: If you want to add a comment with your findings feel free to :-)

Stypox:
> Now, what exactly is the difference between an ExtractionException and a ParsingException?

I guess this was a very early attempt to distinguish exceptions caused by getting and deserializing data from the internet, vs parsing that data using the methods defined in the extractor. However, I don't think that distinction holds now, and the there is no semantic distinction between the two. If you want to open a PR to delete `ParsingException` and replace it with `ExtractionException` everywhere feel free to.

> If we add http response validation throughout the extractor, should the HttpResponseException be swallowed and wrapped in a ParsingException? Because as far as I can see that is what happens throughout the codebase regardless of the type of exception thrown when we make network requests.

I would create a `NetworkException extends ExtractionException` to wrap all network-related exception. Error handling is generally difficult to do, but I think the current state of the extractor mostly works as expected: wrapping exceptions means we still get to look at the original exception in the stacktrace in bugreports, but the API remains clean so NewPipeExtractor users don't have to `catch` tens of exception types.

absurdlylongusername:
@Stypox @TobiGr I've addressed the outstanding code review comments.

I want to get this PR merged ASAP so #1325 can be merged. So in this PR I just want to do initial setup of helper methods for `Response.java`, and adding response code validation in `SoundcloudParsingHelper`.

I added a bunch of initial helper methods that came to mind. I forgot to add `ensureValidResponseCodes` and `throwIfInvalidResponseCodes` that take vararg array of codes. 
Idk if all of them will be necessary anyway: as I add validation thoroughout the codebase in later PRs I will add/remove methods as needed, so perhaps `ensureValidResponseCodes` and `throwIfInvalidResponseCodes` may never be needed at all.

I did not do bounds checking on `min/max` in the range validation methods (I forgot). I will add them in the next PR, but I can add them in this one if you think it's very important enough.

Things I will do in next PR:

- Add tests for `Response` and update tests for `SoundcloudParsingHelper` to test validation
- Remove `ParsingException` and replace with ExtractorException
- Make `HttpResponseException` extends `NetworkException` extends `ExtractorException`
- Add some more validation in places, and add/update tests accordingly


Things worthy of note:

- I have not updated the mocks. I don't really know how: I know how to run the recording downloader, but I don't get how it works, because when I ran it to update mocks for failing tests it looks like it deleted all generated mocks for all the tests I didn't run for that specific folder (soundcloudparsinghelper).
- What region should mocks be run on? E.g. if I use VPN location in Spain or Sweden or Singapore, does it matter as long as the tests pass?
- I have not tested these (recently pushed) changes in NewPipe yet. Likely, if I do proper static analysis I could conclude that it's not possible for the code to actually introduce any bugs since in every case it throws an exception for invalid response code, it would throw an exception regardless: but I cannot confirm this. Also I haven't done unit tests yet either so can't confirm it for that reason too.

extractor/src/main/java/org/schabi/newpipe/extractor/downloader/Response.java
```diff
@@ -80,4 +84,129 @@ public String getHeader(final String name) {
...
+        return this;
+    }
+
+    public Response ensureResponseCodeNotInRange(
+        final int min,
+        final int max,
+        final Function<Response, HttpResponseException> errorSupplier
+    )
+        throws HttpResponseException {
+        if (min > max) throw new RuntimeException("min must be less than max");
```
TobiGr: Why `RuntimeException` and not an `IllegalArgumentException`?


```suggestion
        if (min > max) throw new IllegalArgumentException("min must be less than max");
```

extractor/src/main/java/org/schabi/newpipe/extractor/downloader/Response.java
```diff
@@ -80,4 +84,129 @@ public String getHeader(final String name) {
...
+     */
+    public Response throwIfClientError()
+        throws HttpResponseException {
+        return throwIfClientError(HttpResponseException::new);
+    }
+
+    /**
+     * Throw exception if response code is a 5xx server error
+     * @return this {@code Response}
+     * @throws HttpResponseException if the response code is 4xx
```
TobiGr:
```suggestion
     * @throws HttpResponseException if the response code is 5xx
```

TobiGr (approved):
THank you. Can be merged once these two small things are applied

TobiGr:
@Stypox Do you want to review again or can we merge the PR? Your comments have been addressed.

PR #1323, opened 2025-07-08
[SoundCloud] Miscellaneous refactors

absurdlylongusername:

Some minor refactorings

- Add toString to DateWrapper for better debugging
- make SoundcloudStreamExtractor.getTimeStamp return 0 if not timestamp found in the url
- Use Pattern instead of String in regex for SoundcloudStreamLinkHandlerFactory
- Refactor Parser.java
- Refactor AudioStream constructor to take builder instead of many parameters
- Other minor changes

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/stream/Stream.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/stream/Stream.java
@@ -68,7 +68,7 @@ public abstract class Stream implements Serializable {
      * @param streamList the list of {@link Stream}s which will be compared
      * @return whether the list already contains one stream with equals stats
      */
-    public static boolean containSimilarStream(final Stream stream,
+    public static boolean containSimilarStream(@Nonnull final Stream stream,
                                                final List<? extends Stream> streamList) {
         if (isNullOrEmpty(streamList)) {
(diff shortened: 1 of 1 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
```diff
@@ -121,7 +121,8 @@ public long getLength() {
 
     @Override
     public long getTimeStamp() throws ParsingException {
-        return getTimestampSeconds("(#t=\\d{0,3}h?\\d{0,3}m?\\d{1,3}s?)");
+        final var timestamp = getTimestampSeconds("(#t=\\d{0,3}h?\\d{0,3}m?\\d{1,3}s?)");
+        return timestamp == -2 ? 0 : timestamp;
```
Stypox, line 125: For a future PR, you could make `getTimestampSeconds()` return an `Optional<>`. Also, all regex-related functions should ideally only take `Pattern`s, so we can avoid rebuilding regexes every time.

Stypox (approved):
Looks good to me, thank you!

PR #1324, opened 2025-07-08
[Soundcloud] Initial soundcloud test refactoring

absurdlylongusername:

## Add test cases for SoundcloudStreamExtractorTest

This PR makes initial progress on refactoring testing infrastructure to use test cases to generate tests instead of current architecture that uses a class to define each test case.

It uses [Immutables](https://immutables.github.io/) to define test cases. `ISoundcloudStreamExtractorTestCase` is the test case used for SoundcloudStreamExtractorTest, which inherits from other default extractor test cases.

In future PRs this will be expanded on to further decouple the current test format so we can have tests that can run multiple test cases instead of defining a new class for each test case. This will make tests easier to change, more versatile and extensible.

### Update tests in SoundcloudStreamExtractorTest

Changed how the tests for SoundCloud CDN urls worked to use a regex instead of matching the ID explicitly.

## Please Note

This PR includes changes from #1323 which couldn't be removed as it depends on it. However, since those changes don't depend on _this_, I have included them in that other PR so they can be merged first before this one.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,6 +31,7 @@ allprojects {
         jsr305Version = "3.0.2"
         junitVersion = "5.13.3"
         checkstyleVersion = "10.4"
+        immutablesVersion  = "2.10.1"
     }
 }
 
--- a/extractor/build.gradle
+++ b/extractor/build.gradle
@@ -46,4 +46,10 @@ dependencies {
 
     testImplementation "com.squareup.okhttp3:okhttp:4.12.0"
     testImplementation 'com.google.code.gson:gson:2.13.1'
+    testImplementation "org.immutables:value:$immutablesVersion"
+    testAnnotationProcessor "org.immutables:value:$immutablesVersion"
+
+}
+repositories {
+    mavenCentral()
 }
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/localization/DateWrapper.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/localization/DateWrapper.java
@@ -69,4 +69,12 @@ public class DateWrapper implements Serializable {
     public boolean isApproximation() {
         return isApproximation;
     }
+
+    @Override
+    public String toString() {
+        return "DateWrapper{"
+                + "offsetDateTime=" + offsetDateTime
+                + ", isApproximation=" + isApproximation
+                + '}';
+    }
 }
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudStreamExtractor.java
@@ -121,7 +121,8 @@ public class SoundcloudStreamExtractor extends StreamExtractor {
 
     @Override
     public long getTimeStamp() throws ParsingException {
-        return getTimestampSeconds("(#t=\\d{0,3}h?\\d{0,3}m?\\d{1,3}s?)");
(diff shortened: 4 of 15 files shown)


absurdlylongusername:
@Stypox In hindsight, yes I do agree it's not worth doing this atm purely due to the fact it is not a priority right now, not absolutely necessary, and a lot of this test code is not being used/developed around that much.

That being said, I believe that tests and stuff that we add in the future should use a better pattern rather than the patterns that we use right now.

E.g. #1360 would require refactoring a lot of stuff around Soundcloud, and in turn would mean changing some tests and adding new tests. Ideally, I'd want these new tests to be made in a way where it's not a class per test case. Would you agree?

-----

Also this kind of begs the question: how often do we need tests that have many test cases? Because we do indeed have some tests in our codebase that are structured like this and have several test cases (e.g. [SoundcloudSteamLinkHandlerFactoryTest](https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/test/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudStreamLinkHandlerFactoryTest.java)).

IIRC, my main rationale for wanting to do this refactor was that all tests should be extensible (in the dimension of being able to add more test cases) by default (i.e. it is the same test with multiple test cases, not a new test with @Test annotation); but, if, realistically, most of the tests will not require this functionality then it's not much of a reward doing this refactor everywhere.

If tests with many test cases was a requirement or design concern, then the current test infrastructure that I want to refactor wouldn't have been designed the way it is. So, it makes sense for me to conclude that for most or all of the tests that have this current structure do not really require many test cases in order to be tested approriately. 

So, practically, imo, if anything is going to be refactored it should only be done if the tests actually need to be expanded in the dimension of needing many test cases for tests, OR if that part of the code is already going to undergo a large enough structural change that refactoring it in the process can easily be done in parallel without much overhead. 

Given this, this PR doesn't meet either of these criteria so it is not necessary and therefore can be abandoned.

Stypox:
> Also this kind of begs the question: how often do we need tests that have many test cases?


> If tests with many test cases was a requirement or design concern, then the current test infrastructure that I want to refactor wouldn't have been designed the way it is. So, it makes sense for me to conclude that for most or all of the tests that have this current structure do not really require many test cases in order to be tested approriately.

I would say that for remote content (videos, playlists, etc.) we don't usually need tests with many test cases:
- usually we write tests for specific behaviors of services (e.g. a video being age restricted, not available, with/without comments, etc.), so it would not be very helpful to test the same behavior multiple times, as the format of data going in/out would basically be the same. Whenever we notice that the format of data differs (which usually means adding new conditions somewhere in an extractor), we usually also add a test.
- already with the current amount of tests there is a considerable amount of maintenance, since whenever video authors change the title/description/... we also need to update the tests right after updating mocks. So tripling or quadrupling the amount of test cases would mean having to spend even more time updating mocks, fixing test case data, etc. 

> E.g. https://github.com/TeamNewPipe/NewPipeExtractor/issues/1360 would require refactoring a lot of stuff around Soundcloud, and in turn would mean changing some tests and adding new tests. Ideally, I'd want these new tests to be made in a way where it's not a class per test case. Would you agree?

I think that for link handlers it makes a lot of sense to have test cases, because the link handler functions to test often have just a couple of inputs and outputs, and always the same amount thereof. Videos on the other hand have a lot of fields that should just default to some value, and we'd need the builder pattern to achieve something like that, which is already more verbose than just a list of test cases. But yeah we should definitely use test case lists (I don't know how to call it, maybe it's "instanced tests"?) for link handlers and other input->output functions.

> So, practically, imo, if anything is going to be refactored it should only be done if the tests actually need to be expanded in the dimension of needing many test cases for tests, OR if that part of the code is already going to undergo a large enough structural change that refactoring it in the process can easily be done in parallel without much overhead.

This sounds like a good plan :-)

PR #1325, opened 2025-07-08
[SoundCloud] Fix soundcloud hls expiry and add extractor logging

davidasunmo:

### Fix SoundCloud HLS stream urls expiring

This PR implements extractor side changes to facilitate refreshing expires HLS stream URLs for SoundCloud streams.

### Add Extractor Logging

Added extractor logging, which helped a lot to fix this issue.

In a future PR I will add a proper logging framework that will only print logs if running in debug mode, but for now since the logs are not in high-traffic code then it should be fine.

## Please Note

Includes changes from:
- #1322 
- #1323 

So those PRs must be merged before this.

**Write up for the fix will be in the NewPipe PR**


extractor/src/main/java/org/schabi/newpipe/extractor/utils/ExtractorLogger.java
```diff
@@ -0,0 +1,48 @@
...
+        logger = customLogger;
+    }
+
+    public static void d(final String tag, final String msg) {
+        logger.debug(tag, msg);
+    }
+
+    public static void w(final String tag, final String msg) {
+        logger.warn(tag, msg);
+    }
```
Stypox: These should also optionally take a `Throwable` (that's how they work in Android, and it's often useful)

extractor/src/main/java/org/schabi/newpipe/extractor/stream/RefreshableStream.java
```diff
@@ -0,0 +1,15 @@
+package org.schabi.newpipe.extractor.stream;
+
+import org.schabi.newpipe.extractor.exceptions.ExtractionException;
+
+import javax.annotation.Nonnull;
+import java.io.IOException;
+
+@SuppressWarnings("checkstyle:LeftCurly")
+public interface RefreshableStream  {
```
Stypox, line 9:
```suggestion
public interface RefreshableStream {
```

extractor/src/main/java/org/schabi/newpipe/extractor/utils/ExtractorLogger.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public static void e(final String tag, final String msg) {
+        logger.error(tag, msg);
+    }
+
+    public static void e(final String tag, final String msg, final Throwable t) {
+        logger.error(tag, msg, t);
+    }
+
+    // default logger that prints to stdout
```
TobiGr:
```suggestion
    /**
     * Default logger that prints to stdout.
     */
```

extractor/src/main/java/org/schabi/newpipe/extractor/StreamingServiceId.java
Stypox, line 1: Is this file used anywhere? Also, the reason why services are not implemented as an enum is also because it should be possible to dynamically add service implementations (think e.g. of plugins). Obviously this has never happened yet, but iirc that's why the API is like that.

extractor/src/main/java/org/schabi/newpipe/extractor/utils/ExtractorLogger.java
```diff
@@ -0,0 +1,48 @@
...
+    private static final class ConsoleLogger implements Logger {
```
Stypox: This should be only set by users of the library and should default to a logger that logs nothing. In each extractor test then you would set this `ConsoleLogger` as the logger (i.e. you just add this in the common tests initialization methods).

extractor/src/main/java/org/schabi/newpipe/extractor/stream/HlsAudioStream.java
```diff
@@ -0,0 +1,60 @@
+package org.schabi.newpipe.extractor.stream;
+
+import org.schabi.newpipe.extractor.exceptions.ExtractionException;
+
+import java.io.IOException;
+
+import javax.annotation.Nonnull;
+
+public class HlsAudioStream extends AudioStream implements RefreshableStream {
```
Stypox, line 9: I don't think it's a property of all HLS streams that they are refreshable. I think we have two alternative ways we can handle this instead:
- `RefreshableStream` could just be an interface that some `Stream`s can implement, then each `Service` would override `VideoStream/AudioStream` and also implement `RefreshableStream` if needed, and the player could check for refreshable streams using `instanceof`. This would require a lot of boilerplate though.
- Add a `(Audio/Video)Stream refresh() {}` method to the base `Stream` class which builds a new refreshed stream, and make it return `null` if refreshing is not available.

extractor/src/main/java/org/schabi/newpipe/extractor/stream/RefreshableStream.java
Stypox, line 9: Shouldn't this also somehow store the expiry time? So that the player can prepare in advance if the stream is about to expire

extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudParsingHelper.java
```diff
@@ -189,6 +199,7 @@ public static String resolveUrlWithEmbedPlayer(final String apiUrl) throws IOExc
      *
      * @return the resolved id
      */
+    // TODO: what makes this method different from the others? Don' they all return the same?
```
Stypox, line 203: Maybe you can get an answer by git blaming?

absurdlylongusername: I've found answer to this question as per https://github.com/TeamNewPipe/NewPipeExtractor/pull/1322#discussion_r2415284403

extractor/src/main/java/org/schabi/newpipe/extractor/stream/RefreshableStream.java
absurdlylongusername: How exactly would the player prepare in advance? What is it supposed to do?

extractor/src/main/java/org/schabi/newpipe/extractor/StreamingServiceId.java
absurdlylongusername: It's used in the [NewPipe PR](https://github.com/TeamNewPipe/NewPipe/pull/12418/files) for logging the names of services.

It's possible to have enums and still allow for dynamic implementations. Since plugins are a long way away, I think it makes a lot more sense to have enums for each service to improve readability, traceability, logging and type safety.

Howbeit, I'm not going to change the entire API to use enums because that's long: I just want the nice id -> name static method.

absurdlylongusername:
@Stypox
 
> Regarding logging, if I understand correctly the commit structure, it should be possible to extract it in a separate PR, that we can merge much faster. Could you put the logging in a separate PR? Also the

Seems you didn't finish your sentence there lol.

Yeah sure I can make a separate logging PR that can be merged and then rebase this PR on top of it

extractor/src/main/java/org/schabi/newpipe/extractor/StreamingServiceId.java
Stypox: Can't you use this utility method instead? https://github.com/TeamNewPipe/NewPipe/blob/2a9c6f05387866bef2c42ab67226c14936917d27/app/src/main/java/org/schabi/newpipe/util/ServiceHelper.java#L138

But yeah I agree this is not the best. We should not have the concept of service ID at all except for saving stuff to database, so every `StreamInfo` should not have any `getServiceId()` method but rather `getService()` directly. So we don't have to keep switching back and forth between `int` and `Service`

extractor/src/main/java/org/schabi/newpipe/extractor/stream/RefreshableStream.java
Stypox: If the currently playing stream expires at 18:50:45 and now it's 18:50:35, there are just 10 more seconds left to fetch data before needing to switch over to the refreshed stream. So instead of waiting until the URL returns 403 and then quitting, the player can notice that only 10 seconds are left and already pre-fetch the refreshed stream and switch to it as soon as it's ready.

Stypox:
I have no idea what I wanted to say now, sorry xD

extractor/src/main/java/org/schabi/newpipe/extractor/stream/RefreshableStream.java
absurdlylongusername: ExoPlayer handles buffering of stream data, so I don't think is necessary since stream chunks are already pre-fetched ahead of time.

It fetches chunks ahead of time, so when it gets a 403 it's not for the currently playing chunk.

Stypox: Oh right, that's true. I'd still add a nullable expiration `Instant` just in case other clients need it (or maybe NewPipe's Downloader might need it? Or maybe streams in other services that don't throw 403 and where it's more difficult to detect that the stream is now invalid?), since this is information the extractor usually knows. But this is not totally necessary, so do it only if it seems simple to extract such data.

absurdlylongusername:
@Stypox I've now created a separate PR for logging which this PR depends on: #1403

TobiGr:
what about the linked PRs? should those be merged before this one? If yes, the comments left there should be addresesd.

PR #1330, opened 2025-07-10
[YouTube] Speed up n parameter presence check

litetex:
This code does a quick check if the n parameter is even present before running the extraction regex.

The time required to extract an url that does not contain the n-parameter got down from:
* 4237900ns (initial regex compile included) to 4500ns → ~900x faster
* 51200ns to 600ns → ~85x faster


TobiGr:
Good catch! You might want to add a comment to `THROTTLING_PARAM_PATTERN` that this check needs to be modified in case the pattern changes.

litetex:
@TobiGr done :)

PR #1320, opened 2025-07-07
[YouTube] Add support for video lockup content type

litetex:
Fixes #1314
Fixes TeamNewPipe/NewPipe#12422 

Todo:
- [x] Address TODO comments
- [x] Requires https://github.com/TeamNewPipe/nanojson/pull/26 please merge this first

---

Todo for @litetex :

- [x] Create an issue to convert ``ParsingException`` to ``RuntimeException`` because otherwise the Java Streaming API sucks (because you can't use it) https://github.com/TeamNewPipe/NewPipeExtractor/issues/1328
- [x] Create an issue to check in code format/styling for IntelliJ (and maybe remove the checkstyle check that always runs when building and only run it in the CI) https://github.com/TeamNewPipe/NewPipeExtractor/issues/1329

---

- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [ ] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).
- [ ] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API.

PR #1344, opened 2025-07-26
[YouTube] Correctly set `uploaderUrl` for lockup content type items

FineFindus:
Fixes an issue, where the uploader URL for related items would be incorrect, due to the `YoutubeChannelLinkHandlerFactory.getUrl` expecting the id with a `channel/` prefix. However, the `browseId` used to extract the channel id is missing this prefix.

Ref: https://github.com/TeamNewPipe/NewPipeExtractor/pull/1320


litetex:
Ok I will consolidate this with #1346

litetex (approved):
Consolidated successfully

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemLockupExtractor.java
```diff
@@ -179,19 +179,38 @@ public String getUploaderName() throws ParsingException {
...
+            .getString("url");
+        if (!isNullOrEmpty(webCommandMetadataUrl)) {
+            return resolveUploaderUrlFromRelativeUrl(webCommandMetadataUrl);
+        }
+
+        throw new ParsingException("Could not get uploader url");
+    }
+
+    private String resolveUploaderUrlFromRelativeUrl(final String url) throws ParsingException {
+        return YoutubeChannelLinkHandlerFactory.getInstance().getUrl("c" + url);
```
AudricV, line 213: Adding `c` results in `https://www.youtube.com/c/@handle` which is invalid! Also `url` should be renamed to something else, as it is not an entire URL.

```suggestion
        return YoutubeChannelLinkHandlerFactory.getInstance().getUrl(url);
```

litetex: > ``https://www.youtube.com/c/@handle`` which is invalid!

~~Well then YouTubeChannelExtractor is broken because it can't resolve ``https://www.youtube.com/@handle`` but successfully resolves ``https://www.youtube.com/c/@handle``.~~

~~Update: Looks like there was something wrong on my end because ``https://www.youtube.com/@handle`` now also works...~~

Update v2: Ok now I'm really confused it sometimes works in the app and sometimes not. I'm not quite sure what's going on...

I will provide a fix for that in a new PR.

litetex: @AudricV 

There was a general problem with channelId resolution.

I created https://github.com/TeamNewPipe/NewPipeExtractor/pull/1347

FineFindus:
> For future reference: Please open an issue or mention me/comment in the original PR/issue if something is broken.
I don't have push notifications on Matrix enabled and only read stuff there sporadically.

Will try to, but I usually don't create issue, if it's just quicker for me to write a fix myself :)
Not sure what Matrix has to do with this?

AudricV:
This is unrelated to you, it was for team members.

PR #1347, opened 2025-07-26
[YouTube] Make channelId resolution follow redirects and fix ``YoutubeStreamInfoItemLockupExtractor`` building incorrect handles

### LINKED ISSUE #1345: YoutubeStreamInfoItemLockupExtractor: getUploaderUrl not resolving correct url
Followup for #1320 

``YoutubeStreamInfoItemLockupExtractor#getUploaderUrl`` is currently not working properly and creates non resolveable urls

litetex:
Followup of #1344

Fixes #1345


AudricV:
> When resolving a YT handle like `@TheDailyShow`, YT first returns a response with WEB_PAGE_TYPE_UNKNOWN and a redirect to `thedailyshow`. The redirect must then be resolved again to get the actual channel id with is `UCwWhs_6x42TyRM4Wstoq8HA`.

This is not always the case, this is the first time I see this. You can try for instance with `@google` or `@Gronkh`, you get directly the channel ID in a `browseEndpoint`.

Please update your comment in the code and your PR description :)

litetex:
> Please update your comment in the code and your PR description :)

Done :)

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeChannelHelper.java
```diff
@@ -65,25 +67,44 @@ public static String resolveChannelId(@Nonnull final String idOrPath)
...
+                urlToResolve = "WEB_PAGE_TYPE_UNKNOWN".equals(webPageType)
+                    ? endpoint.getObject("urlEndpoint").getString("url")
+                    : null;
+            }
+
+            final String browseId = endpoint.getObject(BROWSE_ENDPOINT)
+                .getString(BROWSE_ID, "");
 
             if (webPageType.equalsIgnoreCase("WEB_PAGE_TYPE_BROWSE")
                     || webPageType.equalsIgnoreCase("WEB_PAGE_TYPE_CHANNEL")
```
Stypox: Btw this (|| &&) combination might be wrong, since clarifying parenthesis are missing. What is the intended condition here @AudricV?

Stypox (approved):
Code in this PR looks good to me, and this function is only called to resolve URLs upon fetching stuff so it's not like we're going to get many more requests (it'd be bad if it was one request per info item). Thanks!

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeChannelHelper.java
litetex: the ``equalsIgnoreCase`` should also likely be inverted for better null safety (we can also then remove the ``""`` initialization above)

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeChannelHelper.java
```diff
@@ -65,25 +67,44 @@ public static String resolveChannelId(@Nonnull final String idOrPath)
...
+
+            JsonObject endpoint = new JsonObject();
+            String webPageType = "";
+            // Try to resolve YT channel redirects
+            // It works like that:
+            // @TheDailyShow
+            // -> resolves to thedailyshow
+            // -> resolves to the id: UCwWhs_6x42TyRM4Wstoq8HA
+            // Please note that this is not always the case, some handles
+            // e.g. @google or @Gronkh directly resolve the id
```
AudricV, line 80: Not really important, but I wonder whether the contrary is more common instead. If so, this comment should updated.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeChannelHelper.java
AudricV: > What is the intended condition here

The intended condition is if it is a valid webpage type and a valid browse ID so:

```java
(webPageType.equalsIgnoreCase("WEB_PAGE_TYPE_BROWSE")
        || webPageType.equalsIgnoreCase("WEB_PAGE_TYPE_CHANNEL"))
        && !browseId.isEmpty()
```

Thanks for catching this [4 year-old-bug I introduced](https://github.com/TeamNewPipe/NewPipeExtractor/commit/a12c69da7def3049271f6944255e5e30a5b7c7b7#diff-43e2b02f4974550a9a32348de7b19a0d6efd3cba5f13c0ec11f4cb06a01f503aR119-R124)!

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YouTubeChannelHelperTest.java
AudricV, line 1: We should probably add tests for channels without this handle -> c/ -> browseId redirection.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeChannelHelper.java
TobiGr: Fixed in 6397b2ec3f284a2975a732fc0b45881d1e6c656e. It wasn't the only place where the parenthesis was missing. `getChannelResponse` and  `resolveChannelId` are quite similar and might qualify for deduplication at a later point...

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YouTubeChannelHelperTest.java
TobiGr: added in aa70b5a04bb8f1c3e64b5ce450d1b34423a69fb5. I amended the initial commit which creates the `YouTubeChannelHelperTest` because the other mocks were touched while recording the new ones for the class.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeChannelHelper.java
Stypox: I wonder if there's any checkstyle rule for this, but I could only find one which does the opposite :sweat_smile: https://checkstyle.org/checks/coding/unnecessaryparentheses.html

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YouTubeChannelHelperTest.java
```diff
@@ -0,0 +1,45 @@
...
+    void resolveSuccessfulGronkh(final String idOrPath) {
+        final String id = assertDoesNotThrow(
+                () -> YoutubeChannelHelper.resolveChannelId(idOrPath));
+        assertEquals("UCYJ61XIK64sp6ZFFS8sctxw", id);
+    }
+
+    @Test
+    void resolveFailNonExistingTag() {
+        assertThrows(ExtractionException.class, () -> YoutubeChannelHelper.resolveChannelId(
+            "@nonExistingHandleThatWillNeverExist15464"));
```
Stypox, line 43: Mmmh, let me create a channel with that name xD

PR #1352, opened 2025-07-28
 [YouTube] Add custom error for "Sign in to confirm ..."

Stypox:

This PR adds a custom exception when "Sign in to confirm that you're not a bot" is detected from YouTube, so we can show a custom message in NewPipe. I could not test whether the detection works because I am not experiencing temporary bans from YouTube at the moment, but I based my checks on https://github.com/TeamNewPipe/NewPipe/issues/11139 .

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
@@ -53,6 +53,7 @@ import org.schabi.newpipe.extractor.exceptions.PaidContentException;
 import org.schabi.newpipe.extractor.exceptions.ParsingException;
 import org.schabi.newpipe.extractor.exceptions.PrivateContentException;
 import org.schabi.newpipe.extractor.exceptions.YoutubeMusicPremiumContentException;
+import org.schabi.newpipe.extractor.exceptions.SignInConfirmNotBotException;
 import org.schabi.newpipe.extractor.linkhandler.LinkHandler;
 import org.schabi.newpipe.extractor.localization.ContentCountry;
 import org.schabi.newpipe.extractor.localization.DateWrapper;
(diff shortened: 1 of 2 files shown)


Stypox:
Many users reported that the detection works fine, e.g. see https://github.com/TeamNewPipe/NewPipe/issues/12533

Stypox:
I removed detection for "Watch on the latest version" because that's an extractor bug (i.e. the request is invalid, just like other parsing exceptions) and the user can't do anything about it

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -901,7 +902,14 @@ private static void checkPlayabilityStatus(@Nonnull final JsonObject playability
...
 
-        throw new ContentNotAvailableException("Got error: \"" + reason + "\"");
+        // "Sign in to confirm that you're not a bot"
+        if (reason != null && reason.contains("a bot")) {
+            throw new SignInConfirmNotBotException(
+                    "YouTube probably temporarily blocked this IP, got error "
+                            + status + ": \"" + reason + "\"");
+        }
+
+        throw new ContentNotAvailableException("Got error " + status + ": \"" + reason + "\"");
```
Stypox, line 912: I was thinking, should this really be a `ContentNotAvailableException`? I think `ContentNotAvailableException` should be reserved for when the extractor notices that the service explicitly reports a resource as not being available (e.g. age restricted, channel deleted, and so on). It should not be used for unknown errors in fetching resources, which is the case here. I would turn this into just `ParsingException`, add a clarifying comment to `ContentNotAvailableException`, and check all other places where `ContentNotAvailableException` is thrown. What's your opinion @AudricV?

Note: currently in some parts of NewPipe we assume that `ContentNotAvailableException`s are not supposed to be reported.

AudricV: Unless we start detecting all real unavailability errors (deleted video, unavailable/removed due to closed/removed YouTube account, ...), I don't think so.

Stypox: Right now we are unsure what the issue is here: it might be that the video is actually unavailable, but it might also be that YouTube refused to provide data for other reasons. So reporting that the content is unavailable is incorrect, we should instead report that the extractor failed to extract the data. Even if the video was actually unavailable and the extractor was unable to notice that, then it's the extractor's fault so ParsingException fits.

Stypox: This is what I was going after. But for this classification to really work well, we need the two exception types to be distinct.

Image

TobiGr: sounds like a good idea.

PR #1354, opened 2025-07-31
[YouTube] Add support for more kiosks and change default one

AudricV:

This pull request adds support for new kiosks and is the extractor implementation for TeamNewPipe/NewPipe#1805:

- [trending podcasts episodes](https://www.youtube.com/podcasts/popularepisodes);
- [trending gaming videos](https://www.youtube.com/gaming/trending);
- [trending movies and shows trailers](https://charts.youtube.com/charts/TrendingTrailers);
- [trending music clips, unofficial and autogenerated content](https://charts.youtube.com/charts/TrendingVideos/RightNow);
- [running livestreams](https://www.youtube.com/channel/UC4R8DWoMoI7CAwX8_LjQHig/livetab?bp=EgdsaXZldGFikgEDCKEK).

Note that trends extracted from YouTube Charts (trending movie trailers and trending music clips) are not available in every supported YouTube country: for movie trailers, unsupported countries return a small amount of videos; for trending music, a 400 HTTP error.

For this latest trend, if the requested country is not in the hardcoded list of supported YouTube Charts countries (parsing it dynamically would require an additional request to a YouTube Charts HTML page), a new exception, `UnsupportedContentInCountryException`, is thrown (no test has been added for this).

This PR set as default kiosk running livestreams, per the current result of TeamNewPipe/NewPipe#12445. The legacy trending kiosk since July 21, 2025 has been kept, as it still works for now (even if the trending URL now redirects to YouTube's homepage).

Invalid usage of the YouTube Music client version for the YouTube web embed client has also been fixed; some additional changes have been done and are detailed in commits.


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/kiosk/YoutubeChartsBaseKioskExtractor.java
```diff
@@ -0,0 +1,239 @@
...
+        }
+
+        @Nonnull
+        @Override
+        public DateWrapper getUploadDate() {
+            final JsonObject releaseDate = videoObject.getObject("releaseDate");
+            return new DateWrapper(OffsetDateTime.of(
+                    releaseDate.getInt("year"),
+                    releaseDate.getInt("month"),
+                    releaseDate.getInt("day"),
```
Stypox, line 210: Oh, now I see why you mean that there is no textual upload date in this case

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/InnertubeClientRequestInfo.java
```diff
@@ -106,17 +109,27 @@ public static InnertubeClientRequestInfo ofWebClient() {
     public static InnertubeClientRequestInfo ofWebEmbeddedPlayerClient() {
         return new InnertubeClientRequestInfo(
                 new InnertubeClientRequestInfo.ClientInfo(WEB_EMBEDDED_CLIENT_NAME,
-                        WEB_REMIX_HARDCODED_CLIENT_VERSION, EMBED_CLIENT_SCREEN,
-                        WEB_EMBEDDED_CLIENT_ID, null),
+                        WEB_EMBEDDED_CLIENT_VERSION, WEB_EMBEDDED_CLIENT_ID, EMBED_CLIENT_SCREEN,
```
Stypox, line 112: This is the one you fixed right? I tested an age restricted video and it was properly reported as such so we should be fine I guess

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeKioskExtractorTest.java
```diff
@@ -3,22 +3,305 @@
...
+            assertEquals(
+                    "https://www.youtube.com/channel/UC4R8DWoMoI7CAwX8_LjQHig/livetab?ss=CKEK",
+                    extractor().getOriginalUrl());
+        }
+
+        @Override
+        @Test
+        public void testRelatedItems() throws Exception {
+            // As there is sometimes very recently ended livestreams present, we can't test whether
+            // all streams are running live streams
```
Stypox, line 67: We could test that there is at least 10% of livestreams? But yeah this is just a possible improvement for later

Stypox (approved):
Thank you! I re-reviewed again and things look good, and tested out from within NewPipe and kiosks seem to work well.

PR #1374, opened 2025-09-13
Remove LocaleCompat class

Isira-Seneviratne:
----
Remove the `LocaleCompat` class and use the native `Locale.forLanguageTag` method instead.

Since most mobile apps now have a `minSdk` of 21, this change should not have much impact.


Stypox:
The other day I noticed that SkyTube uses NewPipeExtractor and they have [a few commits](https://github.com/TeamNewPipe/NewPipeExtractor/compare/dev...SkyTubeTeam:NewPipeExtractor:dev) on top of our `dev` specifically for retaining compatibility with API 19+. In NewPipe we have 21+, yes, and SkyTube can easily adapt to this change by adding another commit on top, yes. But I don't see the point of making a refactor like this (which also has the downside of introducing more breaking changes for older Android versions) unless it significantly simplifies some part of the code. It was not so bad to have our own `forLanguageTag` implementation: it's quite short, its source is well documented, and especially, it has worked just fine for 3 years. I would prefer not to merge this PR. I'm open for discussion though, sorry for my rough take on this :-)

PR #1384, opened 2025-10-10
[YouTube] Refactor some extraction code using Optional

Isira-Seneviratne:
----
Refactor some of the helper methods in YoutubeParsingHelper to return `Optional`, which allows the number of lines of code to be significantly reduced.


TobiGr:
I am not sure if this is going to be needed anymore. The long-term plan is to port the extractor to kotlin.

Isira-Seneviratne:
> I am not sure if this is going to be needed anymore. The long-term plan is to port the extractor to kotlin.

When doing the migration, it could be useful as a guide, e.g. `Optional#or()` can be translated to a `?:` block in Kotlin.

PR #1388, opened 2025-10-13
on.soundcloud fixes

mjsir911:

TeamNewPipe/Newpipe#12694

also related: 02e14b8931c903062733ab6bbc89a2040145a4da

The name change commit isn't necessary but I would appreciate it for git history view consistency


mjsir911:
I'm unsure how to run this with RecordingDownloader like the failing tests say -- they successfully run on my machine

TobiGr:
The tests are failing because you changed tests which require requests to external servers. To minimize the load on external servers and to make tests more reliable we are using so called "mocks" to simulate the requests. This is done by "recording" network requests and using those records when running the tests. See https://teamnewpipe.github.io/documentation/05_Mock_tests/ for more info.
I'll record the mocks later today for your PR. Don't worry about it.

TobiGr:
thank you!

.mailmap
AudricV, line 1: > The name change commit isn't necessary but I would appreciate it for git history view consistency

Sorry but we don't use this, this needs to be removed.

TobiGr: ok, see #1390

PR #1397, opened 2025-10-31
[YouTube] Extract uploader URL for videos with multiple uploaders

FineFindus:
Fixes an issue, where for videos with multiple uploaders/collaborators, the uploader URL could not be extracted, leading to the entire video failing to be extracted. This is fixed by extracting the URL of the first collaborator, which seems to be the main channel where the video is actually displayed.

However, this PR does _not_ implement support for extracting all the uploaders.

Closes: https://github.com/TeamNewPipe/NewPipeExtractor/issues/1333
Fixes: https://github.com/TeamNewPipe/NewPipe/issues/12899

- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [ ] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).
- [ ] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API.

Issue #1360, opened 2025-08-09
Refactor link handlers


Stypox:
The interface of link handlers and link handler factories is strange, hard to debug, and it's doing something extremely simple (checking if a URL matches and extracting stuff from it) in a highly overcomplicated way. It could be refactored like this in my opinion:

- have a base interface `Link` (the shorter the name the better, current names are huge)
- The interface has only a few extractor related methods, e.g. `getExtractor()` returns the extractor corresponding to the link.
- The interface also has a few app-facing public methods that are meant to be used only by the app (and not by other parts of the extractor). For example, a `getUniqueId()` method that returns a stable and unique ID for each resource, so it can be used as a primary key in NewPipe's database.
- Each extractor has a corresponding `Link` implementation. Every `Link`'s constructor takes a URL and builds an instance of `Link`, but throws an exception if the URL does not match the expected link format.
- Each `Link` implementor may also have other constructors, e.g. `YoutubeSearchLink` would have a constructor that takes the query and any search filters.

absurdlylongusername:
Should also mention, that we also need to refactor the entire extraction architecture in the general, or least in the case of Soundcloud.

Because for Soundcloud, there are many entry points into extraction that end up extracting data that has already been extracted. E.g. when playing a stream, we extract the stream, but then when getting comments we do another network request in SoundcloudCommentsLinkHandlerFactory to get the url and id for the comments for that stream, even though that data was already in memory at an earlier point.
Idk how it works for other services. 

Regardless, in the general case should refactor the code so that extraction isn't split up like it is: like we can have many entry points to extraction and that can get certain data about a resource (like a certain youtube video), and at any point we need to get related data about that resource, we should use the existing data we already have for that resource to get the rest.

Stypox:
A possible interface to consider (as discussed [on Matrix](https://matrix.to/#/!vnsDQotAJWVaEcyzMl:matrix.newpipe-ev.de/$17s30El37XZdqozLHJhqT8b9lCfG03-IphFBbHnidfI?via=matrix.newpipe-ev.de&via=matrix.org&via=tchncs.de)):

```kt
interface LinkHandler {
    // these return null if the URL (or ID) is not valid,
    // and they can cache stuff in LinkHandler class fields if needed,
    // to avoid doing the same network request twice
    fun resolveAsPlaylist(): ResolvedPlaylistUrl?;
    fun resolveAsChannel(): ResolvedChannelUrl?;
[...]

PR #1399, opened 2025-11-04
[YouTube] Implement support for collaborators

FineFindus:
Implements support for extracting the collaborators of a video. This allows client apps to display them, if they wish so. The first collaborator is equivalent to the creator/uploader of the video.

Ref: https://github.com/TeamNewPipe/NewPipeExtractor/pull/1397

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/Creator.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/Creator.java
@@ -1,0 +1,110 @@
+package org.schabi.newpipe.extractor;
+
+import java.util.List;
+import javax.annotation.Nonnull;
+
+/**
+ * Class representing a creator of a piece of media that has been extracted.
+ */
+public final class Creator {
+
+    /**
+     * Constant representing that the amount of subscribers of a {@link Creator} is unknown.
+     */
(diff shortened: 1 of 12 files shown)


FineFindus:
There's currently naturally a lot of duplicate code between extracting all the collaborators details, and extracting just a single detail (e.g. the URL). Maybe it would be better to extract them into separate functions? Or maybe it would be even better deprecated the singular uploader methods and always return at least one creator? That may have a small performance cost, though in our use-case we are always accessing all of them.

absurdlylongusername:
> There's currently naturally a lot of duplicate code between extracting all the collaborators details, and extracting just a single detail (e.g. the URL). Maybe it would be better to extract them into separate functions? Or maybe it would be even better deprecated the singular uploader methods and always return at least one creator? That may have a small performance cost, though in our use-case we are always accessing all of them.

https://github.com/TeamNewPipe/NewPipeExtractor/issues/1360#issuecomment-3170556986 perhaps related

TobiGr:
@Stypox @AudricV I'd be okay with deprecating the singular uploader methods. We should also think about using the `Creator` class for all the other attributes as well and could deprecate those, too. What do you think?

PR #1409, opened 2025-11-21
Refactor build and publishing setup

theimpulson:
#### Summary

This PR:
- Migrates build DSL to Kotlin as it has better syntax, highlights and lint support compared to groovy
- Switches to build version catalog to manage dependencies version across all modules in one place
- Relocates all publishing and test steps to extractor module
- Adds missing information for publishing extractor on mavencentral

#### Due diligence

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -3,9 +3,6 @@
  * SPDX-License-Identifier: GPL-3.0-or-later
  */
 
-import org.gradle.api.tasks.testing.logging.TestExceptionFormat
-import org.gradle.api.tasks.testing.logging.TestLogEvent
-
 plugins {
     alias(libs.plugins.google.protobuf) apply false
 }
@@ -36,15 +33,6 @@ subprojects {
                 "implSpec:a:Implementation Requirements:",
                 "implNote:a:Implementation Note:"
             )
-        }
-    }
-
-    // Test logging setup
-    tasks.withType<Test>().configureEach {
-        testLogging {
-            events = setOf(TestLogEvent.SKIPPED, TestLogEvent.FAILED)
-            showStandardStreams = true
-            exceptionFormat = TestExceptionFormat.FULL
         }
     }
 }
--- a/extractor/build.gradle.kts
+++ b/extractor/build.gradle.kts
@@ -3,6 +3,9 @@
  * SPDX-License-Identifier: GPL-3.0-or-later
  */
 
+import org.gradle.api.tasks.testing.logging.TestExceptionFormat
+import org.gradle.api.tasks.testing.logging.TestLogEvent
+
 plugins {
     alias(libs.plugins.google.protobuf)
     checkstyle
@@ -22,6 +25,13 @@ tasks.jar {
 }
 
 tasks.test {
+    // Test logging setup
+    testLogging {
+        events = setOf(TestLogEvent.SKIPPED, TestLogEvent.FAILED)
+        showStandardStreams = true
+        exceptionFormat = TestExceptionFormat.FULL
+    }
+
     // Pass on downloader type to tests for different CI jobs. See DownloaderFactory.java and ci.yml
     if (System.getProperties().containsKey("downloader")) {
         systemProperty("downloader", System.getProperty("downloader"))


build.gradle.kts
```diff
@@ -0,0 +1,94 @@
...
+    apply(plugin = "java-library")
+    apply(plugin = "maven-publish")
+
+    tasks.withType<JavaCompile> {
+        options.encoding = Charsets.UTF_8.toString()
+    }
+
+    extensions.configure<JavaPluginExtension> {
+        toolchain {
+            languageVersion.set(JavaLanguageVersion.of(17))
```
TobiGr: Why Java 17 and not 11?

We used this before:
``` build.gradle
    sourceCompatibility = JavaVersion.VERSION_11
    targetCompatibility = JavaVersion.VERSION_11
```
The CI also works with Java 11.

theimpulson: Fixed, thanks for noticing

theimpulson:
Signing support is still missing as that's needed for mavencentral

theimpulson:
I couldn't find classes from the `timeago-*` modules in the currently generated jar file for the extractor nor do they seem to be getting generated separately or consumed in the NewPipe app. This made me assume that they are supposed to be consumed by the extractor and aren't supposed to be released.

Hence, I moved the javadoc, source and publishing setup to the extractor module.

extractor/build.gradle.kts
```diff
@@ -0,0 +1,143 @@
...
+            artifactId = "extractor"
+            version = "v0.24.8"
+
+            afterEvaluate {
+                from(components["java"])
+            }
+
+            pom {
+                name = "NewPipe Extractor"
+                description = "Library for extracting things from streaming sites"
```
Stypox:
```suggestion
                description = "A library for extracting data from streaming websites, used in NewPipe"
```
Or we might want to think of something better altogether

build.gradle.kts
```diff
@@ -0,0 +1,37 @@
...
+            encoding = Charsets.UTF_8.toString()
+            links = listOf("https://docs.oracle.com/javase/11/docs/api/")
+            tags = listOf(
+                "apiNote:a:API Note:",
+                "implSpec:a:Implementation Requirements:",
+                "implNote:a:Implementation Note:"
+            )
+        }
+    }
+}
```
Stypox, line 38: Please keep the documentation lines that were there in `build.gradle`

extractor/build.gradle.kts
```diff
@@ -0,0 +1,143 @@
...
+    checkstyle
+    `maven-publish`
+}
+
+java {
+    withSourcesJar()
+    withJavadocJar()
+}
+
+// Prevent .proto files ending up in JARs
```
Stypox: Keep the same comment that we had before

extractor/build.gradle.kts
```diff
@@ -0,0 +1,143 @@
...
+
+tasks.test {
+    // Test logging setup
+    testLogging {
+        events = setOf(TestLogEvent.SKIPPED, TestLogEvent.FAILED)
+        showStandardStreams = true
+        exceptionFormat = TestExceptionFormat.FULL
+    }
+
+    // Pass on downloader type to tests for different CI jobs.
```
Stypox: Also here

extractor/build.gradle.kts
```diff
@@ -0,0 +1,143 @@
...
+    includeEmptyDirs = false
+}
+
+tasks.test {
+    // Test logging setup
+    testLogging {
+        events = setOf(TestLogEvent.SKIPPED, TestLogEvent.FAILED)
+        showStandardStreams = true
+        exceptionFormat = TestExceptionFormat.FULL
+    }
```
Stypox, line 33: Why is this only in the extractor `build.gradle`? I think it was fine to have it for all subprojects (although yeah we don't have tests elsewhere)

Stypox (changes requested):
Thanks! I compared all of the files and they mostly look good to me. I also tried building locally and I was able to run tests, and also to build NewPipe when depending on the local copy of NewPipeExtractor.

> I couldn't find classes from the timeago-* modules in the currently generated jar file for the extractor nor do they seem to be getting generated separately or consumed in the NewPipe app. This made me assume that they are supposed to be consumed by the extractor and aren't supposed to be released.

You mean that the `timeago-parser` JAR is not needed externally? Yeah that's correct, `timeago-parser` is just a subproject used internally in NewPipeExtractor that shouldn't be published separately.

> Signing support is still missing as that's needed for mavencentral

I think it's fine to proceed with this using environment variables. Then we can setup a CI action that creates a new release on mavencentral. However this should be done in a separate PR

extractor/build.gradle.kts
theimpulson: It is the same comment as before though. This block was in the top-level gradle file and I moved it here as only this module is dealing with it.

extractor/build.gradle.kts
theimpulson: Yeah, I moved it here as other modules don't have tests and this already had other test configuration. I am okay with moving everything to the top-level file but doesn't makes much sense to me unless we are planning to add tests for them too.

extractor/build.gradle.kts
theimpulson: This was already here in this module. The comment is the same as well.

build.gradle.kts
theimpulson: Do you mean to generate aggregate API docs? Do we intend to publish API docs for other modules too even though they aren't part of the public API present in the JAR?

extractor/build.gradle.kts
Stypox: The comment was this before:
```
    // Protobuf files would uselessly end up in the JAR otherwise, see
    // https://github.com/google/protobuf-gradle-plugin/issues/390
```

extractor/build.gradle.kts
Stypox: Yeah let's keep it here

extractor/build.gradle.kts
Stypox: No the comment was
```
    // Pass on downloader type to tests for different CI jobs. See DownloaderFactory.java and ci.yml
```

build.gradle.kts
Stypox: No, I mean these:
```
// https://discuss.gradle.org/t/best-approach-gradle-multi-module-project-generate-just-one-global-javadoc/18657/21
    // Fixes unknown tag @implNote; the other two were added precautionary
```

theimpulson:
> Signing support is still missing as that's needed for mavencentral

I will do this in a separate PR

TobiGr:
@theimpulson This removed the `aggregatedJavadocs` task. We are currently only getting a jdocs for each module. But we need to have a combined JDoc containing the docs for all modules.

TobiGr:
I re-added the task to generate aggregatedJavadocs in 0695a0982bf31e5f554419f5531b3af505992b03. We still need to decide whether to keept the reconfiguration of the single javadoc tasks (which is ok for me). Currently there is quite some duplicated code in the file due to this.

theimpulson:
> @theimpulson This removed the `aggregatedJavadocs` task. We are currently only getting a jdocs for each module. But we need to have a combined JDoc containing the docs for all modules.

@TobiGr as we only aim to publish the extractor module, why are we publishing doc for other modules?

PR #1412, opened 2025-11-30
[YouTube] extract additional stream information

FineFindus:
Implements support for extracting `lastModified`, `xtags` and `isDrc` (whether a format is using dynamic range compression) from formats. They are required for initiating SABR playback, but may also be useful on their own (e.g. to filter audio streams with DRC).

Ref: https://github.com/TeamNewPipe/NewPipe/issues/12248

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/ItagItem.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/ItagItem.java
@@ -269,6 +269,7 @@ public class ItagItem implements Serializable {
     private Locale audioLocale;
     private boolean isDrc;
     private long lastModified;
+    private String xtags;
 
     public int getBitrate() {
         return bitrate;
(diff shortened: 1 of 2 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/ItagItem.java
```diff
@@ -647,4 +650,72 @@ public Locale getAudioLocale() {
...
+
+
+    /**
+     * When the stream was last modified.
+     *
+     * @return unix timestamp of when the stream was last modified
+     */
+    public long getLastModified() {
+        return lastModified;
+    }
```
TobiGr, line 694: Please add a LAST_MODIFIED_UNKNOWN constant and document its use here

FineFindus: Done, though I don't think that is necessary, since every format always has a `lastModified` field.

TobiGr: ah, ok. I didn't know that, I just saw that there was a fallback value and thought that having documentation for it might be useful.

PR #1417, opened 2025-12-15
[YouTube] Extract shorts thumbnails from `thumbnailViewModel`

FineFindus:
Fixes an issue, where shorts thumbnails where not extracted, as they're now contained within a `thumbnailViewModel`, instead of a simple `thumbnail` object.

Closes: https://github.com/TeamNewPipe/NewPipe/issues/12907

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeShortsLockupInfoItemExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeShortsLockupInfoItemExtractor.java
@@ -78,7 +78,15 @@ class YoutubeShortsLockupInfoItemExtractor implements StreamInfoItemExtractor {
     @Nonnull
     @Override
     public List<Image> getThumbnails() throws ParsingException {
-        return getImagesFromThumbnailsArray(shortsLockupViewModel.getObject("thumbnail")
+        if (shortsLockupViewModel.has("thumbnail")) {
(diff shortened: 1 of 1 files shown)


TobiGr:
Thank you!

I still get valid thumbails sometimes. Looks like this change is curerntly A/B tested. Could you add a check that extracts depending on whether there is a `thumbnailViewModel` or just a `thumbnail` object?

FineFindus:
Of course. I assumed it was already fully rolled out, otherwise I would've kept the original code.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeShortsLockupInfoItemExtractor.java
```diff
@@ -78,7 +78,14 @@ public String getUrl() throws ParsingException {
     @Nonnull
     @Override
     public List<Image> getThumbnails() throws ParsingException {
-        return getImagesFromThumbnailsArray(shortsLockupViewModel.getObject("thumbnail")
+        if (shortsLockupViewModel.has("thumbnail")) {
+            return getImagesFromThumbnailsArray(shortsLockupViewModel.getObject("thumbnail");
```
TobiGr:
```suggestion
            return getImagesFromThumbnailsArray(shortsLockupViewModel.getObject("thumbnail"));
```

TobiGr (approved):
thanks! LGTM

PR #1422, opened 2025-12-24
Update Gradle and dependencies

theimpulson:

### DIFF
--- a/extractor/build.gradle.kts
+++ b/extractor/build.gradle.kts
@@ -40,6 +40,13 @@ tasks.test {
     dependsOn(tasks.checkstyleMain) // run checkstyle when testing
 }
 
+// https://checkstyle.org/#JRE_and_JDK
+tasks.withType<Checkstyle>().configureEach {
+    javaLauncher = javaToolchains.launcherFor {
+        languageVersion = JavaLanguageVersion.of(21)
+    }
+}
+
 checkstyle {
     configDirectory = rootProject.file("checkstyle")
     isIgnoreFailures = false
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -4,7 +4,7 @@
 #
 
 [versions]
-checkstyle = "10.26.1"
+checkstyle = "12.3.0"
 gson = "2.13.2"
 jsr305 = "3.0.2"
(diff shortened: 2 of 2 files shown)


theimpulson:
We cannot update checkstyle as latest version seems to require JDK 17 and above while we are enforcing JDK 11 bytecode.

theimpulson:
> We cannot update checkstyle as latest version seems to require JDK 17 and above while we are enforcing JDK 11 bytecode.

I figured this out with https://docs.gradle.org/current/userguide/toolchains.html#toolchains_for_tasks

TobiGr (changes requested):
We should apply the use of the newer JDK to all files:
https://github.com/TeamNewPipe/NewPipeExtractor/blob/8972df0a062fc98b623d8e286d83b8aea4343747/build.gradle.kts#L19
https://github.com/TeamNewPipe/NewPipeExtractor/blob/8972df0a062fc98b623d8e286d83b8aea4343747/jitpack.yml

I think the policy was to keep the compile JDK low and to not use new language features to support the lowest LTS version, which is now 17 as 11 went EOL this year. Should we stick to that? @Stypox @FireMasterK

theimpulson:
> We should apply the use of the newer JDK to all files:
> 
> https://github.com/TeamNewPipe/NewPipeExtractor/blob/8972df0a062fc98b623d8e286d83b8aea4343747/build.gradle.kts#L19
> 
> https://github.com/TeamNewPipe/NewPipeExtractor/blob/8972df0a062fc98b623d8e286d83b8aea4343747/jitpack.yml
> 
> I think the policy was to keep the compile JDK low and to not use new language features to support the lowest LTS version, which is now 17 as 11 went EOL this year. Should we stick to that? @Stypox @FireMasterK

Can we go ahead with this PR for now and do this discussion separately to unblock it as bumping JDK version isn't the aim for this PR?

Stypox (approved):

Stypox:
> I think the policy was to keep the compile JDK low and to not use new language features to support the lowest LTS version, which is now 17 as 11 went EOL this year. Should we stick to that?

Yeah I guess it makes sense to switch to Java 17 now

PR #1440, opened 2026-01-26
Use java 21 toolchain on whole extractor and set java via release

ghost:
Toolchain specifies which exact jdk version used for compiling gradle projects. Using several different toolchains in one repo without any reason is wastedul. Unifying used jdk will allow to store only one jdk on dev side. As a side effect unification should make problem encountered by some users on Matrix (it was tried to solve it in #12923)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -14,11 +14,12 @@ allprojects {
 
     tasks.withType<JavaCompile> {
         options.encoding = Charsets.UTF_8.toString()
+		options.release = 11
     }
 
     extensions.configure<JavaPluginExtension> {
         toolchain {
-            languageVersion.set(JavaLanguageVersion.of(11))
+            languageVersion.set(JavaLanguageVersion.of(21))
         }
     }
 }


ghost:
Resources for reading:

Big and a bit messy https://docs.gradle.org/current/userguide/toolchains.html
TLDR version: https://committing-crimes.com/articles/2024-10-30-using-gradle-toolchains-properly/

ghost:
Both extractor and newpipe probably should use the same toolchain version and just set different language versions..(?)

ghost:
CC @theimpulson @AudricV

FireMasterK:
Does android and its desugaring support Java 21? I am interested in this as well, but that's my main concern as to breakages over there.

ghost:
> Does android and its desugaring support Java 21? I am interested in this as well, but that's my main concern as to breakages over there.

The PR is about just using jdk 21 everywhere but it doesn't mean you will be able to use 21 features. You will still use java 11 but just will need only one jdk installed not two for everything to work properly

PR #1442, opened 2026-01-26
[YouTube] Support more YouTube channel URLs

### LINKED ISSUE #965: Some Channel URLs are assumed to be a prefix of `/c/`
Example: https://www.youtube.com/DreamTraps

This does not open in NewPipe.

G-flat:

I had a look at what kinds of YouTube channel URLs there are. They come in 5 forms:

1. A name (`https://www.youtube.com/jacksepticeye`)
2. `"c/" + name` (`https://www.youtube.com/c/jacksepticeye`)
3. `"user/" + name` (`https://www.youtube.com/user/jacksepticeye`)
4. A YouTube handle (`https://www.youtube.com/@jacksepticeye`)
5. A channel ID (`https://www.youtube.com/channel/UCYzPXprvl5Y-Sf0g4vX-m6g`)

In the current `getId()` implementation, for non-handle-based URLs, `c/` is prepended to the URL path if it had only one component (i.e. looks like a name) and that component isn't one of the excluded segments. Then, if the URL path doesn't start with `c/`, `user/` or `channel/`, it's rejected; otherwise, `the first path component + "/" + the second path component` is returned.

While this implementation works for the 5 URLs given above, it doesn't work when `"c/" + name` does not lead to a valid channel URL. I've found that channel URLs are a little complicated:

- `"user/" + name` works for names like `PewDiePie`, `DreamTraps`, `TEDtalksDirector` and `markiplierGAME`.
- `"user/" + name` fails and `"c/" + name` works for `ElectroBOOM` and `노마드코더NomadCoders`.
- `"c/" + name` and `"user/" + name` create technically valid URLs for `LofiGirl` and `veritasium`, but they lead to completely different channels.
- `"@" + name` works for some of the names I listed (e.g. `PewDiePie`), but not for all of them (e.g. `DreamTraps`).
- A channel can have more than one name (`dream` and `DreamTraps`; `markiplier` and `markiplierGAME`; `TED` and `TEDtalksDirector`), with `"c/" + name"`, `"user/" + name` and `"@" + name"` working or failing (e.g. `"c/TED"` and `"@TED"` work but `"user/TED"` doesn't; `"user/TEDtalksDirector"` works but `"c/TEDtalksDirector"` and `"@TEDtalksDirector"` don't).

Since `"c/" + name` and prepending other components to the name doesn't always give a valid channel URL, this pull request modifies `getId()` to consider URLs containing just a name to be valid. This makes the `https://www.youtube.com/DreamTraps` URL given in #965 work, as well as other URLs like `https://www.youtube.com/PewDiePie`.

Though this pull request doesn't add support for every valid YouTube channel URL (e.g. `https://www.youtube.com/jacksepticeye` works but `https://www.youtube.com/jacksepticeye/videos` doesn't; an edge case like `https://www.youtube.com/select_site/videos` doesn't work), it fixes #965. On a side note, it looks like TeamNewPipe/NewPipe#9436 was already fixed before this pull request was made.

### DIFF
--- a/extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeChannelLinkHandlerFactoryTest.java
+++ b/extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeChannelLinkHandlerFactoryTest.java
@@ -53,12 +53,18 @@ public class YoutubeChannelLinkHandlerFactoryTest {
 
         // do not accept URLs which are not channels
         assertFalse(linkHandler.acceptUrl("https://www.youtube.com/watch?v=jZViOEv90dI&t=100"));
+        assertFalse(linkHandler.acceptUrl("https://www.youtube.com/watch"));
         assertFalse(linkHandler.acceptUrl("http://www.youtube.com/watch_popup?v=uEJuoEs1UxY"));
+        assertFalse(linkHandler.acceptUrl("http://www.youtube.com/watch_popup"));
         assertFalse(linkHandler.acceptUrl("http://www.youtube.com/attribution_link?a=JdfC0C9V6ZI&u=%2Fwatch%3Fv%3DEhxJLojIE_o%26feature%3Dshare"));
+        assertFalse(linkHandler.acceptUrl("http://www.youtube.com/attribution_link"));
         assertFalse(linkHandler.acceptUrl("https://www.youtube.com/playlist?list=PLW5y1tjAOzI3orQNF1yGGVL5x-pR2K1d"));
+        assertFalse(linkHandler.acceptUrl("https://www.youtube.com/playlist"));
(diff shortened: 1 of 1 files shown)


TobiGr:
Thank you, please fix the checkstyle issue and add tests for the newly supported channel URLs in YoutubeChannelLinkHandlerFactoryTest

G-flat:
Thank you for going through my pull request!

I fixed the checkstyle issue and added two URLs into YoutubeChannelLinkHandlerFactoryTest that were previously unsupported (`https://www.youtube.com/PewDiePie` and `https://www.youtube.com/DreamTraps`).

From running the tests, I noticed that `linkHandler.acceptUrl("https://www.youtube.com/?app=desktop&persist_app=1")` doesn't return `false` as expected because `splitPath` ended up being just `[""]`, so I fixed the regression by adding a zero-length check in `isCustomShortChannelUrl()`.

TobiGr (approved):
thanks!

PR #1449, opened 2026-01-30
Use EnumSet instead of List for MediaCapabilities inside Service

ghost:

## 🔴 Breaking API change

`Service.getMediaCapabilities()` now returns `EnumSet<MediaCapability>` instead of `List<MediaCapability>`

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/StreamingService.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/StreamingService.java
@@ -26,7 +26,6 @@ import org.schabi.newpipe.extractor.utils.Utils;
 
 import javax.annotation.Nullable;
 import java.util.Collections;
-import java.util.EnumSet;
 import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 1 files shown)


ghost:
Performance win is probably very low but it is good for semantic...

But List implementation *is unmodifiable* while EnumSet can't be and List conveys the semantic very well so this PR, well, the most useless/arguable my PR... Sorry...

TobiGr:
please fix the linting errors

ghost:
I thought if I directly import it in newpipe, it will run checkstyle of extractor automatically 😅

ghost:
@TobiGr Should I replace it with Set?

TobiGr:
Yes, please use Set in the declarations and EnumSet as implementation

TobiGr:
Sorry, i meant instantiation and not implementation.

extractor/src/main/java/org/schabi/newpipe/extractor/StreamingService.java
```diff
@@ -54,23 +56,23 @@ public abstract class StreamingService {
     public static class ServiceInfo {
         private final String name;
 
-        private final List<MediaCapability> mediaCapabilities;
+        private final EnumSet<MediaCapability> mediaCapabilities;
```
ghost: I am confused now. @TobiGr do you mean use Set here? And in constructor?

ghost: Sorry this message wasn't send earlier. Thanks to githubs "pending" feature being not really well spottable...

PR #1450, opened 2026-01-30
[SoundCloud] Use long integers (64-bit) for track IDs to prevent overflows

G-flat:

In TeamNewPipe/NewPipe#12584, loading the https://soundcloud.com/roman-bene-939339090/dreams track will lead to an error that says `no track with id -2139744901 in response`, which comes from `SoundcloudPlaylistExtractor.getPage()`, as well as an HTTP 404 error from `SoundcloudParsingHelper.getStreamsFromApi()`.

From `SoundcloudStreamExtractor.getId()`, the track's ID is `2155222395`, but in `SoundcloudParsingHelper.getStreamsFromApi()`, the API URL becomes `https://api-v2.soundcloud.com/tracks/-2139744901/related?client_id=<client_id>`. This led to the HTTP 404 error because the track ID became `-2139744901` instead of `2155222395`, which came from using a signed 32-bit track ID that caused `2155222395` (`0x8076157b`) to overflow to `-2139744901` (`0x8076157b - 0x100000000`).

Similarly, in TeamNewPipe/NewPipe#12822, the track "Farming Life in Another World Elf Girls Song Funk" by Bemax - ビマックス has a track ID of `2174506221` which overflowed to `-2120461075`, and in TeamNewPipe/NewPipe#12904, "PLUMMET" by Quarters has a track ID of `2214328829` which overflowed to `-2080638467`.

These issues can be fixed by increasing the bit width for track IDs from 32 bits to 64 bits. This pull request fixes TeamNewPipe/NewPipe#12584, fixes TeamNewPipe/NewPipe#12822 and fixes TeamNewPipe/NewPipe#12904.


extractor/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/extractors/SoundcloudPlaylistExtractor.java
```diff
@@ -149,9 +149,7 @@ public InfoItemsPage<StreamInfoItem> getInitialPage() {
                         streamInfoItemsCollector.commit(
                                 new SoundcloudStreamInfoItemExtractor(track));
                     } else {
-                        // %09d would be enough, but a 0 before the number does not create
-                        // problems, so let's be sure
-                        ids.add(String.format("%010d", track.getInt("id")));
```
TobiGr, line 154: This change causes the mocks to be invalid. Please add a test for a very small ID and also for one that is exceeding the 32 bit of an int. You can fix the tests by recording the mocks for all SoundCloud tests again (see [our docs for more info](https://teamnewpipe.github.io/documentation/05_Mock_tests/)).

G-flat: I updated the SoundcloudPlaylistExtractor mocks and also added two SoundcloudStreamExtractor tests, one for a small track ID (`135535700`) and one for a track ID that's larger than 32 bits of an int (`2167944333`).

PR #1456, opened 2026-02-12
Youtube: throw AccountTerminatedException when account is terminated

ghost:


ghost:
Due to strange circumstances (newpipe already has all infrastructure to show the error message) consider this PR a question "why it is not implemented?". Because there should be a reason why scuh simple thing is still not implemented right?

ghost:
Btw:

Before:
20260212_03h41m46s_grim


After:
20260212_03h30m40s_grim

PR #880, opened 2022-07-28
Use StandardCharsets.UTF_8.

Isira-Seneviratne:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).

PR #1458, opened 2026-02-12
fix: use String charset for URLDecoder/URLEncoder for Android < 13 compatibility

mostafaalagamy:
> * [x]  I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
> * [x]  I have tested the API against [NewPipe](https://github.com/TeamNewPipe/NewPipe).
> * [x]  I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API.
> 

URLDecoder.decode(String, Charset) and URLEncoder.encode(String, Charset) are only available on Android 13+ (API 33). This change uses the older String-based API that works on all Android versions.

Fixes crash: java.lang.NoSuchMethodError: No static method decode( Ljava/lang/String;Ljava/nio/charset/Charset;)Ljava/lang/String;

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
@@ -39,7 +39,7 @@ public final class Utils {
             return URLEncoder.encode(string, UTF_8);
         } catch (final UnsupportedEncodingException e) {
             // UTF-8 is always supported, this should never happen
-            throw new RuntimeException(e);
+            throw new IllegalStateException(e);
         }
     }
 
@@ -53,7 +53,7 @@ public final class Utils {
             return URLDecoder.decode(url, UTF_8);
         } catch (final UnsupportedEncodingException e) {
             // UTF-8 is always supported, this should never happen
-            throw new RuntimeException(e);
+            throw new IllegalStateException(e);
         }
     }
 


TobiGr:
Did this crash happen in NewPipe or did it happen while using another app? If it occured in another app, was core library desugaring enabled as described in the README?

mostafaalagamy:
> Did this crash happen in NewPipe or did it happen while using another app? If it occured in another app, was core library desugaring enabled as described in the README?

 The crash happened in my app (Metrolist) which uses NewPipeExtractor as a dependency. 

 I have core library desugaring enabled in my app module, but NewPipeExtractor is used in a pure Kotlin/JVM module (not an Android module), so desugaring doesn't apply to it.

That's why I needed to use the older `URLDecoder.decode(String, String)` method instead of `URLDecoder.decode(String, Charset)`."

extractor/src/main/java/org/schabi/newpipe/extractor/utils/Utils.java
```diff
@@ -34,7 +35,12 @@ private Utils() {
      * @return The encoded URL.
      */
     public static String encodeUrlUtf8(final String string) {
-        return URLEncoder.encode(string, StandardCharsets.UTF_8);
+        try {
+            return URLEncoder.encode(string, UTF_8);
+        } catch (final UnsupportedEncodingException e) {
+            // UTF-8 is always supported, this should never happen
+            throw new RuntimeException(e);
```
ghost: Maybe better be using `IllegalStateException`?

mostafaalagamy: Done!

TobiGr:
Why did you replace the StandardCharset use only in one file? It is used in [many more places](https://github.com/search?q=repo%3ATeamNewPipe%2FNewPipeExtractor+StandardCharsets.UTF_8&type=code) and my guess is that your changes are only a partial fix. [We kept using our own `UTF8 `constant for compatibility with Android 19](https://github.com/TeamNewPipe/NewPipeExtractor/pull/880). I think the proper solution is a conversion to kotlin.

@AudricV @Stypox what do you think about this PR? I tend to reject it and advise to fork.

ghost:
> I think the proper solution is a conversion to kotlin.

Idk know much but I've just looked at java docs and charset version is supported since java 10. Kotlin compiles into jvm byte code of specific version. Which means it can use internally just anything supported in that version. Is there any guarantee that kotlin API of url encoding/decoding won't internally just use charset thing from java api? Especially because extractor is not android project it won't be able to respect androids minSdk (tbh I don't even know if it is respected when kotlin plugin is added to android project). I am not sure but it seems to me kotlin is not a solution to this problem but rather one more probably cause of it? Kotlin is certainly a solution when there is a thing which you can't use with current jvm target at all but use kotlin which will internally do all the hacky stuff like in android `*Compat` classes but here is the problem we can use it but dont do it because of "external" dependency unknown to kotlin...

I am not sure if it is possible but kotlin gradle plugin allows to set jvm target for kotlin compiler separately. Maybe it is possible to use jvm target 11 for java and jvm target 8 for kotlin (which would allow to use fancy kotlin) but performance optimizations added in java 11 bytecode won't be available for kotlin code

I am not fully sure about my reasoning here. Also I don't know about desugaring anything except just it allows to use never api in some very hacky way. I would also ask theimpulson but this message is already too big, offtopic and this PR probably is not place for this

AudricV:
You do not use the correct desugaring artifact [in your project](https://github.com/MetrolistGroup/Metrolist/blob/48ae4274a2c0b06679efb0fc8fec96ebfd3a5e83/gradle/libs.versions.toml#L105), please use the `desugar_jdk_libs_nio` one as described in the README. Closing this pull request as invalid.

mostafaalagamy:
> please use the desugar_jdk_libs_nio 

Oh, thanks!

PR #1460, opened 2026-02-22
[YouTube] Fix retrieving the title for videos with no title

G-flat:

It's possible for YouTube videos to have no title, such as https://www.youtube.com/watch?v=nc1kN8ZSfGQ. In `YoutubeStreamInfoItemExtractor.getName()`, the `title` object will contain a `runs` array for videos with a title, but the array will not be present for videos that don't. This pull request changes the method to return an empty title string if `runs` is not found in the `title` object.

Fixes TeamNewPipe/NewPipe#8782.


TobiGr:
Thank you, could you add a test for this video? Those are edgecases that should be tested against.

G-flat:
> Thank you, could you add a test for this video? Those are edgecases that should be tested against.

I've added a test for the YouTube video without a title mentioned in this pull request.

PR #1463, opened 2026-02-28
extractor: Setup signing and publishing for CI snapshots

theimpulson:
#### Summary

This PR:

- Sets up signing and publishing for CI snapshots on Maven Central

#### Guidelines

### DIFF
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -1,0 +1,40 @@
+name: Publish snapshot
+
+on:
+  push:
+    branches:
+      - dev
+  workflow_dispatch:
+
+jobs:
+  build-and-deploy-docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: set up JDK
+        uses: actions/setup-java@v5
+        with:
+          java-version: 21
+          distribution: 'temurin'
+
+      - name: Cache Gradle dependencies
+        uses: actions/cache@v5
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: ${{ runner.os }}-gradle
+
+      - name: Publish snapshot
+        env:
+          PGP_PRIVATE_SIGNING_KEY: ${{ secrets.PGP_PRIVATE_SIGNING_KEY }}
+          PGP_PRIVATE_SIGNING_KEY_PASSWORD: ${{ secrets.PGP_PRIVATE_SIGNING_KEY_PASSWORD }}
+          SONATYPE_MAVEN_CENTRAL_USERNAME: ${{ secrets.SONATYPE_MAVEN_CENTRAL_USERNAME }}
+          SONATYPE_MAVEN_CENTRAL_PASSWORD: ${{ secrets.SONATYPE_MAVEN_CENTRAL_PASSWORD }}
+        run: ./gradlew publishSnapshotPublicationToSonatypeRepository
+
+      - name: Upload metadata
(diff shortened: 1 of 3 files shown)


theimpulson:
I couldn't verify the publishing as the secrets seems to be only available to a protected branch (?)

TobiGr:
~~I setup a minimal branch protection rule
@Stypox set the variables. so he might know more.~~

The workflow is only running on `dev`, so it won't run here.

theimpulson:
> The workflow is only running on dev, so it won't run here.

I need permission to push to the current `snapshot` branch without a PR or by force push to test changes as now its protected.

TobiGr:
I removed the branch protection rule

.github/workflows/snapshot.yml
```diff
@@ -0,0 +1,28 @@
+name: Publish snapshot
+
+on:
+  push:
+    branches:
+      - dev
```
Stypox, line 6: Maybe also allow running it by manual trigger `workflow_dispatch:`

extractor/build.gradle.kts
```diff
@@ -112,47 +118,79 @@ protobuf {
...
-                    url = "https://github.com/TeamNewPipe/NewPipeExtractor"
-                    connection = "scm:git:git@github.com:TeamNewPipe/NewPipeExtractor.git"
-                    developerConnection = "scm:git:git@github.com:TeamNewPipe/NewPipeExtractor.git"
-                }
+            setupPOM()
+        }
+        create<MavenPublication>("snapshot") {
+            groupId = mavenGroupId
+            artifactId = mavenArtifactId
+            version = "${rootProject.version}-SNAPSHOT"
```
Stypox: Won't this create duplicate versions? What about putting the git commit hash in here?

.github/workflows/snapshot.yml
```diff
@@ -0,0 +1,28 @@
...
+
+      - name: Cache Gradle dependencies
+        uses: actions/cache@v5
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: ${{ runner.os }}-gradle
+
+      - name: Publish snapshot
+        run: ./gradlew publishSnapshotPublicationToSonatypeRepository
```
Stypox, line 34: Seeing https://github.com/TeamNewPipe/NewPipeExtractor/actions/runs/22512576822/job/65224535320, I think you need to manually specify all secrets here to make them available as env vars to gradle: https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets#using-secrets-in-a-workflow

Stypox (changes requested):
Thank you!

extractor/build.gradle.kts
theimpulson: I thought about that too, will try it. As per docs, it will keep the versions for 90 days and overwrite the same ones.

Stypox: Yeah I think it'd be better to be able to refer to a specific version when depending on an extractor version

theimpulson: I am not sure how git is configured or working but on my local system the short SHA I get is `66d9ee18f` but CI seems to be getting `66d9ee1` which is 2 character smaller. Maybe should force a specific number of characters for a consistent output.

theimpulson:
Seems to be working fine, so I will go ahead and remove the branch from the workflow file to limit it to dev and manual trigger.

Stypox (approved):
Thank you! This can be merged from my side! Maybe let's wait for @TobiGr to confirm

TobiGr (approved):
I think we can go ahead and test it

PR #1478, opened 2026-04-19
[YouTube] Fix duration and live stream display in related videos

Ecomont:

Two bugs in YoutubeStreamInfoItemLockupExtractor caused related/suggested videos to show wrong duration badges:

1. "00:00" on Shorts and videos without a duration badge
getDuration() threw ParsingException when no thumbnailBottomOverlayViewModel overlay was found. StreamInfoItemsCollector catches that exception and leaves duration at the Java default of 0, which the app renders as "00:00". Fixed by returning -1 (the established contract for unknown duration) instad of throwing also added a null guard before .matches() to prevent NPE on badges with no text field.

2. Live streams showing 00:00 on duration
determineStreamType() only checked thumbnailOverlayBadgeViewModel.thumbnailBadges for the live badge style. In the lockup format, YouTube places the live badge inside thumbnailBottomOverlayViewModel.badges — the same overlay used for duration text. So live streams were returned as VIDEO_STREAM, isLive() was false, and the app never reached the isLiveStream() branch that renders "LIVE". Fixed by adding a second check for thumbnailBottomOverlayViewModel.badges[].thumbnailBadgeViewModel.badgeStyle == THUMBNAIL_OVERLAY_BADGE_STYLE_LIVE

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
@@ -235,6 +235,10 @@ public final class YoutubeParsingHelper {
      */
     public static int parseDurationString(@Nonnull final String input)
             throws ParsingException, NumberFormatException {
+        if (!input.matches(".*\\d.*") && !input.equalsIgnoreCase("SHORTS")) {
+            throw new ParsingException("Error duration string contains no digits: " + input);
+        }
+
         // If time separator : is not detected, try . instead
         final String[] splitInput = input.contains(":")
                 ? input.split(":")
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
@@ -49,6 +49,7 @@ import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.util.List;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 public class YoutubeStreamInfoItemExtractor implements StreamInfoItemExtractor {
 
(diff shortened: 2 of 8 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
```diff
@@ -235,6 +235,10 @@ public static boolean isY2ubeURL(@Nonnull final URL url) {
      */
     public static int parseDurationString(@Nonnull final String input)
             throws ParsingException, NumberFormatException {
+        if (!input.matches(".*\\d.*") && !input.equalsIgnoreCase("SHORTS")) {
+            throw new ParsingException("Error duration string contains no digits: " + input);
+        }
+
```
TobiGr, line 241: I do not understand this change. Why exclude shorts here? If `SHORTS` is passed, it will fail in `convertDurationToInt`.

TobiGr (commented):
Thank you! Just a small question.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
Ecomont: YouTube does something special with shorts in certain places, instead of showing the numeric duration in the corner of the thumbnail, it simply shows the word SHORTS

SHORTS does not actually fail in convertDurationToInt, and that is precisely why the check exists.

Without the !input.equalsIgnoreCase("SHORTS") guard: parseDurationString("SHORTS") would hit this line first
```
if (!input.matches(".*\\d.*")) {
    throw new ParsingException("Error duration string contains no digits: " + input);
}
```

It would throw immediately, never reaching convertDurationToInt at all.

The check is not there to exclude shorts, but to allow them thrugh so that the downstream fallback (returning 0) can handle the case where YouTube sends SHORTS instead of an actual timestamp

TobiGr (approved):

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeParsingHelper.java
TobiGr: Ah yeah, I see. Thank you.

PR #1481, opened 2026-04-24
Feature/live chat comments

Ecomont:
This PR adds support for extracting YouTube live chat messages and exposing them through the existing comments API. When a live stream has regular comments disabled, the extractor now falls back to fetching live chat messages instead.

### Changes
- **`CommentsExtractor` / `CommentsInfo`**: Added `isLiveChat()` flag to distinguish live chat from regular comments.
- **`YoutubeCommentsExtractor`**: Added `findLiveChatContinuation()` and `fetchLiveChat()` to retrieve live chat via the `live_chat/get_live_chat` endpoint.
- **`YoutubeLiveChatInfoItemExtractor`**: New extractor that maps `liveChatTextMessageRenderer` JSON to `CommentsInfoItem`, with proper emoji run handling.
- **Page routing**: Live chat continuations are marked with a `live_chat` page identifier so `getPage()` routes correctly on subsequent extractor instances.

### Implementation notes
The approach is based on the live chat implementation in [PipePipe](https://github.com/InfinityLoop1308/PipePipe), adapted to fit the NewPipe Extractor architecture. Key adaptations include:
- Reusing the existing `CommentsInfoItem` type instead of introducing a separate live chat item class.
- Proper `isLiveStream` flag handling to avoid hitting the replay endpoint on fresh extractors.
- Emoji extraction with fallback chain: `emojiId` → `shortcuts[]` → `searchTerms[]` → `[emoji]`.

### Checklist


absurdlylongusername (changes requested):
Will do proper review later, but for now:

Please add tests. We should not be adding any new functionality to the extractor without adequate testing.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
```diff
@@ -43,6 +44,16 @@ public class YoutubeCommentsExtractor extends CommentsExtractor {
...
 
+    /**
+     * Live chat continuation token, used when regular comments are disabled.
+     */
+    private String liveChatContinuation;
+
+    /**
+     * Whether this video is / was a live stream.
+     */
+    private boolean isLiveStream;
```
TobiGr: I do not understand why there are two ways to determine whether this extractor handles a live stream/chat. While `isLiveStream` is used and set individually for internal purposes, `isLiveChat()` relies on `liveChatContinuation`. I think `isLiveChat` can be removed.

Please explain why two different ways are needed. As far as I understand YouTube how handles this, a video can either have comments or a live chat, but not both at the same time. Is this correct?

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
```diff
@@ -369,10 +389,110 @@ public void onFetchPage(@Nonnull final Downloader downloader)
...
+                prepareDesktopJsonBuilder(localization, getExtractorContentCountry())
+                        .value("continuation", chatContinuation)
+                        .object("currentPlayerState")
+                        .value("playerOffsetMs", "0")
+                        .end()
+                        .done())
+                .getBytes(StandardCharsets.UTF_8);
+
+        final String endpoint = "live_chat/"
+                + (isLiveStream ? "get_live_chat" : "get_live_chat_replay");
```
TobiGr: `isLiveChat` is always true here

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeLiveChatInfoItemExtractor.java
```diff
@@ -0,0 +1,186 @@
...
+            final JsonObject run = runs.getObject(i);
+            if (run.has("text")) {
+                final String text = run.getString("text", "");
+                textBuilder.append(text);
+            } else if (run.has("emoji")) {
+                final JsonObject emoji = run.getObject("emoji");
+                final String emojiText = extractEmojiText(emoji);
+                if (emojiText != null) {
+                    textBuilder.append(emojiText);
+                }
```
TobiGr: emojiText cannot be null. I think the condition can be removed.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeLiveChatInfoItemExtractor.java
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        // For standard emojis, emojiId is the Unicode character itself.
+        // For custom emojis it is an ID, but still better than nothing.
+        if (emoji.has("emojiId")) {
+            final String emojiId = emoji.getString("emojiId", "");
+            if (!emojiId.isEmpty()) {
+                return emojiId;
+            }
+        }
```
TobiGr, line 88: Having a real id as text is not good. It's way to long and always present (see below). You should check whether `isCustomEmoji` is `false` when using the emoji id.  You could leave a TODO comment to remind us discussing the inclusion of the thumbnail as `img` tag to fix this.

<details><summary>JSON</summary>

``` JSON
{
  "message": {
    "runs": [
      {
        "emoji": {
          "emojiId": "UCkszU2WH9gy1mb0dV-11UJg/uP90Xq6wNYrK8gTUoo3wAg",
          "shortcuts": [
            ":takeout:"
          ],
          "searchTerms": [
            "takeout"
          ],
          "image": {
            "thumbnails": [
              {
                "url": "https://yt3.ggpht.com/FizHI5IYMoNql9XeP7TV3E0ffOaNKTUSXbjtJe90e1OUODJfZbWU37VqBbTh-vpyFHlFIS0=w24-h24-c-k-nd",
                "width": 24,
                "height": 24
              },
              {
                "url": "https://yt3.ggpht.com/FizHI5IYMoNql9XeP7TV3E0ffOaNKTUSXbjtJe90e1OUODJfZbWU37VqBbTh-vpyFHlFIS0=w48-h48-c-k-nd",
                "width": 48,
                "height": 48
              }
            ],
            "accessibility": {
              "accessibilityData": {
                "label": "takeout"
              }
            }
          },
          "isCustomEmoji": true
        }
      }
    ]
  }
}
```
</details>

TobiGr (commented):
thank you

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeCommentsExtractor.java
Ecomont: isLiveChat() cannot be removed because it represents the extractor's configuration mode set externally via setLiveChatContinuation() whereas isLiveStream represents the video's metadata. A live-stream video may or may not have live chat configured on this extractor instance.

However, you are correct that the ternary inside fetchLiveChat() is dead code because isLiveStream is overwriten to true at the top of the method. I will either remove the ternary or pass the replay/live flag as a parameter instead.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeLiveChatInfoItemExtractor.java
Ecomont: I will remove the null check

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeLiveChatInfoItemExtractor.java
Ecomont: I will change it, thanks for the feedback

TobiGr:
LGTM now. Is there a specific reason why this is still a draft?

Ecomont:
> LGTM now. Is there a specific reason why this is still a draft?

I forgot to reopen it

TobiGr (approved):

TobiGr:
Please resolve the conflict. The PR will be merged once there are no conflicts.

extractor/src/test/java/org/schabi/newpipe/extractor/services/youtube/YoutubeCommentsExtractorTest.java
```diff
@@ -427,4 +427,50 @@ public void testGetCommentsFormatting() throws IOException, ExtractionException
...
+            org.schabi.newpipe.extractor.InitNewPipeTest.initEmpty();
+            org.schabi.newpipe.extractor.NewPipe.init(new org.schabi.newpipe.extractor.downloader.Downloader() {
+                @org.jetbrains.annotations.NotNull
+                @Override
+                public org.schabi.newpipe.extractor.downloader.Response execute(
+                        @org.jetbrains.annotations.NotNull final org.schabi.newpipe.extractor.downloader.Request request) {
+                    throw new UnsupportedOperationException("No communication expected");
+                }
+            });
+        }
```
AudricV, line 456: Imports need to be fixed.

extractor/src/main/java/org/schabi/newpipe/extractor/stream/StreamExtractor.java
```diff
@@ -56,6 +56,22 @@ public StreamExtractor(final StreamingService service, final LinkHandler linkHan
...
+
+    /**
+     * @return the live chat continuation token, or {@code null} if not available
+     */
+    public String getLiveChatContinuation() {
+        return liveChatContinuation;
+    }
+
+    protected String liveChatContinuation = null;
+
```
AudricV: This class has no field, so you shouldn't introduce a new one here.

AudricV (changes requested):
I don't really like the class design for this feature. Live chats should get their own extractor and maybe their own info items (maybe not in this PR as it requires some thinking).

AudricV:
Your branch is in a bad state with unrelated changes, please rebase it on the dev one.

PR #1485, opened 2026-05-10
Use streamAsJsonObjects

Isira-Seneviratne:
* Added uses of `streamAsJsonObjects` to avoid unnecessary lambda creations.
* Removed some unnecessary list creations.
----


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
```diff
@@ -1571,12 +1548,11 @@ public List<StreamSegment> getStreamSegments() throws ParsingException {
 
         final long duration = getLength();
         final List<StreamSegment> segments = new ArrayList<>();
-        for (final JsonObject segmentJson : segmentsArray.stream()
-                .filter(JsonObject.class::isInstance)
-                .map(JsonObject.class::cast)
-                .map(object -> object.getObject("macroMarkersListItemRenderer"))
-                .collect(Collectors.toList())
-        ) {
```
TobiGr, line 1579: where did this map ping go?
`.map(object -> object.getObject("macroMarkersListItemRenderer"))`?

TobiGr (changes requested):
thank you. Tests are failing now. I think you forgot a map() call.

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java
Isira-Seneviratne: Fixed, thanks!

TobiGr (approved):

PR #1486, opened 2026-05-11
Convert Description to a record class

Isira-Seneviratne:
* Converted the `Description` value class to a record for conciseness.
* Added a static `of` method to avoid redundant object creation for empty descriptions.
* Added a `Type` enum for type safety.

# Breaking

Increases the target and source compatibility java version from 11 to 17.

----


absurdlylongusername:
If we are intending to convert the extractor to Kotlin in the near future, I don't think refactors such as this will hold much value as they will end up being redundant.

Isira-Seneviratne:
Records have some optimizations: https://android-developers.googleblog.com/2023/06/records-in-android-studio-flamingo.html

TobiGr (commented):
Thank you. While you are at it, please add a SPDX license header to Description as well as JDoc. I have added the bump of the target and compatibility version to the description. That's a breaking change which should be visible.

TobiGr (approved):
LGTM.I'ill squash & rebase once the app PR is ready

PR #1489, opened 2026-05-20
Add more uses of streamAsJsonObjects

Isira-Seneviratne:
A follow-up PR to https://github.com/TeamNewPipe/NewPipeExtractor/pull/1485

# Breaking

Increases the target and source compatibility java version from 11 to 17.

----


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamInfoItemExtractor.java
```diff
@@ -80,36 +85,22 @@ public StreamType getStreamType() {
...
-                    .getObject("thumbnailOverlayTimeStatusRenderer")
-                    .getString("style", "");
-            if (style.equalsIgnoreCase("LIVE")) {
-                cachedStreamType = StreamType.LIVE_STREAM;
-                return cachedStreamType;
-            }
-        }
-
-        cachedStreamType = StreamType.VIDEO_STREAM;
+        final boolean isLiveBadge = videoInfo.getArray(BADGES).streamAsJsonObjects()
```
absurdlylongusername: The previous code here fails fast if isLiveBadge or isLiveThumbnail would have been true, but gets values for both even if it the first one is already true.

If you want to do `= isLiveBadge || isLiveThumbnail` then these should be lazily evaluated somehow and end up calling into the logic above to evaluate their values when it's needed for the logical OR, and not in advance.

build.gradle.kts
```diff
@@ -10,15 +10,15 @@ plugins {
...
+    version = "v0.26.1"
 
     tasks.withType<JavaCompile> {
         options.encoding = Charsets.UTF_8.toString()
     }
 
     extensions.configure<JavaPluginExtension> {
         toolchain {
-            languageVersion.set(JavaLanguageVersion.of(11))
+            languageVersion.set(JavaLanguageVersion.of(17))
```
absurdlylongusername, line 21: IIRC the reason we are still on Java 11 is because we have not set up a way to ensure that the Java APIs we use in the extractor are always available in every Android version that we support.

In Android Studio it gives warnings for this, but we dont have same protection in the build and test ecosystem for the extractor at present (until we move to Kotlin that is, then it will not be a problem).

So I am not confident we can change this to 17 yet. Is this needed in order to use the features in this PR?

absurdlylongusername (dismissed):
Seems mostly okay.

To my understanding, records should work on all versions we support so Java 17 toolchain is fine for now.

Address my comment on YoutubeStreamInfoItemExtractor and then can be merged

TobiGr (approved):

PR #1504, opened 2026-06-01
[YouTube] Use formats xTags to extract audio track type

FineFindus:
Extracts the `audioTrackType` from the explicit `xtags` field of streaming formats, instead of relying on the URL parameter.
This allows the extraction of the `audioTrackType` for streams with SABR enabled, which do have a format-specific URLs.

Ref: https://github.com/TeamNewPipe/NewPipe/issues/12248


AudricV (approved):
Thank you :)

PR #1520, opened 2026-06-29
[YouTube] Extract client version from service worker data

### LINKED ISSUE #872: Find ways to get InnerTube's client version and API key of WEB and WEB_REMIX clients without using YouTube and YouTube's Music HTML pages
I'd like to avoid parsing YouTube and YouTube Music HTML pages to get InnerTube's client version and API key of the `WEB` and `WEB_REMIX` clients (even if for the last one, the hardcoded version is always used first if it is considered as valid), as it also help to avoid reCAPTCHA issues.

Previously, we were able to do so by using YouTube and YouTube Music's service workers (https://www.youtube.com/sw.js and https://music.youtube.com/sw.js) but they do not contain these information anymore.

The "obfuscated versions" of these scripts (which use JSON protobuf) (https://www.youtube.com/sw.js_data and https://music.youtube.com/sw.js_data) still contain them, but their unreadable form makes a bit difficult to be sure to parse the correct data.

I also know that we can use the embed HTML page of a video for the `WEB` client, but versions provided are not the same as the ones of the `WEB` client, but the ones of the `WEB_EMBEDDED_CLIENT` instead. Even if using the last digits of the `WEB_EMBEDDED_CLIENT` as the ones of the `WEB` client would work, this would probably allow YouTube to fingerprint more the extractor.

That's the reason why I am asking all members working on YouTube projects to share there researches and findings about getting client version and API key of these InnerTube clients without using YouTube and YouTube Music's HTML pages in this issue.

FineFindus:
Fixes extracting the web client version, by extracting the version from the service worker data (`https://www.youtube.com/sw.js_data`). This avoids the additional request, which is necessary after failing to extract the version, when falling back to extracting version from the search page.

Closes: https://github.com/TeamNewPipe/NewPipeExtractor/issues/872


absurdlylongusername (changes requested):
Thank you for this PR @FineFindus

After reading #872 I don't think I see how this resolves it.

That issue mentions https://www.youtube.com/sw.js no longer containing something, but from what I see it still has the client version, so I presume it no longer has the API key and we used to get it from there.

The issue says it wants to find way to get **both** client version and API key, but your PR only gets client version from what I can see.

So given that I don't think this closes #872. Please correct me if I'm wrong.

Additionally, what is the benefit of getting client version from `sw.js_data` instead of `sw.js`?

Also:

1. I believe this will need to update all of related YT mocks as it's calling a new endpoint that returns a new response
2. Please add tests that prove that this works

FineFindus:
> That issue mentions https://www.youtube.com/sw.js no longer containing something, but from what I see it still has the client version, so I presume it no longer has the API key and we used to get it from there.

It seems when logged in calling the URL returns different data. When logged out I only get
```javascript
  self.addEventListener('install', event => {
    event.waitUntil(self.skipWaiting());
  });
  self.addEventListener('activate', event => {
    event.waitUntil(
      self.clients.claim().then(() => self.registration.unregister()));
  });
  ```

> The issue says it wants to find way to get both client version and API key, but your PR only gets client version from what I can see.

API keys are not used anymore, see https://github.com/TeamNewPipe/NewPipeExtractor/pull/1168.

> Please add tests that prove that this works

Not quite sure what's the best way to test this, since the method is private and the public function that uses it still works, since it falls back to extracting the version from the HTML page?

AudricV:
> The issue says it wants to find way to get both client version and API key, but your PR only gets client version from what I can see.
> 
> So given that I don't think this closes https://github.com/TeamNewPipe/NewPipeExtractor/issues/872. Please correct me if I'm wrong.

It does. The InnerTube API doesn't require an API key for a few years now.

absurdlylongusername:
@FineFindus Ah okay my bad, I was logged in 🤦‍♂️. Thank you for this information

> Not quite sure what's the best way to test this, since the method is private and the public function that uses it still works, since it falls back to extracting the version from the HTML page?

Okay I've looked at the surrounding code, I think not worth refactoring to be able to test this explicitly neither is it quick or straightforward to do so.

What we should have is a test specifically for getting client version that tests the contract of getting the client version, and tests each branch of the method, but that's prob too much for this PR.

Still need to update the mocks though.

PR #1532, opened 2026-08-14
Update Invidious instances

TobiGr:

### Breaking

`YoutubeParsingHelper.isHooktubeURL(URL url)` is removed because the service has been discontinued.

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeStreamLinkHandlerFactory.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeStreamLinkHandlerFactory.java
@@ -20,7 +20,6 @@
 
 package org.schabi.newpipe.extractor.services.youtube.linkHandler;
 
-import static org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper.isHooktubeURL;
 import static org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper.isInvidiousURL;
 import static org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper.isY2ubeURL;
 import static org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper.isYoutubeServiceURL;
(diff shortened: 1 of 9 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeStreamLinkHandlerFactory.java
```diff
@@ -197,23 +197,11 @@ public String getId(final String theUrlString)
             case "INVIDIOUS.KAVIN.ROCKS":
             case "INVIDIOUS-US.KAVIN.ROCKS":
```
AudricV: `hooktube.com`, `invidious.snopyta.org`, `tube.connect.cafe`, `tubus.eduvid.org`, `invidious.kavin.rocks`, `invidious-us.kavin.rocks`, `dev.invidio.us`, `www.invidio.us`, `invidio.us` can be also removed (some tests other than the ones already updated will need to be updated too).

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeStreamLinkHandlerFactory.java
```diff
@@ -197,23 +197,11 @@ public String getId(final String theUrlString)
...
-            case "INVIDIOUS.BLAMEFRAN.NET":
-            case "INVIDIOUS.MOOMOO.ME":
-            case "YTB.TROM.TF":
-            case "YT.CYBERHOST.UK":
-            case "Y.COM.CM": { // code-block for hooktube.com and Invidious instances
+            case "INV.NADEKO.NET":
+            case "INVIDIOUS.NERDVPN.DE":
+            case "YT.CHOCOLATEMOO53.COM":
+            case "INVIDIOUS.TIEKOETTER.COM":
+            case "INVIDIOUS.F5.SI": { // code-block for hooktube.com and Invidious instances
```
AudricV:
```suggestion
            case "INVIDIOUS.F5.SI": { // code-block for Invidious instances
```

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeStreamLinkHandlerFactory.java
TobiGr: thanks. done

AudricV (commented):
You should probably add `YoutubeParsingHelper.isHooktubeURL` removal is a breaking change, as it is part of the public methods.

`piped.video` support is also missing.

TobiGr:
> You should probably add YoutubeParsingHelper.isHooktubeURL removal is a breaking change, as it is part of the public methods.

Done

> piped.video support is also missing.

Added

AudricV (approved):

PR #1533, opened 2026-08-15
Fix media.ccc.de live streams

TheAssassin:
This PR fixes the fetching of live streams in NewPipe's media.ccc.de "Live" page with the currently running conference (and probably all future conferences, too). The issue originated from a misconception on our side, based on the fact that none of this was documented on media.ccc.de's side at the time of writing.

Based on discussion with a C3VOC member, we should use the mandator instead of the conference slug.

- [x] I agree to create a pull request for [NewPipe](https://github.com/TeamNewPipe/NewPipe) as soon as possible to make it compatible with the changed API. (No API change required, though, just a dependency update)


TobiGr:
Please target the master branch

TobiGr (approved):

PR #1545, opened 2026-09-08
[YouTube] Update n-sig parsing

FineFindus:

Updates the extraction of the n parameter and signature deobfuscation functions (adapted from https://github.com/yt-dlp/ejs/blob/6f8587bb7009a1fc81038538071d2cb66b8b8ed0/src/yt/solver/nsig.ts). While both (`n`, `sig`) use the same deobfuscation function, as far as I could detect, only the `n` deobfuscation is currently used.

Unfortunately, the deobfuscation function uses externally defined functions (i.e. not defined inside the function), requiring to load/parse the entire player code, taking multiple seconds. I did try to optimize this by only extracting the called functions, unfortunately that turned out to be more difficult than expect.
As an alternative, I opted to add a method that allows applications to set their own (faster) JavaScript Interpreter (for example an webview, which is already implemented by most applications to generate poTokens).

### DIFF
--- a/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingParameterUtils.java
+++ b/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingParameterUtils.java
@@ -18,6 +18,12 @@ import java.util.regex.Pattern;
  */
 final class YoutubeThrottlingParameterUtils {
 
+    /**
+     * The name of the deobfuscation function which needs to be called inside the deobfuscation
+     * code.
+     */
+    static final String DEOBFUSCATION_FUNCTION_NAME = "deobfuscate_n_param";
+
     // NOTE: When changing this you should also change the quick exit/shortcut
     // in getThrottlingParameterFromStreamingUrl
     private static final Pattern THROTTLING_PARAM_PATTERN = Pattern.compile("[&?]n=([^&]+)");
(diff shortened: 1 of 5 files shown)


extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingParameterUtils.java
AudricV, line 1: You should use the video ID provided when calling the extraction of the JavaScript code instead of `yt-dlp-wins`.

extractor/src/main/java/org/schabi/newpipe/extractor/utils/JavaScriptInterpreter.java
```diff
@@ -0,0 +1,25 @@
...
+     *
+     * @param code the code that contains the function that should be run
+     * @param functionName the name of the function that should be run
+     * @param parameters the parameters to run with the function
+     * @return the result of the run function
+     */
+    String run(@Nonnull final String code,
+               @Nonnull final String functionName,
+               final String... parameters);
+}
```
AudricV: A new line is missing here.

AudricV (changes requested):

extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeThrottlingParameterUtils.java
FineFindus: That would break the function caching (as we would need to rebuild the function on every video extraction). Considering that video extraction already rather slow, I would prefer to avoid that.

